PR #1, opened 2020-01-29
W3c split val

SebastianSchildt:
I moved everything - UNMODIFIED - from the w3c-visserver folder in kuksa.invehicle to kuksa.val as discussed in the mailing list

This can be merged, as there are no new dependencies and no new code that hasn't been cleared already in KUKSA.

@tspreck: W3C dev should continue here. Some Git-Fu might be needed to migrate existing forks left over from @BojanKV to here.  Basically I used this https://help.github.com/en/github/using-git/splitting-a-subfolder-out-into-a-new-repository to cut out the w3c-visserver directory. Theoretically, if we do that on the  remaining w3c relevant forks of in.vehicle, and push it to a new branch in a kuksa.val fork, it might be mergeable to an existing kuksa.val branch (potentially with `--allow-unrelated-histories`). Then github PRs to Eclipse kuksa.val should work as usual. We can try next week :)

Once this is merged by @rohoet or @daschubert I will update kuksa.invehicle with a PR that replaces the w3c-visserver folder with a git submodule pointing to the initial kuksa.val version. That way no functionality is lost and we do not have double code (which also makes it easier, to use newer kuksa.val versions in in.vehicle, because you just need to bump the submodule reference)


rohoet:
Thanks for the effort, @SebastianSchildt !

PR #3, opened 2020-02-04
 Re-factoring and cleanup

tspreck:
Have re-configured library paths and fixed the problem of why it wasn't building

- No new dependencies
- Have copied across 3rd-party-libs from kuksa.invehicle repo
- Added a setup_build script for making authentication easier prior to build
- Corrected misleading example text and value setting in test-client

@SebastianSchildt  has agreed to test this works on his machine also


SebastianSchildt:
Tested and works. Can be merged. This does not include any new dependencies. Unfortunately this is large this also contains some changes Bojan has ben working on before but did not yet merge. We will try to keep further PRs smaller, but we need to get this one out of the door, because we are working from here....

Hope @rohoet  or @daschubert  can merge this. If there questions please ask.

rohoet:
Ok, thanks for the update!
How about `turtle`? I noticed lots of references in the code and it seems to be a BSD licensed mock object library based on Boost that we do not have a CQ for, yet.

SebastianSchildt:
Ahhrg. It is for our unit tests. I did miss, that we did not upstreamed them here :(  Sorry. What kind of information and what kind of incentive it would take for you to open a CQ? I assume turtle version, link to source code, license information and promise of a sufficient amount of chocolate next time we meet?

rohoet:
No worries 😊 Test-only dependencies are workswith requests and can be combined in a single CQ.
We already have such CQs and I'll just enter another one for turtle.
This is usually only crucial for the next release so I think merging this PR should be fine for now.

PR #4, opened 2020-02-10
Add missing stuff

SebastianSchildt:
The last commit forgot to copy the third party libs from in-vehicle. These are the same dependencies that have already been and cleared in in-vehicle:
 - jsoncons + ext
 - jwt-cp
Those have been copied directly form the old repo.

Also it includes unit-tests and the requried turtle dependencies discussed in #3 

@rohoet  Hope this is the last big one 
@tspreck can you confirm that you have tested this?


tspreck:
I have tested this to be working - so you have the thumbs up from me. This now leaves the master branch at a solid starting point for future development work 👍

PR #8, opened 2020-03-26
Feature/better logging

### LINKED ISSUE #5: Better error messages
We need better error logging. E.g. when not finding the server certificate it does ot give a hint which file it tried to read

`Log START
VERBOSE: SubscribeThread: Started Subscription Thread!
VERBOSE: VssDatabase::VssDatabase : VSS tree initialized using JSON file = /config/vss.json
Could not load server certificate
`

SebastianSchildt:
Print more explicit error messages when key or certificate is missing and print webseocket/REST port. Fixes #5 

main.cpp was formatted according to the coding style in other files

No additional dependencies

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -13,27 +13,26 @@
  */
 
 #include <gio/gio.h>
-#include <stdlib.h>
 #include <stdio.h>
-#include <iostream>
-#include <string>
+#include <stdlib.h>
 #include <ctime>
 #include <exception>
+#include <iostream>
+#include <string>
 
 #include <boost/program_options.hpp>
 #include <jsoncons/json.hpp>
 #include <jsonpath/json_query.hpp>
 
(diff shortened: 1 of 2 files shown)


src/main.cpp
```diff
@@ -369,8 +338,8 @@ int main(int argc, const char *argv[]) {
     cerr << ex.what() << std::endl;
     return -1;
   } catch (const std::runtime_error &ex) {
-    cerr << ex.what() << std::endl;
-        return -1;
+    logger->Log(LogLevel::ERROR, "Fatal runtime error: "+ string(ex.what()));
```
SebastianSchildt: The only functional change in main.cpp. Use logger to print  exceptions (and initialize logger before the big try clause to enable this)

rohoet:
LGTM, thanks !

PR #9, opened 2020-04-01
JWT helper scripts and docs & fix unittests

SebastianSchildt:
This hopefully clears up the bad documentation and unclear way how to use and create JWT tokens.

It includes two helper scripts to regenerate the JWT key pair and to create & sign valid JWT tokens from JSON snippets. It improves JWT related logging and loading of public key in the server code 

Dependencies: I am not sure if relevant, but the certificates/jwt/createToken.py imports "jwt" for token creation. Any version is fine and released 1.7.1 here https://github.com/jpadilla/pyjwt/blob/1.7.1/LICENSE is MIT. Definitely only "works-with", as it is only in the convenience script, not in kuksa.val source

Might also close https://github.com/eclipse/kuksa.invehicle/issues/116 

Because some changes needed to be done in unit tests, I repaired them by removing bogus tests that didn't make sense and repaired the remaining failing ones.

@rohoet Any more information needed?


SebastianSchildt:
And it builds https://kuksaval.northeurope.cloudapp.azure.com/job/kuksaval-upstream/view/change-requests/job/PR-9/

PR #13, opened 2020-05-06
Return error code when docker build runs into problems

SebastianSchildt:
Makes sure scripts returns with an error when docker build fails, so Jenkins notices and doesn't start archiving stale docker images


SebastianSchildt:
No new  dependenices @daschubert @rohoet   , merge plz 🐱

PR #14, opened 2020-06-22
Feature/dbcfeeder

SebastianSchildt:
Example feeder for feeding CAN Bus data directly using a DBC file. 

Dependency https://pypi.org/project/cantools/  any version you like, e.g. https://pypi.org/project/cantools/34.2.0/   MIT license  also seen at https://github.com/eerimoq/cantools/tree/34.2.0 with MIT license

@rohoet @daschubert

### DIFF
--- a/examples/dbc2val/dbcreader.py
+++ b/examples/dbc2val/dbcreader.py
@@ -37,6 +37,7 @@ class DBCReader:
                 #print("Decod" +str(decode))
             except:
                 self.parseErr+=1
+                #print("Error Decoding")
                 continue
             rxTime=time.time()
             for k,v in decode.items():
--- a/examples/dbc2val/dbcreader.py
+++ b/examples/dbc2val/dbcreader.py
@@ -19,14 +19,41 @@ class DBCReader:
     def __init__(self, cfg, rxqueue, mapper):
         self.queue=rxqueue
         self.mapper=mapper
+        self.cfg=cfg
         print("Reading dbc file")
         self.db = cantools.database.load_file(cfg['vss.dbcfile'])
 
+        self.canidwl = self.get_whitelist()
+
         self.parseErr=0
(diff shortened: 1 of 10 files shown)


SebastianSchildt:
FCA repariert @rohoet

examples/dbc2val/dbcreader.py
```diff
@@ -0,0 +1,77 @@
...
+# Copyright (c) 2020 Robert Bosch GmbH
+#
+# This program and the accompanying materials are made
+# available under the terms of the Eclipse Public License 2.0
+# which is available at https://www.eclipse.org/legal/epl-2.0/
+#
+# SPDX-License-Identifier: EPL-2.0
+########################################################################
+
+import can, cantools
```
rohoet, line 13: `python-cantools` v 34.0.0  (MIT licensed) is approved as workswith, but python-can (LGPL) is not.

SebastianSchildt: Ayay always 20% coding 80% red tape at Eclipse :) Thanks for pointing it out. I think it can be done: By definition this whole thing is an optional example, and not part of the server, so workswith should work. For python-can it is just one of those standard packages where verison doesn't really matter. If there isn't any CQ to piggy back, can we do


Python Can 3.3.3 for example here https://github.com/hardbyte/python-can/tree/3.3.3 with LGPL license or alternatively here https://pypi.org/project/python-can/3.3.3/ (dame thing) 

I understand LGPL is ok for workswith?

rohoet: The CQ for python-can is filed, but as of https://www.eclipse.org/lists/eclipse.org-committers/msg01241.html , I feel confident with already merging this PR, since python-can is not a core part of Eclipse Kuksa, we do not distribute it in our repo, and its use is just intended for an example.

Issue #15, opened 2020-07-17
MQTT Publisher


SebastianSchildt:
We should be able to deliver VSS data to a MQTT broker for further processing by northbound apps.

I.e. a path such as `Vehicle.OBD.Speed' shall be published to /prefix/Vehicle/OBD/Speed

To do so we need to get config parameters for MQTT Server, Port, unsername, PW. We also need a config file specifying which paths to push (an alternative mode should allow pushing ALL data, but this may not be always feasible if there are many high-
datarate signals). Pushing ALL data is useful for feeding a local MQTT, pushing only a select view signals is useful for driectly connecting kuksa.val to a cloud MQTT broker without any intermediate processing app.

MQTT library should be  some Eclipse Paho derivative for license reasons. this should be part of the kuksa.val server and not an external program

![image](https://user-images.githubusercontent.com/3707124/87798530-6d987880-c84c-11ea-9f06-f9968dd3f667.png)

Optional

- Add a command&control channel, so during runtime you configure which data points are sent or net (publish/unpublish)
- Also allow kuksa.val to subscribe MQTT (enabling a MQTT feeder)

SebastianSchildt:
Closed by #17

PR #17, opened 2020-07-31
15 mqtt publisher

wenwenchenbosch:


wenwenchenbosch:
solve #15

SebastianSchildt:
@rohoet Only new dependency is mosquitto (the client libriaries) https://mosquitto.org/download/  
As this is an Eclipse project under EPL1.0/EDL1.0 dual license. Is this "auto-cleared" or do you need more information?
https://github.com/eclipse/mosquitto/blob/master/LICENSE.txt

rohoet:
> @rohoet Only new dependency is mosquitto (the client libriaries) https://mosquitto.org/download/
> As this is an Eclipse project under EPL1.0/EDL1.0 dual license. Is this "auto-cleared" or do you need more information?
> https://github.com/eclipse/mosquitto/blob/master/LICENSE.txt

No more information required. EPL software is indeed _auto-cleared_.

PR #19, opened 2020-07-31
Added first version of VSS Test Client written in Python using Cmd2.

nayakned:
Signed-off-by: Nayak Naresh (CR/ADT1) <Naresh.Nayak@de.bosch.com>


SebastianSchildt:
@rohoet   New dependency for this is cmd2: https://pypi.org/project/cmd2/  MIT License.
If you can not piggy back a CQ (any version) try this: Version 1.2.1 here https://github.com/python-cmd2/cmd2/tree/1.2.1

Other dependency websockets _should_ already be cleared, as it is used by the feeder. Otherwise: https://pypi.org/project/websockets/ BSD license  tag 8.1  https://github.com/aaugustin/websockets/tree/8.1

vss-testclient/README.md
```diff
@@ -0,0 +1,37 @@
...
+Ensure all the dependencies are installed.
+```sh
+pip3 install websockets cmd2
+python3 testclient.py
+```
+
+### Usage
+Set the Server IP Address & Connect
+```
+VSS Client> setServerIP 127.0.0.1 8090
```
SebastianSchildt, line 30: It is `setServerAddress`which is also fine, but it only supports IP addresses, e.g. `VSS Client> setServerAddress localhost 8090` fails with `Please give a valid server Address`

For "Address" I would expect also names to work, if only IP is possible at least it should state so clearly

vss-testclient/testclient.py
```diff
@@ -0,0 +1,169 @@
...
+ap_disconnect = argparse.ArgumentParser()
+ap_authorize = argparse.ArgumentParser()
+ap_authorize.add_argument('Token', help='JWT for authorizing the client.')
+ap_setServerAddr = argparse.ArgumentParser()
+ap_setServerAddr.add_argument('IP', help='VSS Server IP Address')
+ap_setServerAddr.add_argument('Port', type=int, help='VSS Server Websocket Port')
+ap_setValue = argparse.ArgumentParser()
+ap_setValue.add_argument("Parameter", help="Parameter to be set")
+ap_setValue.add_argument("Value", help="Value to be set")
+ap_getValue = argparse.ArgumentParser()
```
SebastianSchildt, line 29: Not sure this code is related :-), but for the usability: Wouldn't it be better to just use "get" and "set" insted of "getValue" and "setValue" (as those two are probably used most often)

vss-testclient/README.md
nayakned: With the changes, now even server names work for connection.

vss-testclient/testclient.py
nayakned: _get_ and _set_ are both keywords in the Cmd2 library to work with internal debug variables.
We can ofcourse override them for the testclient, but then this will be counter-productive for further development, in my opinion. 
Hence I suggest to not modify the code for this purpose.

vss-testclient/testclient.py
```diff
@@ -0,0 +1,163 @@
...
+from clientComm import VSSClientComm
+
+ap_getServerAddr = argparse.ArgumentParser()
+ap_connect = argparse.ArgumentParser()
+ap_connect.add_argument('-i', "--insecure", default=False, action="store_true", help='Connect in insecure mode')
+ap_disconnect = argparse.ArgumentParser()
+ap_authorize = argparse.ArgumentParser()
+ap_authorize.add_argument('Token', help='JWT for authorizing the client.')
+ap_setServerAddr = argparse.ArgumentParser()
+ap_setServerAddr.add_argument('IP', help='VSS Server IP Address')
```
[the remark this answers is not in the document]
nayakned: Added default server address as localhost and port as 8090.

vss-testclient/testclient.py
SebastianSchildt: makes sense 👍

SebastianSchildt:
Tested again, and works. From a technical perspective this can be merged. @rohoet : Do you have all license infos you need  (see my first post)?

rohoet:
> Tested again, and works. From a technical perspective this can be merged. @rohoet : Do you have all license infos you need (see my first post)?

Yes - CQ created already some time ago. I'll have another look and merge if I do not find anything else.

Issue #22, opened 2020-08-06
New testclient, websocket connection handling


SebastianSchildt:
Something seems to be wrong with websocket connection handling

Reproduce

- Start server
- Connect with new testclient (connect authorize setValue)
- works fine
- Restart server (so the old connection is broken/invalid)
- in testclient type "connect"
- authorize: That works! (so I assume it DID reconnect?)
 - setValue: first time works
 - setValue second time: Exception: websockets.exceptions.ConnectionClosedError: code = 1006 (connection closed abnormally [internal]), no reason
 - Ctrl-C gives back prompt
 - setValue works again

Already not perfect, but now the super strange thing:
type `connect` again: Looks fine, but now roughly every second call to setValue succeeds while the others give 403
Do we have two sessions now, that are randomly used

@nayakned Can you check, and maybe push a fix?

wenwenchenbosch:
A new robust problem found:

 - Start server
 - Start testclient
 - In testclient type `Connect` two times (second time maybe by mistake)
 - Authorize
 
works fine until now, but then setValue may be failed with the following error message:

```
VSS Client> setValue Vehicle.Speed 500
{
    "action": "set",
    "error": {
        "message": "Path(s) in set request do not have write access or is invalid",
        "number": 403,
        "reason": "Forbidden"
    },
    "requestId": "1235",
    "timestamp": 1596814834
}
```

nayakned:
Try Vehicle.OBD.Speed if the problem persists.

SebastianSchildt:
Closed by #28

PR #28, opened 2020-08-07
Cleaning up previous connection sessions when connect command is re-e…

nayakned:
…xecuted.

Signed-off-by: Nayak Naresh (CR/ADT1) <Naresh.Nayak@de.bosch.com>


nayakned:
Fix for Issue #22

wenwenchenbosch:
> See @wenwenchenbosch comment in the issue, Maybe we have several websockets lying around that are randomly used when doing several `connect`?

I tested with this new branch, the issues are fixed now

PR #30, opened 2020-08-12
Feature/stresstest

SebastianSchildt:
A small testing app. If it works, it works, if it doesn't, it doesn't matter...

No new dependencies (it is basically  a stripped down feeder, the websocket lib was already cleared for dbcfeeder)

Can also be merged @rohoet


rohoet:
LGTM :-)

PR #31, opened 2020-08-13
Clean artifacts before build to prevent recursively nested docker ima…

SebastianSchildt:
Fixes the problem that the ubunut-dev docker on Jenkins always includes the previous version. (and thus keepsgrwing with every build).

The problem was, that the worksapce was ADDed to docker, but the artefacts folder, with the exported docker images form last build wasn't cleaned.

Now we clean before starting the build

@wenwenchenbosch  Can you look at it, as a "sanity" check. It solves the immediate problem of our disk getting filled :)


wenwenchenbosch:
it solve the problem.

SebastianSchildt:
@rohoet or @daschubert  pls merge.
First merger can have this: 🐘

PR #36, opened 2020-08-24
34 renaming

wenwenchenbosch:


CMakeLists.txt
```diff
@@ -12,7 +12,7 @@
 # *****************************************************************************
 
 cmake_minimum_required(VERSION 3.12.0)
-project(w3c-visserver-api VERSION 2.0.1 LANGUAGES CXX)
+project(kuksa-val VERSION 2.0.1 LANGUAGES CXX)
```
SebastianSchildt, line 15: Should we update version? No idea what the meaning in cmake is, but it might be "the right thing to do"?

wenwenchenbosch: I do not think so, we changed only name, not api. VERSION in CMake is for cmake internal usage see https://cmake.org/cmake/help/v3.17/command/project.html?highlight=project.

SebastianSchildt (approved):
Checked local build, docker images and new tokens. Everything works as expected.  Can be merged @rohoet

PR #37, opened 2020-08-24
Fix docker image tagging in  jenkins

SebastianSchildt:
Our "production" dockers did not have tags attached when "loading" them, because we saved them basing on SHA Hash (see here https://github.com/moby/moby/issues/3877 )  

This has been fixed.

Also the docu was missing to expose a port.

@wenwenchenbosch This fix couldn't possibly go wrong, but pls. check if it still builds/works for you 😄

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -27,8 +27,8 @@ node('docker') {
         }
         stage('Collect') {
 			sh '''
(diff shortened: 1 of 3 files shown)


wenwenchenbosch (approved):

SebastianSchildt:
Mergeable @rohoet  Thx.

Issue #33, opened 2020-08-20
Extend data model during runtime


SebastianSchildt:
Currently the server loads a JSON file describing the VSS tree structure at startup time. It would be nice, to add the feature to load additonal JSONs of the same format during runtime using another websocket request or REST call

 - the new file should extend the existing tree (adding branches/leafs)
 - potentially be able to overwrite -> We need to decide what is useful here
 - protected by a new entry/claim in the JWT token

"Unloading" of of elements I would consider optional

SebastianSchildt:
First version in #42

PR #42, opened 2020-08-31
33 extend data model

wenwenchenbosch:
solve #33


wenwenchenbosch:
using `updateMetaData <path> <json/file>` in testclient to update a datamodel

SebastianSchildt (changes requested):
JWT access permission need to be reeavluated when modifying tree. Reproduce

- Use token with *=rw acces
- Extend tree
- Try to acces newly added item
  -  Expectation: Works
  - Situation: 403
- Authorize with same token again
- Access newly added item: Works now

certificates/jwt/super-admin-metadata.json
```diff
@@ -0,0 +1,12 @@
+{
+  "sub": "kuksa.val",
+  "iss": "Eclipse KUKSA Dev",
+  "admin": true,
+  "metadata": "rw",
```
SebastianSchildt, line 5: modifyTree? modifyVSSTree?

wenwenchenbosch:
> JWT access permission need to be reeavluated when modifying tree. Reproduce
> 
>     * Use token with *=rw acces
> 
>     * Extend tree
> 
>     * Try to acces newly added item
>       
>       * Expectation: Works
>       * Situation: 403
> 
>     * Authorize with same token again
> 
>     * Access newly added item: Works now

this bug fixed now

docker/Dockerfile-Jenkins-Build-Env
```diff
@@ -4,7 +4,7 @@ FROM ubuntu:18.04
 
 # Install all the pre-requisits we need in the docker image
 RUN apt-get update \
-  && apt-get install -y qemu-user-static docker.io sudo bzip2
+  && apt-get install -y qemu-user-static docker.io sudo bzip2 ccache
```
SebastianSchildt, line 7: Isn't ccache useless here? We never run a compiler in this container

wenwenchenbosch: with `docker build`: `RUN make -j8` the local gcc should be used. But using ccache is still not configured within jenkins. There are still more works to do. It is a preparation.

certificates/jwt/super-admin-metadata.json
wenwenchenbosch: really necessary to change?

SebastianSchildt: I think `modifytree`  describes more clearly what we are doing here. But now I am puzzled about "r" .. are we also protecting getMetadata now?

wenwenchenbosch: no, default always read-only. But it is also possible to extend it

wenwenchenbosch: I update the key work to `modifyTree`. Question: do we need two token?

SebastianSchildt (approved):
Working fine, can be merged now @rohoet

docker/Dockerfile-Jenkins-Build-Env
rohoet: Is ccache now required here or not? Its license is not compliant with EPL.

SebastianSchildt: Hi. 
It is just our build environment, it has nothing to do with the software  (and it is a complete ubuntu 18.04, we can not clear everything in there)

wenwenchenbosch: `ccache` removed

PR #43, opened 2020-08-31
fetch boost 1.67 if boost is not installed on pc

wenwenchenbosch:


SebastianSchildt (approved):
Works. Still working with system boost when installed, and will use a local copy, if not found.

Can be merged @rohoet

SebastianSchildt:
or @daschubert  🍰

docker/Dockerfile-Jenkins-Build-Env
```diff
@@ -4,7 +4,7 @@ FROM ubuntu:18.04
 
 # Install all the pre-requisits we need in the docker image
 RUN apt-get update \
-  && apt-get install -y qemu-user-static docker.io sudo bzip2
+  && apt-get install -y qemu-user-static docker.io sudo bzip2 ccache
```
rohoet, line 7: Same here. Is `ccache` necessary?

SebastianSchildt: https://github.com/eclipse/kuksa.val/pull/42#discussion_r483722967

PR #48, opened 2020-09-04
add autocompletion and colored output

wenwenchenbosch:
add the following feature:
- autocompletion path for `getValue`, `setValue` and `getMetaData`
- autocompletion files only `.token` file
- colored json output

unfortunately, cmd2 does not support coloring input and autocompletion text. @SebastianSchildt & @nayakned : If you can some thing about it, pls let me know

### DIFF
--- a/vss-testclient/testclient.py
+++ b/vss-testclient/testclient.py
@@ -10,45 +10,89 @@
 ########################################################################
 
 import argparse, json, sys
-from cmd2 import Cmd, with_argparser, with_category
+from typing import Dict, List
 import queue, time, os
 from clientComm import VSSClientComm
-
+from pygments import highlight, lexers, formatters
+from cmd2 import Cmd, with_argparser, with_category, Cmd2ArgumentParser, CompletionItem
+from cmd2.utils import CompletionError, basic_complete
+import functools
 DEFAULT_SERVER_ADDR = "localhost"
 DEFAULT_SERVER_PORT = 8090
-
-ap_getServerAddr = argparse.ArgumentParser()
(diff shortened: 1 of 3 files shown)


SebastianSchildt:
Cool! @nayakned  can you check if it works for you?

vss-testclient/testclient.py
```diff
@@ -10,45 +10,89 @@
 ########################################################################
 
 import argparse, json, sys
-from cmd2 import Cmd, with_argparser, with_category
+from typing import Dict, List
 import queue, time, os
 from clientComm import VSSClientComm
-
+from pygments import highlight, lexers, formatters
```
rohoet, line 16: `pygments` is not part of the standard python library, right? Even if it is just a test dependency, it requires approval.

SebastianSchildt: @wenwenbosch Can you provide Robert with a link to the pip/github and license?

wenwenchenbosch: https://github.com/pygments/pygments has BSD-2-Clause License

nayakned:
@wenwenchenbosch You need to add the pygments to Pipfile & update the Pipfile.lock

vss-testclient/testclient.py
SebastianSchildt: And this is only part of the testclient, not part of the main server

wenwenchenbosch:
> 
> 
> @wenwenchenbosch You need to add the pygments to Pipfile & update the Pipfile.lock

fixed

PR #53, opened 2020-09-12
Feature/update dependencies

### LINKED ISSUE #44: Update JSONCons
We use jsonscons to deal with... JSON

We use a pretty old version (0.105.0) and should update it.

We already had an update for 0.130.0 with CQ20506 here https://github.com/eclipse/kuksa.apps/pull/24 

Current version is  0.156.0  https://github.com/danielaparker/jsoncons/releases

@rohoet  How hard is it to clear a new version? As it is just bugfixe with same license, would you be able to merge such a change and CQ later? If not is there alreay a newer version for piggy back? If not can you make a CQ for 0.156.0? (should have written this in pseudocode 😄  ) 

@wenwenchenbosch  Any opinion whether we should switch to submodule (or repo tool), instead of mirroring here, in case we update?

SebastianSchildt:
Pull request updating some depenedenices. 

Fixes #44 

**Updated JSONCOns**
Remove source and added as submodule. This is Version 0.156.0  
@rohoet Compare our submodule reference to jsoncons commit 5be801 which is the final commit of  https://github.com/danielaparker/jsoncons/tree/v0.156.0

**Updated jwt-cpp**
Remove source and added as submodule. I do not know our current version. I added v0.4.0. https://github.com/Thalhammer/jwt-cpp/tree/v0.4.0 @rohoet It is still MIT license and see our submodule commit id pointing to 34bb06, the last commit of v.0.4.0

I also needed to replace some deprecated API calls for JSONCons. Seems to work fine

⚠️  * Do not merge before somebody else tested this* 

@wenwenchenbosch  Please review/test this!

@nayakned You may want to give this a small testride, to see it is still working

@rohoet Please check if you forsee any troubles with dependencies. I think it should be fine, but as I said I have no idea which jwt-cpp version was cleared before


SebastianSchildt:
Can smebody check @rohoet  @eriksven  waht the problem with the build is? The hook says `This commit cannot be built`

Looking at the llink I see https://ci.eclipse.org/kuksa/job/kuksa.val/view/change-requests/job/PR-53/2/console fails with some missing includes, howeer the identical code base builds here https://kuksaval.northeurope.cloudapp.azure.com/job/kuksaval-boschresearch/job/feature%252Fupdate-dependencies/5/console  

I think the root cause is, eclipse not checling out the submodules, if you see the first lines in the Jenkinsfile, correct is
```
******** -e ******** -e ******** -e ******** -e ******** -e ******** -e ******** -e ******** my-image:5 cat
$ docker top 4577edbadd724499b2705f63ce862773342fdd8e60a9f34cbf0dd667fc0e6b08 -eo pid,comm
[Pipeline] {
[Pipeline] stage
[Pipeline] { (Prepare)
[Pipeline] sh
+ git submodule update --init
+ mkdir -p artifacts
+ rm -f artifacts/kuksa-val-amd64.tar.bz2 artifacts/kuksa-val-arm64.tar.bz2 artifacts/kuksa-val-dev-ubuntu20.04.tar.bz2
[Pipeline] }
```` 

while the Eclipse CI only does

```
[Pipeline] {
[Pipeline] stage
[Pipeline] { (Prepare)
[Pipeline] sh
+ mkdir -p artifacts
+ rm -f artifacts/*
[Pipeline] }
[Pipeline] // stage
[Pipeline] stage
[Pipeline] { (Build)
[Pipeline] sh
+ ./docker/build.sh amd64
Sending build context to Docker daemon  5.306MB
```

so no submodules, I _do_ see however this suspicious message in Eclipse side

```
‘Jenkinsfile’ has been modified in an untrusted revision
```

So does it use an older, cached "safe" version? If so, can you let the Jenkins trust the Jenkinsfile for this PR, so the build is checked correctly?

File: src/WebSockHttpFlexServer.cpp [hunk 1]
```diff
@@ -22,6 +22,7 @@
 #include <boost/make_unique.hpp>
 #include <cstdlib>
 #include <iostream>
+#include <fstream>
```
wenwenchenbosch, line 25: do we need this?

src/VssDatabase.cpp
```diff
@@ -179,13 +179,13 @@ namespace {
...
                     string key) {
-    if (!source.has_key("type")) {
+    if (!source.contains("type")) {
       string msg = "Unknown type for signal found at " + key;
       logger->Log(LogLevel::ERROR, "VssDatabase::setJsonValue : " + msg);
 
       throw genException(msg);
     }
-    dest[key] = source["value"];
+    dest[key] = source["value"].as<string>();
```
wenwenchenbosch, line 188: why as<string>()?

src/VssCommandProcessor.cpp
```diff
@@ -498,7 +498,7 @@ string VssCommandProcessor::processQuery(const string &req_json,
         return JsonResponses::malFormedRequest("Unknown action requested");
       }
     }
-  } catch (jsoncons::json_parse_exception &e) {
+  } catch (jsoncons::ser_error &e) {
     logger->Log(LogLevel::ERROR, "JSON parse error");
```
wenwenchenbosch, line 502: not only parse error, right?

wenwenchenbosch (approved):
some questions, but I tested with testclient and it works.

src/WebSockHttpFlexServer.cpp
SebastianSchildt: Without it, does not compile `‘std::ifstream inputFile’ has initializer …
…but incomplete type error`, i.e. https://github.com/eclipse/kuksa.val/blob/0dd7ab3c3a5a07c986b627d3f59c30027054bcee/src/WebSockHttpFlexServer.cpp#L1403
i

Maybe with the older jsoncons/jwt was included somewhere deeper in the hierarchy.... But here really is correct

src/VssDatabase.cpp
SebastianSchildt: Current JSONCons doesn't compile otherwise (Doesn't assume any default type)

src/VssCommandProcessor.cpp
SebastianSchildt: json_parse_exception has been replaced, see https://github.com/danielaparker/jsoncons/blob/5be801990f996f9905e8a3d442f3810ae21c74de/include/jsoncons/json_exception.hpp#L172

SebastianSchildt:
@rohoet  From a techical point of view, pls merge 😄

wenwenchenbosch:
> 
> 
> Can smebody check @rohoet @eriksven waht the problem with the build is? The hook says `This commit cannot be built`
> 
> Looking at the llink I see https://ci.eclipse.org/kuksa/job/kuksa.val/view/change-requests/job/PR-53/2/console fails with some missing includes, howeer the identical code base builds here https://kuksaval.northeurope.cloudapp.azure.com/job/kuksaval-boschresearch/job/feature%252Fupdate-dependencies/5/console
> 
> I think the root cause is, eclipse not checling out the submodules, if you see the first lines in the Jenkinsfile, correct is
> 
> ```
> ******** -e ******** -e ******** -e ******** -e ******** -e ******** -e ******** -e ******** my-image:5 cat
> $ docker top 4577edbadd724499b2705f63ce862773342fdd8e60a9f34cbf0dd667fc0e6b08 -eo pid,comm
> [Pipeline] {
> [Pipeline] stage
> [Pipeline] { (Prepare)
> [Pipeline] sh
> + git submodule update --init
> + mkdir -p artifacts
> + rm -f artifacts/kuksa-val-amd64.tar.bz2 artifacts/kuksa-val-arm64.tar.bz2 artifacts/kuksa-val-dev-ubuntu20.04.tar.bz2
> [Pipeline] }
> ```
> 
> while the Eclipse CI only does
> 
> ```
> [Pipeline] {
> [Pipeline] stage
> [Pipeline] { (Prepare)
> [Pipeline] sh
> + mkdir -p artifacts
> + rm -f artifacts/*
> [Pipeline] }
> [Pipeline] // stage
> [Pipeline] stage
> [Pipeline] { (Build)
> [Pipeline] sh
> + ./docker/build.sh amd64
> Sending build context to Docker daemon  5.306MB
> ```
> 
> so no submodules, I _do_ see however this suspicious message in Eclipse side
> 
> ```
> ‘Jenkinsfile’ has been modified in an untrusted revision
> ```
> 
> So does it use an older, cached "safe" version? If so, can you let the Jenkins trust the Jenkinsfile for this PR, so the build is checked correctly?

I think, after merging the PR to master, jenkins can build it again, because in https://github.com/boschresearch/kuksa.val, jenkins can build the branch successfully. If you really want the jenkins can also build this PR. I think, just also bringe the branch to eclipse repo will also solve the problem

rohoet:
I think that if the Jenkinsfile is changed in a PR, Jenkins checks with GitHub whether the PR author has the permission to push to the origin repository, which is _trusted_ if yes, and _untrusted_ otherwise. For the latter (our case here), I guess the last trusted Jenkins file version is used causing the new submodules to be excluded, ultimately causing the includes to fail.

So lets see if the build passes when the PR is _trustfully_ merged into master.

Updated libs are already _filed_ for IP checks.

PR #55, opened 2020-09-17
Feature/better image compression

SebastianSchildt:
This switches docker image format to xz instead of bzip. Advantages

 - The compressors is multithreaded, so actually compression is faster during build
 - The images are smaller
 - xz also directly supported by docker load, so no disadvantages

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -34,7 +34,7 @@ node('docker') {
             '''
         }
         stage ('Archive') {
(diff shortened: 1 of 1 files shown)


wenwenchenbosch (approved):

SebastianSchildt:
Can be merged @rohoet , see succesful build here https://kuksaval.northeurope.cloudapp.azure.com/job/kuksaval-boschresearch/job/feature%252Fbetter-image-compression/

PR #62, opened 2020-10-30
Feature/56 s3 parquet uploader

### LINKED ISSUE #56: S3 Parquet Uploader
We need a configurable uploader that batches a number of datapoints over a time window into https://parquet.apache.org/ and is able to upload them to an S3 bucket.

This should be an external program using the websocket (or REST) IF to talk to KUKSA and not integrated into the server directly

wenwenchenbosch:
close #56


SebastianSchildt:
Needs workswith (this is just an example app, not part of the core server) CQ for pyarrow

https://pypi.org/project/pyarrow/
Version 2.0.0
License Apache License 2.0
https://github.com/apache/arrow/tree/apache-arrow-2.0.0

@wenwenchenbosch   any other dependencies outside of python standard lib?

wenwenchenbosch:
> Needs workswith (this is just an example app, not part of the core server) CQ for pyarrow
> 
> https://pypi.org/project/pyarrow/
> Version 2.0.0
> License Apache License 2.0
> https://github.com/apache/arrow/tree/apache-arrow-2.0.0
> 
> @wenwenchenbosch any other dependencies outside of python standard lib?

[boto3](https://github.com/boto/boto3)
Version:  latest (0.0.14)
Apache-2.0 License

[parquet-tools](https://pypi.org/project/parquet-tools)
Version: latest (0.2.4)
License: MIT License (MIT)

Is any possibility to just scan the [requirements.txt](https://github.com/eclipse/kuksa.val/blob/d57e340f1288e4da1d40680b3ecdc974ff682402/clients/s3/requirements.txt) and some automatisation instead of searching and comment such information?

wenwenchenbosch:
@SebastianSchildt  pls test with your AWS account, which should also work, when you configure your credential information

SebastianSchildt:
https://pypi.org/project/awswrangler/
Version: latest ( 1.10.0)  https://github.com/awslabs/aws-data-wrangler/tree/1.10.0
License:  Apache License 2.0

wenwenchenbosch:
> https://pypi.org/project/awswrangler/
> Version: latest ( 1.10.0) https://github.com/awslabs/aws-data-wrangler/tree/1.10.0
> License: Apache License 2.0

@SebastianSchildt this package is not necessary any more, I saw, there was still import there, I will remove it

wenwenchenbosch:
@SebastianSchildt: eclipse jenkins slave seems to be full
```
xz: (stdout): Write error: No space left on device
```

SebastianSchildt:
Fixed. But no rights to trigger build in Eclipse Jenkins. Ask @eriksven or push another commit

SebastianSchildt:
Seems to be working now!
Still need to do the CQs

clients/s3/requirements.txt
```diff
@@ -0,0 +1,3 @@
+boto3~=1.16.5
```
wenwenchenbosch, line 1: @SebastianSchildt  only these libraries are necessary

SebastianSchildt:
This breaks testclient, I think mainly when there is a timeout.

try this on a slow computer

```
sebastian@xbionic:~/kuksa.val/clients/vss-testclient$ pipenv run python testclient.py 
connect to wss://127.0.0.1:8090
Websocket connected!!
VSS Client> authorize ../../certificates/jwt/super-admin.json.token 
{
    "TTL": 1767225599, 
    "action": "authorize", 
    "requestId": "1238", 
    "timestamp": 1607087253
}

VSS Client> getValue Vehicle
```

(This getValaue provokes a large answer, if it doesn't fail immedaitely, repeat until you get

```
VSS Client> getValue Vehicle
EXCEPTION of type 'Empty' occurred with message: ''
To enable full traceback, run the following command: 'set debug true'
VSS Client> 
```

Now all answers are off by one:
```
VSS Client> getValue Vehicle.Speed
[...]
        "Vehicle.VehicleIdentification.vehicleConfiguration": "---", 
        "Vehicle.VehicleIdentification.vehicleModelDate": "---", 
        "Vehicle.VehicleIdentification.vehicleSeatingCapacity": "---", 
        "Vehicle.VehicleIdentification.vehicleSpecialUsage": "---", 
        "Vehicle.VehicleIdentification.vehicleinteriorColor": "---", 
        "Vehicle.VehicleIdentification.vehicleinteriorType": "---", 
        "Vehicle.accelerationTime": "---", 
        "Vehicle.cargoVolume": "---", 
        "Vehicle.emissionsCO2": "---"
    }
}

VSS Client> 
```

Now all replies are "off by one"

```
VSS Client> getValue Vehicle.OBD.TimingAdvance 
{
    "action": "get", 
    "path": "Vehicle.Speed", 
    "requestId": "1234", 
    "timestamp": 1607087525, 
    "value": "---"
}

VSS Client> getValue Vehicle.Speed
{
    "action": "get", 
    "path": "Vehicle.OBD.TimingAdvance", 
    "requestId": "1234", 
    "timestamp": 1607087540, 
    "value": "---"
}
```

I think the timeouts are not handled well, if a reply is received later. It always needs to read all replies and do a requestID matching, or only do unlimited  blocking waits

SebastianSchildt (changes requested):
Testclient broken

clients/common/clientComm.py
```diff
@@ -37,6 +37,58 @@ def stopComm(self):
...
+        req["requestId"] = 1235
+        req["action"]= "set"
+        req["path"] = path
+        req["value"] = value
+        return self.sendReceiveMsg(req, timeout)
+
+
+    def getValue(self, path, timeout = 5):
+        req = {}
+        req["requestId"] = 1234
```
SebastianSchildt, line 87: This means, it is still broken in testclient. To provoke it: Reduce timeout to a small value and try this:

```
pipenv run python testclient.py 
connect to wss://127.0.0.1:8090
Websocket connected!!
VSS Client> authorize ../../certificates/jwt/super-admin.json.token 
{
    "TTL": 1767225599, 
    "action": "authorize", 
    "requestId": "1238", 
    "timestamp": 1607118379
}

VSS Client> getValue Vehicle
{
  "requestId": 1234,
  "action": "get",
  "path": "Vehicle",
  "error": "timeout"
}


VSS Client> getValue Vehicle.Speed
[...]
        "Vehicle.VehicleIdentification.vehicleModelDate": "---", 
        "Vehicle.VehicleIdentification.vehicleSeatingCapacity": "---", 
        "Vehicle.VehicleIdentification.vehicleSpecialUsage": "---", 
        "Vehicle.VehicleIdentification.vehicleinteriorColor": "---", 
        "Vehicle.VehicleIdentification.vehicleinteriorType": "---", 
        "Vehicle.accelerationTime": "---", 
        "Vehicle.cargoVolume": "---", 
        "Vehicle.emissionsCO2": "---"
    }
}

VSS Client> 
```
So when the first one times out, the id comparison can not rescue you, because everything has the same id

clients/common/clientComm.py
```diff
@@ -37,6 +37,58 @@ def stopComm(self):
...
 
+
+    def sendReceiveMsg(self, req, timeout): 
+        jsonDump = json.dumps(req)
+        self.sendMsgQueue.put(jsonDump)
+        while True:
+            try:
+                resp = self.recvMsgQueue.get(timeout = timeout)
+                respJson =  json.loads(resp) 
+                if str(req["requestId"]) == str(respJson["requestId"]):
```
SebastianSchildt, line 48: I think this loop is clever, but does not help you, when all helper functions are using static ids (i.e. you may notice the difference between a set and get, but not between different sets

I suggest a globally counting id for all helpers (maybe safest), or a random number (shoudl besafe enough if large enough range) , or separately counting ones for each message type with diffeerent prefixes (not the most clever I think, because they might run into each other)

SebastianSchildt (changes requested):
Still broken 💔

clients/common/clientComm.py
wenwenchenbosch: you are right, but I think, it should be fixed in another PR so that the PR does not become more complicated. because the behavior currently is better than master, not worse. An there is some other things, that I also like to improve from test client, but not in this PR.

SebastianSchildt (approved):
Will merge, but we should fix the request id handling in common lib asap.
It is ugly and will bite us evenutally , if we forget it....

PR #66, opened 2020-11-05
Fix w3cunittest utests

phongt:
Base on the json key "datatype" respond the right value key type. This detected by some unit tests failure.

eg: w3cunittest.cpp: process_query_set_get_simple  the Vehicle.OBD.EngineSpeed data type is float as definition in test_vss_rel_2.0.json

### DIFF
--- a/test/unit-test/w3cunittest.cpp
+++ b/test/unit-test/w3cunittest.cpp
@@ -1113,7 +1113,7 @@ BOOST_AUTO_TEST_CASE(process_query_set_get_simple)
      )");
 
 
-   BOOST_TEST(set_response_json.has_key("timestamp") == true);
+   BOOST_TEST(set_response_json.contains("timestamp") == true);
    // remove timestamp to match
    set_response_json.erase("timestamp");
 
@@ -1141,7 +1141,7 @@ BOOST_AUTO_TEST_CASE(process_query_set_get_simple)
    json response_json = json::parse(response);
 
 
-   BOOST_TEST(response_json.has_key("timestamp") == true);
+   BOOST_TEST(response_json.contains("timestamp") == true);
 
    // remove timestamp to match
    response_json.erase("timestamp");
@@ -1184,7 +1184,7 @@ BOOST_AUTO_TEST_CASE(process_query_get_withwildcard)
 
 
 
-   BOOST_TEST(response_json.has_key("timestamp") == true);
+   BOOST_TEST(response_json.contains("timestamp") == true);
 
    // remove timestamp to match
    response_json.erase("timestamp");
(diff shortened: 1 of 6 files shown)


wenwenchenbosch (approved):

test/unit-test/w3cunittest.cpp
```diff
@@ -1135,7 +1135,7 @@ BOOST_AUTO_TEST_CASE(process_query_set_get_simple)
     "action": "get",
     "path": "Vehicle.OBD.EngineSpeed",
     "requestId": "8756",
-    "value": 2345
```
wenwenchenbosch, line 1138: @SebastianSchildt why sometime speed is flot sometime int32

SebastianSchildt: Is "I have no idea" a valid answer here? Vehicle.OBD.Speed is defined as float  and Vehicle.Speed as int32  in the VSS model (maybe be a question to take to Genivi :) ) . I am not sure, if maybe the rendering via our jsonlib changed somehow for floats with no decimal places... And I am not sure how clever our impl is when writing an int to a float data item or vice versa.....

SebastianSchildt: This:

image

doesn't inspire confidence...

SebastianSchildt:
Currently I get 

```
*** 39 failures are detected (6 failures are expected) in the test module "w3c-unit-test"
```

which is definitely an improvement. Is it the expected result?

Is the idea to get this merged or are more fixes coming in shortly?  (Getting greedy here 😁 )

phongt:
> Is the idea to get this merged or are more fixes coming in shortly? (Getting greedy here  )

Please don't merge now. need to clarify more regrading some datatype definition in VSS. Just check the [vss_rel_2.0.0-alpha+006.json](https://genivi.github.io/vehicle_signal_specification/releases/nightly/vss_rel_2.0.0-alpha+006.json). 

There are inconsistency points, eg: "datatype": "Double", "datatype":"Float", "datatype": "UInt16" vs "datatype": "uint16"...

phongt:
more fix in fb37b3d

*** 28 failures are detected (6 failures are expected) in the test module "w3c-unit-test"

SebastianSchildt:
Ah that seems true. I think we need to fix it in VSS, via the issue there (I will try to make a PR there, but not today). Next merge window in Genivi VSS is coming Tuesday 19:00 MET.  Also (as you have noticed I guess) the shipped json within kuksa.val is a bit old, the idea was to update it, once version 2 is released. So when fixing things on our end we should always refer to the current one at Genivi, as you did.

SebastianSchildt:
@phongt  Are you still planning to work on this in the forseeable future? Otherwise we might merge it, so the work isn't lost. (the unittest were broken before, and are broken after merging this. But less so, so still a net win 😄 )

phongt:
@SebastianSchildt  I have limited time for it now. Please check and merge it.

SebastianSchildt:
Thank for your contribution @phongt .Feel free to be back any time :)  I will try to fix (or disable) remaining test, so we can enable this in CI, so we at least do not regress here again

PR #79, opened 2020-12-04
buildx and simplified jenkins

SebastianSchildt:
This is an udpated version for our CI pipeline. Still need some cleanup in docs and removing `build.sh`files , but since it already builds all artifacts (and more)  than the old one, putting a PR.

Changes

 -  `buildx` for building this has several advantages: 
    - We can get rid of the `build.sh`scripts that did the Qemu injection for corss compiling, as this is done behind the scenes
    - it is much easier to add support for other arhcitectures later
    - can compile natively instead of Qemu, this isn't hard coded in Jenkinsfile, but depends on the setup of docker on the build machine. Already tested for arm64 via Pi 4 and a Amazon Graviton 2 instance
 - getting rid of the "outer" docker, as it didn't do anything. The idea was, to provide a repeatable  "custom" build environment from scratch, however as all our builds are via docker anyway, not much customisation required, so we can remove one layer.
 - also building vss-testclient as container image
 - parallel building of amd64 and arm64


Jenkinsfile
```diff
@@ -10,39 +10,44 @@
...
-        stage('Prepare') {
-        sh '''
-            git submodule update --init
-            mkdir -p artifacts
-            rm -f artifacts/*
-            '''
-        }
-        stage('Build') {
-			sh '''
-			 ./docker/build.sh amd64
```
wenwenchenbosch, line 24: should we remove the build.sh file, if it is not usable any more?

SebastianSchildt: yes, this is why I marked it draft. Still need doc adjustment and removal of some files

SebastianSchildt: Clean up done

SebastianSchildt:
Build is running, check here: https://kuksaval.northeurope.cloudapp.azure.com/job/kuksaval-boschresearch/job/feature%252Fbuildx-and-simplified-jenkins/ 
(for security reasons PR is build with old Jenkinsfile, that is not working, after merge, new one will be used)

README.md
```diff
@@ -23,16 +23,20 @@ Download a current docker image from our CI server:
...
+docker load -i kuksa-val-b3084b9-amd64.tar.xz 
 ```
 
+Your build tag may vary, and for ARM machines you need to choose an arm64 images.
+
 Prepare an empty directory `$HOME/kuksaval.config`.  Run the docker image using the tag reported by `docker load`:
 
 ```bash
-sudo docker run -it --rm -v $HOME/kuksaval.config:/config  -p 127.0.0.1:8090:8090 -e LOG_LEVEL=ALL amd64/kuksa-val:0.1.1
+docker run -it --rm -v $HOME/kuksaval.config:/config  -p 127.0.0.1:8090:8090 -e LOG_LEVEL=ALL amd64/kuksa-val:b3084b9
```
wenwenchenbosch, line 39: good improvement 👍

wenwenchenbosch (approved):

Issue #41, opened 2020-08-31
Deadlock caused by exceptions 


wenwenchenbosch:
Changing of databases are protected by mutex, see https://github.com/boschresearch/kuksa.val/blob/a696d9f0dac1faf84ab8da18a7dd8a932345a352/src/VssDatabase.cpp#L616

But jsoncons raises exceptions if sth. goes wrong. All exceptions are cached, but mutex is not released after sending `JsonResponses::malFormedRequest` response. Then, the vss server will go to a deadlock when a second request comes.

SebastianSchildt:
Is there any easy/generic way to make this more robust?

SebastianSchildt:
Will close for now, until it bites us again :)

PR #84, opened 2020-12-08
Fix/unittests

SebastianSchildt:
Trying to get unittests to run. Found some bugs in VSSDatabase.cpp (seriously this thing needs to be refactored).

For now I am just fixing things to "make them work". Clean up later

Whenever there are different assumptions between server and unittests I try to align.

Incidentally I fixed some occurrences of https://github.com/eclipse/kuksa.val/issues/41 , but this is still pending a major refactoring/better design.

Mid-term plan would be, uddating VSSCommandPRocessor/VSSDatabase codepaths one by one for different actiosn scuh as `set`and `get` maybe with an eye to VISS Gen. 2, so that we at least implement a compatible subset.

But for now.... Still working on unit tests....


SebastianSchildt:
Ready for review Jenkinsfile also changed, see working build here:  https://kuksaval.northeurope.cloudapp.azure.com/job/kuksaval-boschresearch/job/fix%252Funittests/

This was not aimed to de-uglify the code, just getting stuff running. Unittests already helped finding a bug in subscribe permission mgmt (fixed now)

src/VssDatabase.cpp
```diff
@@ -701,8 +753,22 @@ void VssDatabase::setSignal(WsChannel& channel,
...
+  try {
+      setValues = getPathForSet(path, valueJson);
+  }
+  catch( noPathFoundonTree& e) {
+    rwMutex_.unlock();
+    throw e;
+  }
+  catch ( genException &e) {
+    logger_->Log(LogLevel::ERROR, "Exception VssDatabase::setSignal: " + string(e.what()));
+    rwMutex_.unlock();
```
wenwenchenbosch, line 769: using https://en.cppreference.com/w/cpp/thread/lock_guard will solve the problem better, but yes, we can refactor it later

test/unit-test/AuthenticatorTests.cpp
```diff
@@ -144,49 +144,6 @@ BOOST_AUTO_TEST_CASE(Given_GoodToken_When_Validate_Shall_ValidateTokenSuccessful
   BOOST_TEST(res == std::chrono::time_point_cast<std::chrono::seconds>(exprTime).time_since_epoch().count());
 }
 
-BOOST_AUTO_TEST_CASE(Given_BadPathInToken_When_Validate_Shall_ThrowException)
```
wenwenchenbosch, line 147: why remove this unittest?

test/unit-test/KuksavalUnitTest.cpp
```diff
@@ -2391,56 +2397,6 @@ BOOST_AUTO_TEST_CASE(permission_basic_write_with_branch_permission)
    get_response_json.erase("timestamp");
    BOOST_TEST(get_response_json == get_expected);
 
-   // Now try to read with a valid token.
```
wenwenchenbosch, line 2394: why removed?

test/unit-test/AuthenticatorTests.cpp
SebastianSchildt: Because this one does not make sense, I couldn't gather the intention. It claims it is about "Bad path in tokens", but the paths are ok, and anyway, I think there is no such thing as a bad path in a token.

test/unit-test/KuksavalUnitTest.cpp
SebastianSchildt: According to the name `permission_basic_write_with_branch_permission` and the idoamtic way this tests are written, this testcase is done here. 
The rest seems unrelated (maybe copy paste residue? It is just a basic read with a basic Permission, but written in slightly different style. And there already is another separate testcase for that behavior

wenwenchenbosch (approved):

Issue #74, opened 2020-11-19
Update to newer Boost version


SebastianSchildt:
We are using a relatively old boost version.

We should update it. 

See also #72

PR #87, opened 2021-01-13
74 update boost

wenwenchenbosch:
solve #74 

update to boost 1.75


SebastianSchildt:
Seemed to work, but compiling unittests was super noisy, I used the chance to replace our turtle version with submodule and worked around this issue https://github.com/mat007/turtle/issues/98   in our unittests with some defines. Those are only used  for unittests (which still work fine), and should not cause any sideffects anywhere

Next question, why this still can not merge....

SebastianSchildt (approved):
Good for me, but pls check my changes

PR #91, opened 2021-01-20
Use shortuuid

wenwenchenbosch:
use short requestId in testclient, so that, output looks better

### DIFF
--- a/clients/common/clientComm.py
+++ b/clients/common/clientComm.py
@@ -11,7 +11,7 @@
 ########################################################################
 
 import os, sys, threading, queue, ssl, json
-import shortuuid
+import uuid
 import asyncio, websockets, pathlib
 
 class VSSClientComm(threading.Thread):
@@ -40,7 +40,7 @@ class VSSClientComm(threading.Thread):
 
 
     def sendReceiveMsg(self, req, timeout): 
-        req["requestId"] = shortuuid.uuid()
+        req["requestId"] = str(uuid.uuid4())
         jsonDump = json.dumps(req)
         self.sendMsgQueue.put(jsonDump)
         while True:
--- a/clients/common/clientComm.py
+++ b/clients/common/clientComm.py
@@ -11,7 +11,7 @@
 ########################################################################
 
 import os, sys, threading, queue, ssl, json
-import uuid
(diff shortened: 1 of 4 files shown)


SebastianSchildt:
Shorter uuids are nice, but maybe non-numeric IDs are dangerous for other implementation (I know for us it is string)? Any reason why not using "half" of normal uuid insteand of another dependency?

SebastianSchildt:
I checked: https://rawcdn.githack.com/w3c/automotive/64f3a1cfdd53ccb8411ccabb52e1cc39a7148d32/spec/VISSv2_Transport.html#dfn-requestid

RequestId is a string, but it states


> Unique id value specified by the client. Returned by the server in the response and used by the client to link the request and response messages. The value MAY be an integer or a Universally Unique Identifier (UUID).

So maybe we *should* use int or UUID?

SebastianSchildt:
How about

```
Python 3.7.0 (v3.7.0:1bf9cc5093, Jun 26 2018, 23:26:24) 
[Clang 6.0 (clang-600.0.57)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import uuid
>>> str(uuid.uuid4())
'4e6a630b-6738-4f81-947b-c985fe328aa9'
>>> 
```
Yeah long, but actually it is shorter than some of the numbers generated by master :) And easier to recognize due to the uuid standard formatting. And in line with the recommendation in spec

wenwenchenbosch:
> How about
> 
> ```
> Python 3.7.0 (v3.7.0:1bf9cc5093, Jun 26 2018, 23:26:24) 
> [Clang 6.0 (clang-600.0.57)] on darwin
> Type "help", "copyright", "credits" or "license" for more information.
> >>> import uuid
> >>> str(uuid.uuid4())
> '4e6a630b-6738-4f81-947b-c985fe328aa9'
> >>> 
> ```
> 
> Yeah long, but actually it is shorter than some of the numbers generated by master :) And easier to recognize due to the uuid standard formatting. And in line with the recommendation in spec

just 2 character shorter, I am think about to remove the "-"

SebastianSchildt (approved):

PR #95, opened 2021-01-22
Bugfix/41 fix deadlock caused by exceptions

wenwenchenbosch:
solve #41 
also fix the problem with --insecure flag, which can also be configured using a config file


SebastianSchildt (approved):
Looking good

PR #112, opened 2021-02-08
Cleanup

wenwenchenbosch:
just cleanup websocket client python code


SebastianSchildt:
Will check, but maybe @nayakned  can have a look/try to run this?

I guess this means testing test client still works (because common code changed) and getting the gpsd feeder to run?

wenwenchenbosch:
> I guess this means testing test client still works (because common code changed) and getting the gpsd feeder to run?

the master gpsd should also still work. really only cleanup to make code clean.

clients/feeder/gps2val/gpsd_feeder.py
```diff
@@ -36,52 +36,20 @@ def __init__(self, config):
...
-        resp = self.recvMsgQueue.get(timeout = 1)
-        #print(resp)
-        
     def setPosition(self, position):
-        self.setValue('Vehicle.Cabin.Infotainment.Navigation.CurrentLocation.Altitude', position['alt'])
-        self.setValue('Vehicle.Cabin.Infotainment.Navigation.CurrentLocation.Latitude', position["lat"])
-        self.setValue('Vehicle.Cabin.Infotainment.Navigation.CurrentLocation.Longitude', position["lon"])
-        self.setValue('Vehicle.Cabin.Infotainment.Navigation.CurrentLocation.Accuracy', position["hdop"])
-        self.setValue('Vehicle.Cabin.Infotainment.Navigation.CurrentLocation.Speed', position["speed"])
+        self.client.setValue('Vehicle.Cabin.Infotainment.Navigation.CurrentLocation.Altitude', position['alt'])
```
SebastianSchildt, line 48: This will need to be changed soon, as the current positon is deprecated and will move to "Vehicle.Location"

nayakned:
Since we are taking about cleanup, perhaps lets remove this line.

import configparser

clients/feeder/gps2val/gpsd_feeder.py
wenwenchenbosch: but not for now right?

SebastianSchildt: for now it works. For VSS 2.0 also. But for releases after 2.1 we need to change that. Just to keep in mind

PR #113, opened 2021-02-08
Feature/refactor set

SebastianSchildt:
This is work in progress. 

Replace old set code with streamlined code supporting "/" paths.

Advanced features such as multiset in a later step, when migrating msg semantic to Gen2


test/unit-test/VssDatabaseTests.cpp
```diff
@@ -148,22 +148,22 @@ BOOST_AUTO_TEST_CASE(Given_ValidVssFilenameAndChannelAuthorized_When_GetBranch_S
...
   channel.setAuthorized(true);
 
   // expectations
 
-  MOCK_EXPECT(accCheckMock->checkReadDeprecated)
+  MOCK_EXPECT(accCheckMock->checkReadAccess)
     .returns(true);
 
-  std::string expectedJsonString{R"({"value":{"Vehicle.Acceleration.Lateral":"---","Vehicle.Acceleration.Longitudinal":"---","Vehicle.Acceleration.Vertical":"---"}})"};
+  std::string expectedJsonString{R"({"value":[{"Vehicle.Acceleration.Vertical":"---"},{"Vehicle.Acceleration.Longitudinal":"---"},{"Vehicle.Acceleration.Lateral":"---"}]})"};
```
wenwenchenbosch, line 161: pls document such api changes

src/VssDatabase.cpp
```diff
@@ -826,198 +778,72 @@ jsoncons::json VssDatabase::getPathForSet(const string &path, jsoncons::json val
...
+      if (resJson.contains("datatype")) {
+        string value_type = resJson["datatype"].as<string>();
+        checkTypeAndBound(logger_, value_type, value);
+        resJson.insert_or_assign("value", value);
+        {
+          jsonpath::json_replace(data_tree__, path.getJSONPath(), resJson);
+        }
+      }
+      else {
+        throw genException(path.getVSSPath()+ "is invalid for set"); //Todo better error message. (Does not propagate);
```
wenwenchenbosch, line 808: may be, does not contain field datatype

src/VssDatabase.cpp
```diff
@@ -826,198 +778,72 @@ jsoncons::json VssDatabase::getPathForSet(const string &path, jsoncons::json val
...
+
+  return answer;
 }
 
+//Only needed for DBUS currently- 
 // Method for setting values to signals.
-void VssDatabase::setSignal(const string &path,
-                            jsoncons::json valueJson) {
-  if (path == "") {
+void VssDatabase::setSignal(const string &dbuspath,
```
wenwenchenbosch, line 818: path here is really dbus specific? Comment is good, but change the name here is wired.

src/VssCommandProcessor.cpp
```diff
@@ -65,82 +65,7 @@ VssCommandProcessor::~VssCommandProcessor() {
...
-    res["action"] = "get";
-    res["requestId"] = request_id;
-    res["timestamp"] = JsonResponses::getTimeStamp();
-    stringstream ss;
-    ss << pretty_print(res);
-    return ss.str();
-  }
-}
-
-string VssCommandProcessor::processSet(WsChannel &channel,
```
wenwenchenbosch, line 95: where is the code moving? not necessary any more? but https://github.com/eclipse/kuksa.val/blob/7ddad8da5c3f208592f64d5e31041cafcc6d898d/include/VssCommandProcessor.hpp#L46 declaration is still there

wenwenchenbosch (commented):
I like, that many code is removed. I will test the code now

wenwenchenbosch:
Issue 1:

```
VSS Client> setValue Vehicle.Speed 200
{
    "action": "set", 
    "requestId": "0bc7f8af-7e9d-46c9-8de3-21d778b1d168", 
    "timestamp": 1613404411366
}

VSS Client> setValue Vehicle.Speed2 200
{
    "action": "set", 
    "requestId": "02eeebf2-f6fe-44f2-aefb-4dbd8ba30493", 
    "timestamp": 1613404413198
}

VSS Client> setValue bla 200
{
    "action": "set", 
    "requestId": "0a8883b9-546b-4e59-8ed6-4871ca7fb4e4", 
    "timestamp": 1613404429551
}

```

wenwenchenbosch:
expected behavior:

```
VSS Client> setValue Vehicle.Speed2 200
{
    "action": "set", 
    "error": {
        "message": "I can not find Vehicle.Speed2 in my db", 
        "number": 404, 
        "reason": "Path not found"
    }, 
    "requestId": "63ca27e8-1afe-457c-a9cf-56c2d246c38c", 
    "timestamp": 1613404486744
}
```

wenwenchenbosch:
issue 2: mqtt publisher does not work any more :(

src/VssDatabase.cpp
SebastianSchildt: This function is only used in DBUS code path. Options where removing it,with complete Dbus support, leaving it for the time being, or upgrading DBUS code to reuse websocket code paths (which may be useful one day, but currrently I do not know, what that code is doing, and if it is really functional

src/VssCommandProcessor.cpp
SebastianSchildt: Good catch. Not needed (I hope), superseded by the VSSPath variant

test/unit-test/VssDatabaseTests.cpp
SebastianSchildt: As the original behaviour was not documented, and the V1 spec does not specify this, I suggest not documenting the "change" but rather provide an exhaustive documentation of the expected valid queries and responses, as we have done for "get" (and yes, that is still missing)

src/VssDatabase.cpp
SebastianSchildt: better message for set know (certain checks done beforehand), still need to beautify/simplify the VssDatabse code nevertheless

SebastianSchildt:
Now non-existing paths can not be set anymore. Fresh unittestests still missing....

src/VssCommandSet.cpp
```diff
@@ -28,25 +34,57 @@ std::string VssCommandProcessor::processSet2(WsChannel &channel,
...
+      msg << "Can not set " << std::get<0>(setTuple).getVSSPath() << ". Only sensor or actor leaves can be set.";
+      logger->Log(LogLevel::VERBOSE,msg.str());
+      return JsonResponses::noAccess(request["requestId"].as<string>(), "set", msg.str());
+    } 
+  }
+
+
   try {
-    database->setSignal(channel, path, request["value"], gen1_compat_mode);
+    for ( std::tuple<VSSPath,jsoncons::json> setTuple : setPairs) {
```
wenwenchenbosch, line 85: @SebastianSchildt You looped `setPairs` three times!

test/unit-test/Gen2GetTests.cpp
```diff
@@ -133,6 +132,86 @@ BOOST_AUTO_TEST_CASE(Gen2_Get_Sensor) {
...
+    "number": 400,
+    "reason": "Bad Request"
+  },
+  "requestId": "UNKNOWN",
+  "timestamp": 0
+}
+      )"};
+  jsoncons::json expectedJson = jsoncons::json::parse(expectedJsonString);
+
+  // run UUT
```
wenwenchenbosch, line 202: what is UUT?

test/unit-test/KuksavalUnitTest.cpp
```diff
@@ -2277,13 +2281,13 @@ BOOST_AUTO_TEST_CASE(permission_basic_write_with_wildcard_permission)
...
+   string AUTH_TOKEN = "eyJhbGciOiJSUzI1NiIsInR5cCI6IkpXVCJ9.eyJzdWIiOiJFeGFtcGxlIEpXVCIsImlzcyI6IkVjbGlwc2Uga3Vrc2EiLCJhZG1pbiI6dHJ1ZSwiaWF0IjoxNTE2MjM5MDIyLCJleHAiOjE3MDkzNzI4MDAsImt1a3NhLXZzcyI6eyJWZWhpY2xlLk9CRC4qIjoicncifX0.U4zpy38DBTx2thPWfOMKALfZovJdkr5xQWDQaGSQo6sMhG-MgDz2UZDvVyRllHiv76f2VmZ2X0TnUZ_Z6yFUmjVYq83LGzUjSDY-xFLLRqOPvFKsQbTmOLbT629uU5Pz2m6x1HcxHBVCxNvI3pZw1D_rHODkVVlgSjycDK5NaHi9EdbSDmccmwk3jI6tPag0eKC7adzBlqkt-h6vqlx_FRz7kjyhiLXmR95QIV3_0MIyq484SsU4ztMPgmHBn1VUWz0x5EW065J2BIavXpaF20oNek0ZzsGcDq-UEQOLPDFoAnYykFrcYmIlXc51wRU2Mw-IIIIs4CV9-Vp1GkQfhQCicmR3OprRepzi_C0Fdh6kwSetAYJWqlVOeo1oUIxFIYbxGFjtdG38dOaacDddBnJ93K8JlnbFdp79Ins39HCmRPit198vdf7gLRcM_GLtvGF6SbT0V2tn17CNt2UxLEniHp7hGybxBVKqgTws2H73ZLe4HEq3UU8vhbeg2V2KXeSXvDJeE3LZvDWARafyHVXrkOx8-O9HqIClfAr-LAl7nnBqq2f9DneW1hjFZOFV5RUCrJ_UpZR_hyEU_aroXO6MNw_MC6KvZeOC6ti1NO6uAB3BYHzpNP1mlyUHlfRl-oLuKZFFqFVBrxY3tBrQwgzfhPZwi427ZIwrNlfv-aU";
```
wenwenchenbosch, line 2290: why this change necessary?

SebastianSchildt: That was an old token (see https://jwt.io/), the old code just never checked it (there are ealier and additional access checks now, in a place that is not mocked in this test unit. as the test already provided a token (albeit it was irrelevant) I replaced it with a validone, so the test goes through)

test/unit-test/Gen2GetTests.cpp
SebastianSchildt: I took the style form the existing tests, i assume "Unit under test"

src/VssCommandSet.cpp
SebastianSchildt: I was (that is what I told myself) going for readability instead of performance and deeper nesting. I hope the datastructure with the vector of tuples thing access compiles down to just moving a pointer, but I do not know 😕 
Maybe we add an enhancement issue to look into that later?

wenwenchenbosch: you are right with readability. But in this case, I do not think, that putting everything inside one loop is less readable than using 3 loops. If you like to improve the readability, better refact the method with multiple functions. It is always best practices to have smaller function/method, better than a function with more than 100 lines of code. But If you like, you can create an issue for this, instead of fixing it now.

SebastianSchildt: #124

wenwenchenbosch (approved):

PR #118, opened 2021-02-22
Feature/94 daprclients

wenwenchenbosch:
implement pubsub example. pls use readme.md to try it out.


anntexx:
Hi,
I am currently testing the PR and have some trouble to execute the DAPR example.
Especially, the DAPR publisher where it seems that there is a syntax error in this file (see attached screenshot).
My setup: Python3.9.2, DAPR 1.0.1 and PR commit: commit f06c92ac48d11d27910201a93bc35b6723c13520

Screenshot from 2021-03-23 14-53-01

anntexx:
updating pip3 to point to latest version of python3 installed did the trick to install the dependencies.
Now, I am able to start the DAPR executables, however, the DAPR subscriber does not get any data from the publisher.
Do I missed something to be executed? I see some messages from DAPR publisher to kuksa.val but no message appears at subscriber side (figure top right window).
Screenshot from 2021-03-23 15-44-00

anntexx:
using a virtual env solved the problem. I was able to verify the PR for both pub/sub and state-store.
So I verify the PR as ready to be merged

SebastianSchildt (approved):

PR #123, opened 2021-03-08
improve cmake ctest

wenwenchenbosch:
default `ctest` should show real results and result code. Only for jenkins, it is better to not return errer when some testcases failed, sothat jenkins can continue with generating test reports. Therefore add `TEST_RESULT_CODE` option for cmake


SebastianSchildt (approved):
Mostly harmless.
👍

Issue #128, opened 2021-03-18
docker images for kuksa.val are broken


callalilychen:
kuksa.val images can not be startet any more. see
https://github.com/eclipse/kuksa.val/pull/121#pullrequestreview-614703059

SebastianSchildt:
FYI @mwagner1886 
It is a regression. Working on it.

As a workaorund with newer images try setting an mqtt topic prefix like so:

```
docker  run -it --rm -v $HOME/kuksaval.config:/config  -p 127.0.0.1:8090:8090 -e LOG_LEVEL=ALL  -e KUKSAVAL_OPTARGS="--mqtt.topic-prefix kuksa"  amd64/kuksa-val:test
```

wenwenchenbosch:
found an other issue:
```
docker run -it --rm   -p 127.0.0.1:8090:8090 -e LOG_LEVEL=ALL <kuksa.val image> 
Starting kuksa.val
No VSS tree, initilaize with example
cp: can't create '/config/': Is a directory
No server keys configured, initialize with example
No jwt key configured, initialize with example
kuksa.val server
Commit b163c47-dirty from 2021-03-18T11:57:49+01:00
Read configs from config.ini
Log START
VERBOSE: Try reading JWT pub key from /config/certs/jwt.key.pub
VERBOSE: SubscribeThread: Started Subscription Thread!
ERROR: Exception occured while initializing database/ tree structure. Probably the init json file not found!Unexpected end of file at line 1 and column 1
terminate called after throwing an instance of 'std::exception'
  what():  std::exception
Aborted (core dumped)
```

SebastianSchildt:
This happens, when not mounting a config volume (which is described in the Readme). Should however fix this, so it can also run with default config without any external volumes

PR #129, opened 2021-03-18
fix default version in topic prefix

wenwenchenbosch:
solve #128


wenwenchenbosch:
@SebastianSchildt Do you have any idea to have better error messages for boost options so such errors?

SebastianSchildt (approved):
Tested, this fixes #128

PR #130, opened 2021-03-18
128 fix start script in docker images

wenwenchenbosch:
solve #128 
now i can also start docker images without giving config


SebastianSchildt (changes requested):
Will fail with existing VSS file, because than cert folder will not be created

Need to move up as first command

SebastianSchildt:
I mean, then wie only need `mkdir -p`once

Issue #46, opened 2020-09-03
setValue does not check min/max


wenwenchenbosch:
Currently, I have the following observation: 

```
VSS Client> getMetaData Vehicle.Speed
{
    "action": "getMetaData",
    "metadata": {
        "Vehicle": {
            "children": {
                "Speed": {
                    "datatype": "int32",
                    "description": "Vehicle speed, as sensed by the gearbox.",
                    "max": 250,
                    "min": -250,
                    "type": "sensor",
                    "unit": "km/h",
                    "uuid": "1efc9a11943b5a15ac159786051c5836"
                }
            },
            "description": "High-level vehicle data.",
            "type": "branch",
            "uuid": "1c72453e738511e9b29ad46a6a4b77e9"
        }
    },
    "requestId": "1236",
    "timestamp": 1599086861
}
VSS Client> setValue Vehicle.Speed 366
{
    "action": "set",
    "requestId": "1235",
    "timestamp": 1599087046
}
VSS Client> getValue Vehicle.Speed 
{
    "action": "get",
    "path": "Vehicle.Speed",
    "requestId": "1234",
    "timestamp": 1599087196,
    "value": "366"
}
```
But I do not expect the `setValue  Vehicle.Speed 366` can be successful, because the defined max value is 250.

SebastianSchildt:
The observation is correct, currently the min, max are ignored. I am not sure what we want here in the short term, as not all limit(as in speed) in the core data model make sense. If it is a "general" limitation for all land vehicles it shoudl be more like 600, but how useful it is? Mire useful would be the max of the current vehicle, but that would be a matter of a concrete deployment of the data model.

We might track what Genivi is doing here, a related discussion is https://github.com/GENIVI/vehicle_signal_specification/issues/166

But let's keep this open, because we should do "something" with min/max eventually

PR #134, opened 2021-03-19
Correctly type check boolean values

### LINKED ISSUE #133: Boolean values cannot be updated by specifying a "real" boolean value
Description:
Sensor and actuator values having the type "boolean" cannot be updated using "real" boolean values (true, false)

Environment:
- Error occurs on image build #55 (https://ci.eclipse.org/kuksa/job/kuksa.val/job/master/55/)

Steps to reproduce:
- Try to publish a sensor value of type "boolean" by transferring "real" booleans like true or false as value
  - An error response is sent back from Kuksa.VAL: "string or number is expected, received bool"
- If 0 or 1 is used, the sensor value can be successfully updated

Expected behavior:
- true and false are accepted, numbers and strings are rejected

SebastianSchildt:
Should fix #133 

```json
Test Client> setValue Vehicle.OBD.DriveCycleStatus.MIL foo
{
    "action": "set", 
    "error": {
        "message": "foo is not a bool. Valid values are true and false ", 
        "number": 400, 
        "reason": "Value passed is out of bounds"
    }, 
    "requestId": "9dfe6e60-3374-493a-aca8-30736a82d6f3", 
    "timestamp": 1616169129422
}

Test Client> setValue Vehicle.OBD.DriveCycleStatus.MIL true
{
    "action": "set", 
    "requestId": "a10e1ee1-9732-4704-ad28-e9c614d00501", 
    "timestamp": 1616169134986
}

Test Client> getValue Vehicle.OBD.DriveCycleStatus.MIL 
{
    "action": "get", 
    "path": "Vehicle.OBD.DriveCycleStatus.MIL", 
    "requestId": "2d6d790f-8c48-40fb-bb30-a22f2da76e01", 
    "timestamp": 1616169138448, 
    "value": true
}

Test Client> 
```

I guess the whole typeCheck code is due for a refactoring when we tacke #46

### DIFF
--- a/src/VssDatabase.cpp
+++ b/src/VssDatabase.cpp
@@ -65,7 +65,7 @@ jsoncons::json tryParse(jsoncons::json val) {
 }
 
     // Check the value type and if the value is within the range
-  void checkTypeAndBound(std::shared_ptr<ILogger> logger, string value_type, jsoncons::json val) {
+  void checkTypeAndBound(std::shared_ptr<ILogger> logger, string value_type, jsoncons::json &val) {
     bool typeValid = false;
 
     boost::algorithm::to_lower(value_type);
@@ -171,7 +171,6 @@ jsoncons::json tryParse(jsoncons::json val) {
         msg << "The type " << value_type << " with value " << val.as<float>()
             << " is out of bound";
         logger->Log(LogLevel::ERROR, "VssDatabase::setSignal: " + msg.str());
-
         throw outOfBoundException(msg.str());
       }
     }  else if (value_type == "float") {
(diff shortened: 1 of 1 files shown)


src/VssDatabase.cpp
```diff
@@ -203,6 +202,22 @@ jsoncons::json tryParse(jsoncons::json val) {
         throw outOfBoundException(msg.str());
       }
     } else if (value_type == "boolean") {
+      string v=val.as<string>();
+      boost::algorithm::erase_all(v, "\"");
+
+      if ( v == "true") {
```
wenwenchenbosch, line 208: would we like to also support "True", "TRUE"?

src/VssDatabase.cpp
```diff
@@ -65,7 +65,7 @@ jsoncons::json tryParse(jsoncons::json val) {
 }
 
     // Check the value type and if the value is within the range
-  void checkTypeAndBound(std::shared_ptr<ILogger> logger, string value_type, jsoncons::json val) {
+  void checkTypeAndBound(std::shared_ptr<ILogger> logger, string value_type, jsoncons::json &val) {
```
wenwenchenbosch, line 68:
```suggestion
  void checkTypeAndBound(std::shared_ptr<ILogger> logger, const string& value_type, const jsoncons::json &val) {
```

src/VssDatabase.cpp
SebastianSchildt: I checked and according to https://www.w3schools.com/js/js_json_datatypes.asp   only "true" and "false" in JSON, so I elft it at that. (and made the error message explicitly mentioning that)

wenwenchenbosch (approved):

PR #140, opened 2021-03-23
use correct stop function now

wenwenchenbosch:
I just see an error while stopping dapr apps, because of api changed last time


SebastianSchildt:
Looks fine for me @anntexx  can you quickly spin it up again, as you already have everything set tup

anntexx:
I checked the PR, it worked for both client apps pub/sub and state-store: See my output of the pub/sub scenarios
Screenshot from 2021-04-09 11-04-29

PR #143, opened 2021-03-29
141 test client version number

wenwenchenbosch:
solve #143


SebastianSchildt:
I would suggest removing the logo, as it is very large, an probably a little annyoing if you see it every time starting the testclient, especially since it sort of defeats the purpose of  seeing the verison number,as it easily scrolls out of view
grafik

(or it can somehwo be made MUCH smaller, (i.e. no "KUKSA" text and version beside the "graphic")

wenwenchenbosch:
@SebastianSchildt  now the logo should be smaller.

SebastianSchildt (approved):
Still not a fan of the logo, but want the feature anyway 🍰

wenwenchenbosch:
@SebastianSchildt also no merge commit? no merge commit very fatal

Issue #154, opened 2021-04-06
GPSD feeder can not be killed with strg-c


SebastianSchildt:
When starting GPSD feeder in docker, it can not be stopped via Strg-C

wenwenchenbosch:
actually the pull request for the dockerfile https://github.com/eclipse/kuksa.val/pull/147 is still open

wenwenchenbosch:
ok I find out the problem. gspd blocks, if no connection via udp there. I will have look at it and fix it direkt in the pull request above

wenwenchenbosch:
I reported a bug to gps library: see https://gitlab.com/gpsd/gpsd/-/issues/131

PR #156, opened 2021-04-06
using gpsd.waiting() to ask if any data there

wenwenchenbosch:
solve #154 

Thanks the great support from gpsd, I find now the function to avoid blocking call.


SebastianSchildt (approved):
Works 👍

Issue #149, opened 2021-03-31
Check enumerations


SebastianSchildt:
Some string values are limited via enums, We might want to check that to during set

PR #172, opened 2021-04-19
149 check enumerations

wenwenchenbosch:
solve #149


src/VSSTypeSanitizer.cpp
```diff
@@ -113,11 +126,16 @@ void VssDatabase::checkAndSanitizeType(jsoncons::json &meta, jsoncons::json &val
     else if (dt == "double") {
         checkNumTypes<double>(meta,val);
     }
+    else if (dt == "enum") {
```
SebastianSchildt, line 129: When will this ever happen? Can you point to an example in VSS?

wenwenchenbosch: good catch, I forget to remove it

SebastianSchildt (approved):

Issue #174, opened 2021-04-20
better output for limitcheck


wenwenchenbosch:
I got the following error message for Integer overflow:
```
Test Client> setValue Vehicle.Drivetrain.Transmission.Gear 130
{
    "action": "set", 
    "error": {
        "message": "Value \"130\" can not be converted to neccessary type:Dynamic exception type: jsoncons::json_runtime_error<std::runtime_error, void>\nstd::exception::what: Integer overflow\n", 
        "number": 400, 
        "reason": "Value passed is out of bounds"
    }, 
    "requestId": "9c4b4faf-a81c-495d-ad6c-bef8b2d37ab4", 
    "timestamp": 1618922543530
}
```

Expected output:
```
Test Client> setValue Vehicle.Drivetrain.Transmission.Gear 130
{
    "action": "set", 
    "error": {
        "message": "Value \"130\" can not be converted to defined type int8. Integer overflow", 
        "number": 400, 
        "reason": "Value passed is out of bounds"
    }, 
    "requestId": "9c4b4faf-a81c-495d-ad6c-bef8b2d37ab4", 
    "timestamp": 1618922543530
}
```

PR #176, opened 2021-04-20
improve error message for integer overflow

wenwenchenbosch:
solve #174

### DIFF
--- a/src/VSSTypeSanitizer.cpp
+++ b/src/VSSTypeSanitizer.cpp
@@ -37,14 +37,14 @@ void checkNumTypes(jsoncons::json &meta, jsoncons::json &val )
     try {
         cval = val.as<T>();
     }
-    catch(...) {
+    catch(std::exception const& e) {
         std::stringstream msg;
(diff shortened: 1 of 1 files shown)


src/VSSTypeSanitizer.cpp
```diff
@@ -37,14 +37,14 @@ void checkNumTypes(jsoncons::json &meta, jsoncons::json &val )
     try {
         cval = val.as<T>();
     }
-    catch(...) {
```
SebastianSchildt, line 40: Is it clever to change this? I am not sure what kind of exceptions  are thrown by the underlying jsoncons (and other standard libs it uses on different OS), and wanted to be sure here, even if it is bad style :)

Theoretically the change can miss excpetions, that are not derived std::exception
https://stackoverflow.com/questions/315948/c-catching-all-exceptions/24142104

Maybe the risk is small, but it was a conscious discussion to put "..."

Opinions?

wenwenchenbosch: I will also add this catch

SebastianSchildt (approved):
Nice.

Issue #160, opened 2021-04-08
Change timestamp format


SebastianSchildt:
Except for tokens VISS2 demands ISO8601 timestamps 

> Timestamps in transport payloads MUST conform to the ISO8601 standard, using the UTC format with a trailing Z. Time resolution SHALL at least be seconds, with subsecond resolution as an optional degree of precision when desired. The time and date format shall be as shown below, where the sub-second data and delimiter is optional.
YYYY-MM-DDTHH:MM:SS.ssssssZ
The exception to this is timestamps within tokens, which MUST conform to Unix time.

Issue #158, opened 2021-04-08
Change JSON format for data


SebastianSchildt:
In VISS 2 data is not in a simple item "value", but in a structure that allwos returning several data points

```json
 "data": {
            "description": "Data including path and one or more data point(s).",
            "type": "object/array",
            "properties": {
                "path": {
                    "description": "The path to the desired vehicle signal(s), as defined by the metadata schema.",
                    "type": "string"
                },
                "dp": {
                    "description": "Data point including one or more value and time samp",
                    "type": "object/array"                   
                    "properties": {
                        "value": {
                            "description": "The value related to the associated path.",
                            "type": "string"                    
                        },
                        "ts": {
                            "description": "Timestamp associated with the value capture.",
                            "type": "string"                    
                        }
                }
            }   
        },
```

SebastianSchildt:
See also https://raw.githack.com/w3c/automotive/gh-pages/spec/VISSv2_Core.html#multiple-signals-request

wenwenchenbosch:
according to the specification, response should look like this:
```
“data”:{“path”:”X”, “dp”:{“value”:”Y”, “ts”:”Z”}}
or
“data”:{“path”:”X”, “dp”: [{“value”:”Y1”, “ts”:”Z1”}, …, {“value”:”Yn”, “ts”:”Zn”}]}
or
“data”:[{“path”:”X1”, “dp”:{“value”:”Y1”, “ts”:”Z1”}}, …, {“path”:”Xm”, “dp”:{“value”:”Ym”, “ts”:”Zm”}}]
or
“data”:[{“path”:”X1”, “dp”:[{“value”:”Y11”, “ts”:”Z11”}, …, {“value”:”Y1n”, “ts”:”Z1n”}]}, …]
```
@SebastianSchildt: do we really need `description`, `type` and `properties`?

SebastianSchildt:
I I do not quite get the question. Your examples are correct and they should all be in line with the JSON schema above. A single  object or an array of dp where value and ts are mandatory properties.

You can also test here: https://www.jsonschemavalidator.net  if you find a mismatch between schema and the examples I would consider this a bug in VSS

PR #178, opened 2021-04-22
Fix/timestamp update

### LINKED ISSUE #159: Only update timestamps on set
The timestamp on "get" should not be the current time, but the last time it was set

SebastianSchildt:
Makes sure timestamps in "get" are the same as last "set" (and not from current time)

Fixes #159 

Some notes

 - I already added the new timestamps as "string", currently we have not tools that complain about this, and once we do #160 to be VISS compliant it is string anyway
 - There is no "good" solution when querying a complete branch. I just `randomly` put the timestamp of the last item that is found in the tree. It might be better to use the oldest or newest timestamp (both can be argued), of a set of elements, but once we do #158  it is not relevant anyway, and it is no regression as the old behavior also did not provide any useful information


wenwenchenbosch:
@SebastianSchildt 

recommendation:

```
Test Client> getValue Vehicle.OBD
{
    "action": "get", 
    "requestId": "28256eb9-8f78-4237-ad70-cedecfed1760", 
    "timestamp": 1619102925190, 
    "value": [
        {
            "path": "Vehicle.OBD.AbsoluteLoad", 
            "value": "---"
        },
... 
```

to 

```
Test Client> getValue Vehicle.OBD
{
    "action": "get", 
    "requestId": "28256eb9-8f78-4237-ad70-cedecfed1760", 
    "value": [
        {
            "path": "Vehicle.OBD.AbsoluteLoad", 
            "value": "---",
           "timestamp": 1619102925190
        },
...

**Comment by wenwenchenbosch** [comment]
just see, that getMetaData has also a timestamp
https://github.com/eclipse/kuksa.val/blob/master/src/VssCommandProcessor.cpp#L225

Maybe we need in a more cleanform, what is the meaning of timestamp

**Comment by SebastianSchildt** [comment]
As mentioned in my description above I do not want to change behaviour of stuff that goes out any way (#158) . As there won't be any "value" properties anymore, I don't think we should change them (because then they would neither VISS1 NOR VISS2 comapliant)

**Comment by SebastianSchildt** [comment]
Timestamp definition here https://raw.githack.com/w3c/automotive/gh-pages/spec/VISSv2_Core.html#timestamp  and here https://raw.githack.com/w3c/automotive/gh-pages/spec/VISSv2_Core.html#read "The sample time of the value"   Wrt to the "getMeta", it is not relevant anymore, as I think it will move "inside" data properties (as you suggested for values), so it won't have nay timestamp I guess

**Comment by wenwenchenbosch** [comment]
another idea:
Test Client> getValue Vehicle.OBD
{
    "action": "get", 
    "requestId": "28256eb9-8f78-4237-ad70-cedecfed1760", 
    "value": [
        {
            "path": "Vehicle.OBD.AbsoluteLoad", 
            "value": "---"
        },
    ...
   ],
    "timestamp": [
       "path": "Vehicle.OBD.AbsoluteLoad", 
        "value": 1619102925190
   ...
   ],
}
```

src/VssDatabase.cpp
```diff
@@ -580,6 +588,15 @@ jsoncons::json VssDatabase::getSignal(class WsChannel& channel, const VSSPath& p
...
       }
       valueArray.insert(valueArray.array_range().end(), value);
+      //This will add the "last"  timestamp, changing behavior from previous
+      //"timestamp of the get request" approach 
+      //Both are not very helpful when querying multiple values.
+      //This will be fixed once https://github.com/eclipse/kuksa.val/issues/158
+      //is implemented, as VISS2 will allow attaching individual timestamps to
+      //individual data
+      if (result.contains("timestamp")) {
+        answer["timestamp"] = result["timestamp"].as<string>();
```
wenwenchenbosch, line 598: if we will fix it for #158, why do we need this temp solution? It is totally confused to use the timestamp of the last element!

SebastianSchildt: The VISS1 standard demands a timestamp in answers, see the Schema in https://www.w3.org/TR/vehicle-information-service/

Before we had it as "timestamp of get request" (useless but syntactically compliant), now we have he timestamp of "one of" the requested data (also not that useful, but still compliant), Where is is "better", might when querying "ligically connected" data such a GPS position where a whole subtree will be updated at the same time. 
 
Removing timestamp would not be protocol compliant.

Because this problem has been recognized VISS2 change the behaviour, attaching timestamp to data. But that is a larger change, hence a separate ticket

src/VssDatabase.cpp
```diff
@@ -580,6 +588,15 @@ jsoncons::json VssDatabase::getSignal(class WsChannel& channel, const VSSPath& p
         value[spath] = "---";
       }
       valueArray.insert(valueArray.array_range().end(), value);
+      //This will add the "last"  timestamp, changing behavior from previous
```
wenwenchenbosch, line 591:
```suggestion
      // TODO: This will add the "last"  timestamp, changing behavior from previous
```

src/VssDatabase.cpp
wenwenchenbosch: pls add openpoints to the issue or create a new issue

wenwenchenbosch (approved):

Issue #182, opened 2021-04-23
Mix of gen1 & gen2 path format in error message


wenwenchenbosch:
```
Test Client> connect 
connect to wss://127.0.0.1:8090
Websocket connected!!
Test Client> setValue Vehicle.Speed 200
{
    "action": "set", 
    "error": {
        "message": "No write  access to Vehicle/Speed", 
        "number": 403, 
        "reason": "Forbidden"
    }, 
    "requestId": "4049604a-9c96-4483-b127-7e244f6debf9", 
    "timestamp": 1619164327584
}
```

wenwenchenbosch:
testcases error:
https://github.com/eclipse/kuksa.val/blob/761420934fd7cf3497032565a21c937f35c7c0d3/test/unit-test/KuksavalUnitTest.cpp#L2121-L2136

SebastianSchildt:
@wenwenchenbosch  I assume this can be closed for now, as we should have repaired most (hopefully all) occurences of this with merged PR?

wenwenchenbosch:
yes

PR #183, opened 2021-04-23
correct vsspath parser

wenwenchenbosch:
solve #182


include/VSSPath.hpp
```diff
@@ -29,17 +29,23 @@ class VSSPath {
         std::string getVSSPath() const;
         std::string getVSSGen1Path() const;
         std::string getJSONPath() const;
+        std::string to_string() const;
         bool isGen1Origin() const;
         static VSSPath fromVSSGen2(std::string vss); //Expect Gen2 path
         static VSSPath fromVSSGen1(std::string vssgen1); //Expect Gen1 path
-        static VSSPath fromJSON(std::string json); //Expect json path
+        static VSSPath fromJSON(std::string json, bool gen1); //Expect json path
```
SebastianSchildt, line 36: Why this parameter? JSONPAth do not have "gen1/gen2" difference, they are JSONpath expressions, so I do not understand the meaning here.

wenwenchenbosch: JSON has no difference for gen1 and gen2, therefore, external input needed, therefore the parameter here to construct a new VSSPath

SebastianSchildt: But there should no usecase for this ever? "gen1" and "gen2" is only ever relevant when constructing from an incoming JSON query, when do we ever construct a fresh VSSPath from a JSONPath where that is relevant?
At least there should be a version where it is not needed/set to default

SebastianSchildt (approved):

Issue #211, opened 2021-07-06
Problems setting types for mapping signals to VSS using MQTT


anntexx:
When I configure kuksa.val to publish vehicle signals to a MQTT mosquitto broker, it seems that only some data points are published to the MQTT broker. It seems in the logs, that not every new signal triggers an update in the  SubscriptionHandler  as it happens with the "Vehicle.OBD.Speed" signal. @wenwenchenbosch: Is this related to a type check problem?

extract of the log of kuksa.val:

```
VERBOSE: Set request with id 899e5736-4304-4dee-8c39-2ac264e5bcc8 for path: Vehicle.Drivetrain.Transmission.Gear
VERBOSE: Receive action: set
VERBOSE: Set request with id 35ab08f9-f3d7-4a53-a482-904374ef5ca8 for path: Vehicle.OBD.Speed
VERBOSE: SubscriptionHandler::updateByPath: new value set at path Vehicle/OBD/Speed29.519999999999996
VERBOSE: MQTTPublisher::Publish topic vss/Vehicle/OBD/Speed
VERBOSE: SubscriptionHandler::updateByUUID: new value set at path 91ed0bb43eb054759813cd784b07176429.519999999999996
VERBOSE: Receive action: set
VERBOSE: Set request with id ef137a98-326f-4dfa-a78f-58eaa0875d38 for path: Vehicle.Drivetrain.Transmission.Gear
VERBOSE: Receive action: set
VERBOSE: Set request with id b4d665fa-df4c-472d-9251-7293e7afa59e for path: Vehicle.OBD.Speed
VERBOSE: SubscriptionHandler::updateByPath: new value set at path Vehicle/OBD/Speed29.60000000000001
VERBOSE: MQTTPublisher::Publish topic vss/Vehicle/OBD/Speed
VERBOSE: SubscriptionHandler::updateByUUID: new value set at path 91ed0bb43eb054759813cd784b07176429.60000000000001
VERBOSE: Receive action: set
VERBOSE: Set request with id 701f7eaf-6561-4354-bff1-a1a8c3c3ad4b for path: Vehicle.Drivetrain.Transmission.Gear
VERBOSE: Receive action: set
VERBOSE: Set request with id f882c94b-917f-4a49-83dd-1795a3f74037 for path: Vehicle.OBD.Speed
VERBOSE: SubscriptionHandler::updateByPath: new value set at path Vehicle/OBD/Speed29.680000000000007
[...]

PR #212, opened 2021-07-07
Feature/improve error messages

SebastianSchildt:
This gives better (error) messages in the server log

 - fixed wording in SubscriptionHandler messages
 - emit a log when a write fails due to missing permissions or non-existant path

That should have made the failure in #211 more obvious

### DIFF
--- a/src/VssCommandSet.cpp
+++ b/src/VssCommandSet.cpp
@@ -62,11 +62,15 @@ std::string VssCommandProcessor::processSet2(WsChannel &channel,
   //(set all or none)
   for ( std::tuple<VSSPath,jsoncons::json> setTuple : setPairs) {
     if (! database->pathExists(std::get<0>(setTuple) )) {
+      stringstream msg;
+      msg << "Path " << std::get<0>(setTuple).to_string() << " does not exist";
+      logger->Log(LogLevel::VERBOSE,msg.str());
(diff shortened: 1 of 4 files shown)


src/VssCommandSet.cpp
```diff
@@ -62,11 +62,15 @@ std::string VssCommandProcessor::processSet2(WsChannel &channel,
   //(set all or none)
   for ( std::tuple<VSSPath,jsoncons::json> setTuple : setPairs) {
     if (! database->pathExists(std::get<0>(setTuple) )) {
+      stringstream msg;
+      msg << "Path " << std::get<0>(setTuple).to_string() << " does not exist";
+      logger->Log(LogLevel::VERBOSE,msg.str());
```
wenwenchenbosch, line 67: VERBOSE enough? INFO or ERROR or WARNING better?

SebastianSchildt: Good point. ERROR not. WARNING or INFO I think. Any opinions? (I woudl try to also make this consistent with get and aother erros then)

wenwenchenbosch: I think WARNING is better. thx

SebastianSchildt: Changed. I tried to use "WARNING" for errors such as element not found or no access rights. I keep "ERROR" for syntactically invalid JSON and for  unexpected errors (i.e. in the catch std::exception clauses, where we actually do not expect to land)

wenwenchenbosch (approved):

PR #214, opened 2021-07-08
Transfromations and better error logging for DBC feeder

SebastianSchildt:
- Adds the abilty to transform data (mapping or math formula) to dbc feeder, allowing more flexible mapping. See updated `mapping.yml` for examples
 - Will print error messages from kuksa.val server

Should have made detecting #211 much easier @anntexx 

Sidenote: The feeder/kuksa_viss _client is immensely slow for me.  I think some blocking problem, but that is a topic for another ticket :)

Can't add you as a reviewer @anntexx , but if can test whether DBC feeder works at least as good for you as before, would be nice

### DIFF
--- a/kuksa_feeders/dbc2val/dbc2vssmapper.py
+++ b/kuksa_feeders/dbc2val/dbc2vssmapper.py
@@ -11,12 +11,23 @@
 ########################################################################
 
 import yaml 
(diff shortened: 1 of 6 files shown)


wenwenchenbosch (approved):
tested, and it works for me

Issue #220, opened 2021-07-27
kuksa_viss_client sdk shows bad performance 


wenwenchenbosch:
It seems that, the while loop blocks somehow the async call of websocket:
https://github.com/eclipse/kuksa.val/blob/4c72ab008359e642a8b59c05d86264199961e99f/kuksa_viss_client/__init__.py#L51-L59

for example: dbc feeders shows very bad performance using kuksa_viss_client

PR #221, opened 2021-08-09
220 improve client performance

wenwenchenbosch:
solve #220

### DIFF
--- a/kuksa_feeders/dbc2val/dbcfeeder.py
+++ b/kuksa_feeders/dbc2val/dbcfeeder.py
@@ -44,6 +44,7 @@ kuksa = KuksaClientThread(kuksaconfig)
 kuksa.start()
 kuksa.authorize()
 
+
 mapping = dbc2vssmapper.mapper(kuksaconfig.get('mapping', "mapping.yml"))
 canQueue = queue.Queue()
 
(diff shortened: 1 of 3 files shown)


kuksa_feeders/dbc2val/dbcfeeder.py
```diff
@@ -44,6 +44,7 @@
 kuksa.start()
 kuksa.authorize()
 
+
```
SebastianSchildt, line 47: Why changing this?

kuksa_viss_client/__init__.py
```diff
@@ -189,5 +210,6 @@ async def mainLoop(self):
     # Thread function: Start the asyncio loop
     def run(self):
         self.loop = asyncio.new_event_loop()
+        self.loop.set_debug(True)
```
SebastianSchildt, line 213: Is this needed in production?

SebastianSchildt (commented):
See small comments.

SebastianSchildt (changes requested):
Much faster 👍   🥇 

Should merge, but I still feel we should not run in asyncio debug mode

kuksa_viss_client/__init__.py
wenwenchenbosch:
```suggestion
```

SebastianSchildt (approved):

PR #224, opened 2021-08-24
Replay script for record feature

linussieweke:
Python replay script for the previously implemented record feature. It uses the [kuksa stress Client](https://github.com/eclipse/kuksa.val/tree/master/test/stresstest/stressClient.py) as the base for connection to the server.
I could not find any fitting timing libraries, so the time delta implementation was done manually.

### DIFF
--- a/test/replay/_replay.py
+++ b/test/replay/_replay.py
@@ -1,0 +1,94 @@
+import sys, os
+import time, datetime
+scriptDir= os.path.dirname(os.path.realpath(__file__))
+sys.path.append(os.path.join(scriptDir, ".."))
+import argparse
+from stresstest import stressClient
+import csv
+from pickle import FALSE,TRUE
+import string
+
+def workaround_dict(a,b,timeout=0):
+    pass
+
+def getTimeDiff(timestamps):
+    times = []
+    time = []
+    hours = []
+    minutes = []
+    seconds = []
+    microseconds = []
+    timeDiff = [0]
+
+    for i in range(0,len(timestamps)):
+        times.append(timestamps[i][timestamps[i].find(" ")+1:])     #take only the information about time from the timestamp
+        
+        #cut out corresponding information from the time string
+        hours.append(int(times[i][0:2]))
+        minutes.append(int(times[i][3:5]))
+        seconds.append(int(times[i][6:8]))
+        microseconds.append(int(times[i][9:]))
+
+        #calculate time delta in seconds by adding up all the values with their correlating factor (i.e. 1h = 3600s)
+        time.append((float(microseconds[i]*10**-6)+float(seconds[i])+float(minutes[i]*60)+float(hours[i]*3600)))
+
+    for i in range(1,len(time)):
+        timeDiff.append(time[i]-time[i-1])
+
+    return timeDiff        #return all time differences in one list
+
+
+rowIDs = [
+    "timestamp",
+    "ID",
+    "action",
+    "path",
+    "value"
+]
+timestapIDs =  []
+
+cmdParser = argparse.ArgumentParser(description="Initialise replay function")
+
+cmdParser.add_argument('path', type=str, help="Specify path for replay file")
+cmdParser.add_argument('-g', '--getValue', action='store_true', help="Enable replay of getValue function for debugging purposes (might overstress the server)")
+args=cmdParser.parse_args()
+
+try:
+    VissClient = stressClient.StressClient()
+
+    actionFunctions = {
+        "get": VissClient.commThread.getValue,
+        "set": VissClient.commThread.setValue
+        }
+
+    if not args.getValue:
+        actionFunctions["get"] = workaround_dict   #don't call get functions when getValue is not specified
+
+        try:
+            open(args.path,"r")
+        except:
+            print("Could not open log file at " + args.path)
+
+        with open(args.path,"r") as recordFile:
+            fileData = csv.DictReader(recordFile,rowIDs,delimiter=';')
+
+            timestamps = []
+
+            for row in fileData:
+                timestamps.append(row["timestamp"])
+
+        timeDiff = getTimeDiff(timestamps)              #Array with all time deltas in seconds
+
+    with open(args.path,"r") as recordFile:             #fileData gets reset to empty list, reopen for further action
+        fileData = csv.DictReader(recordFile,rowIDs,delimiter=';')
+
+        for i,row in enumerate(fileData):
+            time.sleep(timeDiff[i])
+            actionFunctions.get(row['action'])(row['path'],row['value'],timeout=0)
+            
+    print("Replay successful")
+
+except:
+    print("Aborted replay script")
+
+os._exit(1)
--- a/test/replay/config.ini
+++ b/test/replay/config.ini
@@ -1,0 +1,10 @@
+[vss]
+
+ip=127.0.0.1
+port=8090
+insecure=false
+token=../../kuksa_certificates/jwt/all-read-write.json.token
+timeout=0.1
+certificate=../../kuksa_certificates/Client.pem
+key=../../kuksa_certificates/Server.key
+cacertificate=../../kuksa_certificates/CA.pem


test/replay/_replay.py
```diff
@@ -0,0 +1,94 @@
...
+        times.append(timestamps[i][timestamps[i].find(" ")+1:])     #take only the information about time from the timestamp
+        
+        #cut out corresponding information from the time string
+        hours.append(int(times[i][0:2]))
+        minutes.append(int(times[i][3:5]))
+        seconds.append(int(times[i][6:8]))
+        microseconds.append(int(times[i][9:]))
+
+        #calculate time delta in seconds by adding up all the values with their correlating factor (i.e. 1h = 3600s)
+        time.append((float(microseconds[i]*10**-6)+float(seconds[i])+float(minutes[i]*60)+float(hours[i]*3600)))
```
wenwenchenbosch, line 33: `timedelta` https://www.geeksforgeeks.org/python-datetime-timedelta-function/ does not work?

test/replay/config.ini
```diff
@@ -0,0 +1,10 @@
+[vss]
+
+ip=127.0.0.1
+port=8090
+insecure=false
+token=../../kuksa_certificates/jwt/all-read-write.json.token
+timeout=0.1
```
wenwenchenbosch, line 7: this argument used?

test/replay/_replay.py
```diff
@@ -0,0 +1,94 @@
+import sys, os
```
wenwenchenbosch, line 1: I would like to move the script into https://github.com/eclipse/kuksa.val/tree/master/kuksa_feeders

wenwenchenbosch (changes requested):
a readme file would be nice

test/replay/config.ini
linussieweke: must be in for the stress Client

test/replay/_replay.py
linussieweke: You would have to make the same calculations as in my solution, so I assumed it to be unnecessarily complicated

kuksa_feeders/replay/stressClient.py
```diff
@@ -0,0 +1,55 @@
+import configparser
+import sys
+import argparse
+import time
+from pickle import FALSE
```
wenwenchenbosch, line 5: why this import?

kuksa_feeders/replay/_replay.py
```diff
@@ -0,0 +1,97 @@
...
+timestapIDs =  []
+
+cmdParser = argparse.ArgumentParser(description="Initialise replay function")
+
+cmdParser.add_argument('path', type=str, help="Specify path for replay file")
+cmdParser.add_argument('-g', '--getValue', action='store_true', help="Enable replay of getValue function for debugging purposes (might overstress the server)")
+args=cmdParser.parse_args()
+
+try:
+    VissClient = stressClient.StressClient()
```
wenwenchenbosch, line 60: I would like to use `KuksaClientThread` directly instead of using `StressClient`. I do not see any benefit of `StressClient`.

kuksa_feeders/replay/_replay.py
```diff
@@ -0,0 +1,97 @@
...
+            for row in fileData:
+                timestamps.append(row["timestamp"])
+
+        timeDiff = getTimeDiff(timestamps)              #Array with all time deltas in seconds
+
+    with open(args.path,"r") as recordFile:             #fileData gets reset to empty list, reopen for further action
+        fileData = csv.DictReader(recordFile,rowIDs,delimiter=';')
+
+        for i,row in enumerate(fileData):
+            time.sleep(timeDiff[i])
```
wenwenchenbosch, line 89: here array out of bound? len(timeDiff) = len(fileData)-1 !

kuksa_feeders/replay/_replay.py
```diff
@@ -0,0 +1,97 @@
...
+    "path",
+    "value"
+]
+timestapIDs =  []
+
+cmdParser = argparse.ArgumentParser(description="Initialise replay function")
+
+cmdParser.add_argument('path', type=str, help="Specify path for replay file")
+cmdParser.add_argument('-g', '--getValue', action='store_true', help="Enable replay of getValue function for debugging purposes (might overstress the server)")
+args=cmdParser.parse_args()
```
wenwenchenbosch, line 57: we decided to use configparser instead of argumentParser, because we have too many argument to be parsed. Therefore, it would be better to also use config parser to parse the argument here

kuksa_feeders/replay/_replay.py
```diff
@@ -0,0 +1,97 @@
...
+    with open(args.path,"r") as recordFile:             #fileData gets reset to empty list, reopen for further action
+        fileData = csv.DictReader(recordFile,rowIDs,delimiter=';')
+
+        for i,row in enumerate(fileData):
+            time.sleep(timeDiff[i])
+            actionFunctions.get(row['action'])(row['path'],row['value'],timeout=0)
+
+    print("Replay successful")
+
+except:
```
wenwenchenbosch, line 94: please also print the exception information

kuksa_feeders/replay/_replay.py
```diff
@@ -0,0 +1,97 @@
...
+
+cmdParser.add_argument('path', type=str, help="Specify path for replay file")
+cmdParser.add_argument('-g', '--getValue', action='store_true', help="Enable replay of getValue function for debugging purposes (might overstress the server)")
+args=cmdParser.parse_args()
+
+try:
+    VissClient = stressClient.StressClient()
+
+    actionFunctions = {
+        "get": VissClient.commThread.getValue,
```
wenwenchenbosch, line 63: good idea here

kuksa_feeders/replay/_replay.py
```diff
@@ -0,0 +1,97 @@
...
+        hours.append(int(times[i][0:2]))
+        minutes.append(int(times[i][3:5]))
+        seconds.append(int(times[i][6:8]))
+        microseconds.append(int(times[i][9:]))
+
+        #calculate time delta in seconds by adding up all the values with their correlating factor (i.e. 1h = 3600s)
+        time.append((float(microseconds[i]*10**-6)+float(seconds[i])+float(minutes[i]*60)+float(hours[i]*3600)))
+
+    for i in range(1,len(time)):
+        timeDiff.append(time[i]-time[i-1])
```
wenwenchenbosch, line 39:
```suggestion
        curr = datetime.datetime.strptime(timestamps[i], '%Y-%b-%d %H:%M:%S.%f')
        pre = datetime.datetime.strptime(timestamps[i-1], '%Y-%b-%d %H:%M:%S.%f')
        delta = (curr-pre).total_seconds()

        timeDiff.append(delta)
```

test/replay/_replay.py
wenwenchenbosch: I gave you a suggestion for the calculation in your new code

kuksa_feeders/replay/_replay.py
```diff
@@ -0,0 +1,97 @@
...
+
+    if not args.getValue:
+        actionFunctions["get"] = workaround_dict   #don't call get functions when getValue is not specified
+
+        try:
+            open(args.path,"r")
+        except:
+            print("Could not open log file at " + args.path)
+
+        with open(args.path,"r") as recordFile:
```
wenwenchenbosch, line 75: Actually, I would like to avoid open the file twice, it is also not necessary to calculate all the delta at the beginning. You can calculate the current delta while reading lines, so that the python interpreter can optimize your code.

kuksa_feeders/replay/_replay.py
linussieweke: `StressClient` is an easily usable version for just connecting to the server and playing get and set Value functions into it. If I used `KuksaClientThread` I would just re-implement the `StressClient`

kuksa_feeders/replay/stressClient.py
linussieweke: stayed in there from previous versions, but now unnecessary. Will be deleted

kuksa_feeders/replay/_replay.py
linussieweke: No, same size. `timeDiff` also has an entry for the 0th index in `fileData`, which is 0 (for no difference to the on before)

kuksa_feeders/replay/_replay.py
wenwenchenbosch: The meaning of `KuksaClientThread` is to provide user a sdk to implement a client in an easy way. If some features was missing, we can discuss about it. Currently, I actually can not see more benefit of `StressClient`.

linussieweke: `StressClient` is an already implemented client, stripped down to the bare minimum of simply connecting to the server and being able to access data, which is exactly what this task needs. Why would you re-implement the same client?

kuksa_feeders/replay/_replay.py
wenwenchenbosch: ok I see it now

Issue #227, opened 2021-08-24
Compiling error


wenwenchenbosch:
After merging #206, the following compile error happens:
```
/home/chw2lr/code/kuksa.val/src/main.cpp: In function ‘int main(int, const char**)’:
/home/chw2lr/code/kuksa.val/src/main.cpp:269:7: error: reference to ‘thread’ is ambiguous
  269 |       thread http(httpRunServer, variables, httpServer, docRoot, cmdProcessor);
```

SebastianSchildt:
Fixed

PR #228, opened 2021-08-24
fix compile error and warning

wenwenchenbosch:
solve #227


SebastianSchildt (changes requested):
Compiled it but segfaults immediately

SebastianSchildt (approved):
Ok, was probably some workspace problem. Starting with a fresh and empty build folder worked for Debug and Release builds

PR #231, opened 2021-08-30
Unittest business logic of gRPC Client + test Services via grpc_cli

lukasmittag:
Unittests for the gRPC Services and testing the logic of the gRPC Client


test/unit-test/kuksa_grpc_client_mock.cpp
```diff
@@ -0,0 +1 @@
+
```
wenwenchenbosch, line 1: empty file?

src/CMakeLists.txt
```diff
@@ -48,6 +48,8 @@ set(_GRPC_CPP_PLUGIN_EXECUTABLE $<TARGET_FILE:grpc_cpp_plugin>)
 set(_REFLECTION grpc++_reflection)
 set(_PROTOBUF_PROTOC $<TARGET_FILE:protoc>)
 
+include(GoogleTest)
```
wenwenchenbosch, line 51: googletest needed?

wenwenchenbosch (commented):
I just see, the PR is a draft pr.

src/CMakeLists.txt
lukasmittag: no, I will fix. I tried something.

wenwenchenbosch (approved):

PR #233, opened 2021-08-30
144 githubaction

wenwenchenbosch:


.github/workflows/kuksa_val_docker_amd64.yml
```diff
@@ -0,0 +1,58 @@
...
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+    
+jobs:
+  build:
+    runs-on: [ self-hosted ]
```
SebastianSchildt, line 11: Should mention the correct task, to prevent amd64 builds on wrong architecture (ALL private runners wil listen to self.hosted tag).

kuksa-x64 shoudl be used here

.github/workflows/kuksa_val_docker_arm64.yml
```diff
@@ -0,0 +1,53 @@
...
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+    
+jobs:
+  build:
+    runs-on: [ self-hosted, kuksa-x64 ]
```
SebastianSchildt, line 11: kuksa-x64 is wrong, needs to be kuksa-arm64  , as we do not have that yet, self-hosted also needs to remain for the time being

.github/workflows/kuksa_viss_client.yml
```diff
@@ -0,0 +1,46 @@
...
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+    
+jobs:
+  build:
+    runs-on: ubuntu-latest
```
SebastianSchildt, line 11: Fine for now, in the future we should discuss whether we als should run this on our own runners

.github/workflows/kuksa_feeders_docker.yml
```diff
@@ -0,0 +1,45 @@
...
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+    
+jobs:
+  build:
+    runs-on: ubuntu-latest
```
SebastianSchildt, line 11: Fine for now, in the future we should discuss whether we als should run this on our own runners

SebastianSchildt (changes requested):
Runs-on needs some work, otherwise looking good

SebastianSchildt:
No artefact for VISS client: https://github.com/eclipse/kuksa.val/actions/runs/1717727366

SebastianSchildt:
Not artifact in feeders https://github.com/eclipse/kuksa.val/actions/runs/1717727365

SebastianSchildt (approved):

PR #236, opened 2021-09-17
Feature/load overlays

### LINKED ISSUE #177: Loading multiple models
Currently kuksa.val expects to load on VSS model (`-vss parameter`)

This tree can be extended/updated during runtime using updateVSSTree https://github.com/eclipse/kuksa.val/blob/master/doc/liveUpdateVSSTree.md

However, this needs to be done by an application (which also needs elevated rights in kuksa.val to do this)

It might be useful for a specific deployment, that kuksa.val loads and overlays several VSS files during startup, e.g.

 - a core VSS (standard) model
 - some vehicle-deployment specific additions
 - other models (e.g. trailer, system information, etc...)

The goal of this feature would be to have an optional parameter giving a directory, where kuksa.val upon start and parsing of the main VSS file would load all JSON files sorted alphanumerically and apply them to the VSSDataBase using the same semantic as the liveUpdateVSSTree.

SebastianSchildt:
Add support to overlay several VSS models. New parameter  `--overlays` expecting a directory. All *.json files will be applied in alphanumeric order.

The effect is the same, as authenticated and adding the files via updateVSSTree, but it happens automatically at server start (terminating in case of any problems), so it is possible to refrain from doing this potentially dangerous and complex operations during runtime.

This will close #177 

Also this changed the log level for applying a default from INFO to VERBOSE

### DIFF
--- a/src/OverlayLoader.cpp
+++ b/src/OverlayLoader.cpp
@@ -1,0 +1,68 @@
+/*
+ * ******************************************************************************
+ * Copyright (c) 2021 Robert Bosch GmbH.
+ *
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v2.0
+ * which accompanies this distribution, and is available at
+ * https://www.eclipse.org/org/documents/epl-2.0/index.php
+ *
+ * *****************************************************************************
+ */
+
+/* helper fuctions to load overlays during server startup */
+
+#include <stdexcept>
+#include <jsoncons/json.hpp>
+
+#include "OverlayLoader.hpp"
+#include "kuksa.pb.h"
+
+using jsoncons::json;
+
+std::vector<boost::filesystem::path> gatherOverlays(
+    std::shared_ptr<ILogger> logger, boost::filesystem::path overlaydir) {
+  std::vector<boost::filesystem::path> overlayfiles;
+  logger->Log(LogLevel::VERBOSE,
+              "Searching overlays in \"" + overlaydir.generic_string() + "\"");
+  if (!boost::filesystem::is_directory(overlaydir)) {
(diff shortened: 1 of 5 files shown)


src/OverlayLoader.cpp
```diff
@@ -0,0 +1,68 @@
...
+  for (auto &p : boost::filesystem::directory_iterator(overlaydir)) {
+    if (boost::filesystem::is_regular_file(p.path()) &&
+        p.path().extension() == ".json") {
+      logger->Log(LogLevel::VERBOSE, "Found " + p.path().generic_string());
+      overlayfiles.push_back(p);
+    }
+  }
+
+  // Sort according to filename
+  std::sort(overlayfiles.begin(), overlayfiles.end(),
```
wenwenchenbosch, line 41: why sort?

SebastianSchildt: Because the resulting VSS depends on the order the VSS files are merged. So the documentation states will be merged in "alphanumeric order". This sorts the detected filenames (it is not defined in what order we get them, when just iterating directory entries)

wenwenchenbosch:
tested and it works. only a doubts: Is it meaningful to consider all *.json file of the given folder?

For testing, I have a json file created in the given folder:
```
> cat test.json 
{
    "bla": "blu"
}
```
Then
```
Test Client> getMetaData bla
{
    "action": "getMetaData", 
    "metadata": {
        "bla": "blu"
    }, 
    "requestId": "6f19c09f-dafe-4f36-a4a8-29cf5ac5caca", 
    "ts": "2021-09-24T12:24:11.1632482651Z"
}

```
it is also perfectly merged. but it may be somehow unexpected.

Maybe better to support list of file with wildcard support like `--overlay myfolder/vss_*.json`

wenwenchenbosch (approved):

PR #241, opened 2021-11-18
Accept default only for attributes according to VSS spec

### LINKED ISSUE #201: Default values only valid for attributes
We added support for default values in https://github.com/eclipse/kuksa.val/pull/190  

There has been discussion for which elements `default` is applicable. We opted to support it for sensors, attributes and actors, as those where seen in current VSS specification.

Subsequently VSS decided `default` is only valid for attribute and declared other usage bugs. See:

https://github.com/GENIVI/vehicle_signal_specification/issues/283
https://github.com/GENIVI/vehicle_signal_specification/pull/284
https://github.com/GENIVI/vss-tools/pull/63

We should change kuksa.val code (and tests) again, so `default` is only checked and processed for attributes

SebastianSchildt:
Fixes #201

### DIFF
--- a/src/VssDatabase.cpp
+++ b/src/VssDatabase.cpp
@@ -58,14 +58,13 @@ bool VssDatabase::isAttribute(const json &element) {
 }
 
 
-/** Iterates over a given VSS tree and checks for sensors and actuators specifying the "default"
- *  metadata. If a default is present, it will be used as "value" (and thus returned upon get
- *  even if there has been no previous set)
+/** Iterates over a given VSS tree and checks attributes specifying the "default"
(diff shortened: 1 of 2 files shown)


wenwenchenbosch:
do we also need error/warning info, if default value is applied for none attributes?

SebastianSchildt:
I don't think so,becasue basically that would mean you have an invalid VSS file, VSS file validity SHOULD be checked by css-tools.

During "runtime" with the client there is no way to "set defaults" (other than loading another VSS file)

wenwenchenbosch (approved):

wenwenchenbosch:
I am thinking about `updateMetaData` function, but I think it is seperate story, if we should check it during change metadata

Issue #237, opened 2021-09-17
Remove JSON signing 


SebastianSchildt:
I wonder, whether we should remove the code for JSON signing.

It has not seen much (any?) use, and TLS is the cleaner solution anyway.

PR #242, opened 2021-11-18
Remove JSON signing feature

SebastianSchildt:
Remove JSON signing feature as mentioned in #237


test/unit-test/KuksavalUnitTest.cpp
```diff
@@ -233,21 +230,21 @@ BOOST_AUTO_TEST_CASE(Test_GetMetaDataService){
             \"Vehicle\": {\
                 \"children\": {\
                     \"Speed\": {\
-                        \"datatype\": \"int32\",\ 
-                        \"description\": \"Vehicle speed, as sensed by the gearbox.\",\ 
-                        \"max\": 250,\ 
-                        \"min\": -250,\ 
-                        \"type\": \"sensor\",\ 
-                        \"unit\": \"km/h\",\ 
+                        \"datatype\": \"int32\",\
```
wenwenchenbosch, line 233:
```suggestion
                        \"datatype\": \"int32\",
```

But why rm all ending `\`

SebastianSchildt: There were many warnings from the compiler, because there was a space after \

wenwenchenbosch (approved):

PR #247, opened 2022-02-08
fix security issues found by dependabot

wenwenchenbosch:
upgrade node-red version and websocket version

@SebastianSchildt do we need clear the new websocket version?


kuksa_apps/s3/requirements.txt
```diff
@@ -1,4 +1,3 @@
-importlib_metadata >= 3.6
-websockets~=8.1
+kuksa-viss-client
```
SebastianSchildt, line 1: No version. Should at least have >= or ~=

PR #257, opened 2022-03-25
Codecov

wenwenchenbosch:
- Add coverage information using codecov, you can find report here: https://app.codecov.io/gh/eclipse/kuksa.val/commits?page=1 still no overview page, I think, it will come, if the change is merged to master
- revert boost unittest result format to XML again, because JUnit format causes randomly issues


src/CMakeLists.txt
```diff
@@ -35,6 +38,15 @@ protobuf_gen(PROTO ${kuksa_proto}
     OUTPUT ${proto_gen_dir}
 )
 include_directories("${proto_gen_dir}")
+add_library(${GRPC_GENERATED_LIB_NAME} OBJECT ${PROTO_SRCS}
```
wenwenchenbosch, line 41: add generated grpc code as separate lib, since there is a cmake problem for collecting coverage information. The codecov cmake always use a wrong path for the gerade path under build directory. I think it is a bug of the codecov cmake module, but I can not figure out, how to fix it. 
But I think, it is not necessary to collect coverage information for generated code. Therefore, I define it as a separate lib to workaround the problem.

wenwenchenbosch:
unfortunately, if people create pr from bosch research orga, because bosch research orga does not give permission to codecov, codecov can not comment the coverage changes in PR. Otherwise, it looks like this: https://github.com/wenwenchenbosch/kuksa.val/pull/4

Issue #261, opened 2022-04-08
Inconsistency in data type of returned JSON objects


nayakned:
Currently all fields of the request/response JSON object is string.
The only exception is error.number in an event of an error.

Either we need to correct this inconsistency or document the exceptions.

```
{
    "action": "get", 
    "data": {
        "dp": {
            "ts": "2022-04-08T11:26:23.1649413583Z", 
            "value": "68"
        }, 
        "path": "Vehicle.Speed"
    }, 
    "requestId": "05b9eb16-5216-4371-9c47-bff2c8aa0421", 
    "ts": "2022-04-08T11:26:23.1649413583Z"
}

```

```
{
    "action": "get", 
    "error": {
        "message": "I can not find Vehicle.Speed1 in my db", 
        "number": 404, 
        "reason": "Path not found"
    }, 
    "requestId": "51a2b9fa-f576-45b3-8767-6c5504b6518b", 
    "ts": "2022-04-08T11:27:19.1649413639Z"
}

```

nayakned:
Also found in the subscriptionId which is returned as an int instead of a string.
```
{
    "action": "subscribe", 
    "requestId": "6bbccf7c-e8ca-4cbe-b869-6e8535491ea3", 
    "subscriptionId": 1350997395, 
    "ts": "2022-04-10T12:07:01.1649588821Z"
}
```

This has the consequence that if the subscriptionId is used for unsubscribing as a string, then there may be an unintended side effects in kuksa.val server.

```
VERBOSE: Receive action: unsubscribe
terminate called after throwing an instance of 'jsoncons::json_runtime_error<std::runtime_error, void>'
  what():  Integer overflow
```

SebastianSchildt:
Fixed

Issue #263, opened 2022-04-10
Subscription handling in Kuksa.val


nayakned:
It looks like the kuksa.val server is not deleting subscription sessions if the client dies before unsubscribing or forgets to unsubscribe.

Here a log. I ran a client session multiple times while subscribing for same data Vehicle.Speed. I didnt unsubscribe while the client terminated. 
```
VERBOSE: SubscriptionHandler::update: set value "68" for path Vehicle.Speed for UUID efe50798638d55fab18ab7d43cc490e9
VERBOSE: SubscriptionHandler::updateByUUID: new value set at path 675526736SubscriptionHandler::update: set value "68" for path Vehicle.Speed for UUID efe50798638d55fab18ab7d43cc490e9
VERBOSE: SubscriptionHandler::updateByUUID: new value set at path 675407987SubscriptionHandler::update: set value "68" for path Vehicle.Speed for UUID efe50798638d55fab18ab7d43cc490e9
VERBOSE: SubscriptionHandler::updateByUUID: new value set at path 674175860SubscriptionHandler::update: set value "68" for path Vehicle.Speed for UUID efe50798638d55fab18ab7d43cc490e9
VERBOSE: SubscriptionHandler::updateByUUID: new value set at path 677776468SubscriptionHandler::update: set value "68" for path Vehicle.Speed for UUID efe50798638d55fab18ab7d43cc490e9
VERBOSE: SubscriptionHandler::updateByUUID: new value set at path 676432633SubscriptionHandler::update: set value "68" for path Vehicle.Speed for UUID efe50798638d55fab18ab7d43cc490e9
VERBOSE: SubscriptionHandler::updateByUUID: new value set at path 676451158SubscriptionHandler::update: set value "68" for path Vehicle.Speed for UUID efe50798638d55fab18ab7d43cc490e9
VERBOSE: SubscriptionHandler::updateByUUID: new value set at path 674637585SubscriptionHandler::update: set value "68" for path Vehicle.Speed for UUID efe50798638d55fab18ab7d43cc490e9

```
The network sessions are being closed but the subscription sessions arent.
```
tcp        0      0 127.0.0.1:8090          0.0.0.0:*               LISTEN     
tcp        0      0 127.0.0.1:38784         127.0.0.1:8090          FIN_WAIT2  
tcp      143      0 127.0.0.1:8090          127.0.0.1:38784         CLOSE_WAIT 
[...]

PR #264, opened 2022-04-11
Subscriptions Fixes 1

SebastianSchildt:
Fixes nothing in #263 , but makes subscriptions a little more robust

 - enforcing string type subscriptionIDs both directions
   - fixes the occasional "Integer Overflow" on unsubscribe
 - JSON schema checking for subscribe and unsubscribe requests
 - as a result, overall more helpful error messages

@nayakned  please check if it, at the very least, does not make things worse :)

### DIFF
--- a/include/JsonResponses.hpp
+++ b/include/JsonResponses.hpp
@@ -28,10 +28,10 @@ namespace JsonResponses {
                         std::string message,
                         jsoncons::json& jsonResponse);
 
-  std::string malFormedRequest(std::string message);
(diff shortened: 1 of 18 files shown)


nayakned:
Looks good. These changes fix a part of #261.
Maybe also fix the error code returned in JSON object or document that it will be an int.
Then we can also close #261.

SebastianSchildt:
Now all error codes are returned as string. This will fix #261 , please check @nayakned

nayakned:
Checked a couple of errors. Those were strings as expected.

Issue #47, opened 2020-09-03
Error after testclient quit


wenwenchenbosch:
After using testclient connect the kuksa client like this:
```
> python -m pipenv run python testclient.py
VSS Client> connect 
Websocket connected!!
VSS Client> quit
```

Kuksa client report the following errors:
```
> ./kuksa-val-server
kuksa.val server
Commit 517fad4-dirty from 2020-09-03T09:09:49+02:00
Update vss path to /home/wenwen/tools/kuksa.val/build/src/vss_rel_2.0.json
Update cert-path to /home/wenwen/tools/kuksa.val/build/src/.
Log START
VERBOSE: Try reading JWT pub key from /home/wenwen/tools/kuksa.val/build/src/./jwt.key.pub
VERBOSE: SubscribeThread: Started Subscription Thread!
VERBOSE: VssDatabase::VssDatabase : VSS tree initialized using JSON file = /home/wenwen/tools/kuksa.val/build/src/vss_rel_2.0.json
INFO: Initializing Boost.Beast web-socket and http server on port 8090
INFO: Starting Boost.Beast web-socket and http server
VERBOSE: Receive action: getMetadata
INFO: VssCommandProcessor::processQuery: Unknown action getMetadata
ERROR: read: stream truncated
ERROR: JSON parse error
```

wenwenchenbosch:
I created an issue in websockets repo to ask why https://github.com/aaugustin/websockets/issues/893

wenwenchenbosch:
@nayakned I tested with node red, no similar issues there. `disconnect` function works well with node red websocket.

nayakned:
So as it turns out, the Kuksa Python Client awaits on the incorrect close() for websocket.
https://github.com/boschresearch/kuksa.val/blob/master/kuksa_viss_client/__init__.py#L219

websocket.close() only initiates the closing of the websocket and not its completion. 
Awaiting on wait_closed() solves the issue. PR coming up shortly.

nayakned:
Similar case also on the golang client. Current close() function just closes the underlying TCP connection.
https://pkg.go.dev/github.com/gorilla/websocket#Conn.Close

nayakned:
Problem still exists for SSL sockets. Maybe related to this:
https://stackoverflow.com/questions/32046034/what-is-the-proper-way-to-securely-disconnect-an-asio-ssl-socket/32054476#32054476

SebastianSchildt:
[...]

PR #266, opened 2022-04-14
Fix/subscriptions2

SebastianSchildt:
Refactor subscription handler and improving websocket connection handling to improve stability

@nayakned Still need to fix at leas unit tests, but you might take a kook and check if this fixes anything from #263 

Please note, that if a websocket connection is shut down uncleanly it will likely only be detected, when the next value is pushed to it.


nayakned:
Works like a charm. Couldn't reproduce #263 with either secure or insecure mode. You can merge #266 and close #263.
#47 remains nonetheless and needs to be looked into.

PR #269, opened 2022-04-23
Subscribe/Unsubscribe for Python Client

nayakned:
- Implemented the unsubscribe call in __init__.py
- Implemented subscribe and unsubscribe in the interactive client. To test execute the client and after authorization execute "subscribe Vehicle.Speed". Subsequently execute the command (e.g. tail -f <path>/log_Vehicle.OBD.EngineLoad_1650736673.377693) printed on the client in another terminal instance.


kuksa_viss_client/__main__.py
```diff
@@ -164,6 +183,36 @@ def do_getValue(self, args):
...
+        """Subscribe the value of a path"""
+        if self.checkConnection():
+            resp = self.commThread.subscribe(args.Path, lambda msg: self.subscribeCallback(args.Path, msg))
+            resJson =  json.loads(resp) 
+            if "subscriptionId" in resJson:
+                fileName = os.getcwd() + "/log_"+args.Path+"_"+str(time.time())
+                self.subscribeFileDesc[args.Path] = open(fileName, "w")
+                self.subscribeIdToPath[resJson["subscriptionId"]] = args.Path
+                print("Subscription log available at " + fileName)
+                print("Execute tail -f " + fileName + " on another Terminal instance")
```
SebastianSchildt, line 198: Is it possible to ASK first, whether to open an xterm? Because often there might not be an X system on the system, and this would fail

nayakned: Ofcourse we can add a flag for the user to use (or not use) xterm.

Currently it would fail rather silently. 
image
If you like, we can even redirect the stderr of the subprocess to avoid this.
This way user would not see this at all. The client would use xterm if it is available. Else the user needs to execute the corresponding tail command.

What would be a better solution?

nayakned:
The commit 44ebac0e88653296934108d74a5437d23b9e5619 fixes the issues on subscribing on machines without _xterm_ and also the exceptions thrown when incorrect subscription ids are passed.

Issue #246, opened 2022-01-23
Container start SSL error


SebastianSchildt:
wen starting the kuksa.val continer with an empty config file, there is an SSL realted error during startup

```
sudo docker run -it --rm -v $HOME/kuksamaster.config:/config  -p 127.0.0.1:8090:8090 -e LOG_LEVEL=ALL amd64/kuksa-val:91cb503
Starting kuksa.val
No config file, initialize with example config
No VSS tree, initialize with example
No server keys configured, initialize with example
No jwt key configured, initialize with example
kuksa.val server
Commit 91cb503 from 2022-01-21T17:58:43+01:00
Read configs from /config/config.ini
Update vss path to /config/vss.json
Update cert-path to /config/certs
Log START
[...]
FO: Initializing Boost.Beast web-socket and http server on 0.0.0.0:8090
INFO: Starting Boost.Beast web-socket and http server
Failed to read SSL certificate/config/certs/CA.pem
INFO: Kuksa viss gRPC server Version 1.0.0
INFO: gRPC Server listening on 127.0.0.1:50051

```

SebastianSchildt:
Fixed.

PR #270, opened 2022-04-28
Copy CA.pem to config if not available in the config at start-up.

nayakned:


nayakned:
This should fix #246

PR #272, opened 2022-05-11
Feature/target value

nayakned:
Added the feature of target value for the leaves of the VSS tree which are actors.
The Python VISS Client is also updated to use this feature.

How to test this PR:
- Compile kuksa.val as usual
- Start the updated Python Client
- After authorization, you can execute SetTargetValue and GetTargetValue on the leaves which are actors. These calls are separate from GetValue and SetValue.

**To clarify**: The current implementation adds an attribute "targetValue" to the database, but nonetheless returns the target value in the "value" field of the JSON object. The action field can be used to figure out if its indeed a value or the target value. Alternative, we extend the JSON object to return a key called targetValue.


nayakned:
We will add this as a generic feature in another PR.

SebastianSchildt:
Some tests
 - PR server, master client: Works
 - PR server, PR client: Works
 - Master server, PR client
      - works, of course setTargetValue errors out (as expected), unfortunately getTaregtValue works, but just gives the value (becuase the old JSON schema just ignores the extra field I guess). I think we can live with that.


Observation: When setting targetValue on something not an actor, the error message is misleading:

```json
Test Client> setTargetValue Vehicle/Speed 200
{
    "action": "set", 
    "error": {
        "message": "Can not set path:Vehicle/Speed with attribute:targetValue.", 
        "number": "403", 
        "reason": "Forbidden"
    }, 
    "requestId": "175fb023-d429-40db-9bbe-a406f80bebdd", 
    "ts": "2022-05-18T11:58:14.1652875094Z"
}
```

It hsould either say that it is only allowed for actuators, OR it should set it anyway (considering later we may have arbitrary attributes, some of them might apply to sensors as well). For now maybe the first option is better?

PR #274, opened 2022-05-17
Get/Set Attribute

nayakned:
- Implemented generic handling to get/set attributes
- Currently limited to "targetValue" in the JSON Schema
- Adapted the Python Kuksa VISS Client to use the same.


SebastianSchildt:
Some tests
 - PR server, master client: Works
 - PR server, PR client: Works
 - Master server, PR client
      - works, of course setTargetValue errors out (as expected), unfortunately getTaregtValue works, but just gives the value (becuase the old JSON schema just ignores the extra field I guess). I think we can live with that.


Observation: When setting targetValue on something not an actor, the error message is misleading:

```json
Test Client> setTargetValue Vehicle/Speed 200
{
    "action": "set", 
    "error": {
        "message": "Can not set path:Vehicle/Speed with attribute:targetValue.", 
        "number": "403", 
        "reason": "Forbidden"
    }, 
    "requestId": "175fb023-d429-40db-9bbe-a406f80bebdd", 
    "ts": "2022-05-18T11:58:14.1652875094Z"
}
```

It hsould either say that it is only allowed for actuators, OR it should set it anyway (considering later we may have arbitrary attributes, some of them might apply to sensors as well). For now maybe the first option is better?

SebastianSchildt (approved):
Merging this as basis for the GRPC refactoring.

We will still need some Readmes/exmaples once that is done

PR #290, opened 2022-06-21
Added the feature to get/set attributes to the go client.

nayakned:
You can now get, set and subscribe attributes using the go client.
The APIs have been extended in a backwards compatible manner accordingly.
The main.go now calls the new APIs (for get and set attributes) as an example.


SebastianSchildt (approved):
lgtm

PR #291, opened 2022-06-26
Viss client grpc

nayakned:
This PR can be tested once the #287 is merged.

How to use the kuksa VISS Client for gRPC.

- Ensure that the kuksa.val server implements the get/set/authorize gRPC calls (done in PR #287 ).
- Start the client. It will automatically connect using ws protocol.
- Disconnect it using _disconnect_ command
- Change Server Address using _setServerAddress_ command and use -p switch to change the protocol
-- If insecure mode, IP address would suffice, e.g., setServerAddress -p grpc 127.0.0.1 50051
-- If secure mode, server name (be default: Server ) is needed from the used certificate, e.g., setServerAddress -p grpc Server 50051
- Connect using _connect_ command
-- insecure mode: connect -i
-- secure mode: connect
- Finally execute commands as usual.


kuksa_viss_client/__main__.py
```diff
@@ -11,16 +11,21 @@
...
 from re import sub
 from typing import Dict, List
-import queue, time, os
+import queue, time, os, threading
 from pygments import highlight, lexers, formatters
 from cmd2 import Cmd, with_argparser, with_category, Cmd2ArgumentParser, CompletionItem
 from cmd2.utils import CompletionError, basic_complete
 import functools, subprocess
-DEFAULT_SERVER_ADDR = "127.0.0.1"
+DEFAULT_SERVER_ADDR = "172.17.0.1"
```
SebastianSchildt, line 19: Change back to 127.0.0.1

SebastianSchildt (changes requested):

kuksa_viss_client/__main__.py
nayakned: Chaned with [31dc5a4](https://github.com/eclipse/kuksa.val/pull/291/commits/31dc5a4d0b7a0c1c4fe89a87f0b3426e7f65d9f6)

PR #299, opened 2022-07-07
Change Subscription Handler to track Channel objects instead of conne…

SebastianSchildt:
…ctionID


SebastianSchildt:
Unittest workable again, so basically contains the whole grpc subscription code now, just missing clients @nayakned (still can't you add as reviewer here. We should check what is the issue here)

nayakned:
Tested parallel subscriptions with websockets and gRPC. Looks good. 
Can be merged.

nayakned (approved):
Tested and found good enough to be merged.

Issue #297, opened 2022-07-05
Return error when getting unset values


nayakned:
The attempt to execute getValue on VSS datapoints whose values have not been explicitly set returns "---" string as value(see attached figure).
This being an ambiguous returned value must be replaced with an error.

![Screenshot from 2022-07-05 15-44-48](https://user-images.githubusercontent.com/69036497/177342408-f8b085d3-7fc8-4a67-9538-c9364a7ad15a.png)

nayakned:
Fixed with #303

PR #303, opened 2022-07-22
Throw an error while reading unset values.

nayakned:
getValue on values not explicitly set should throw an error.

This should implement #297


SebastianSchildt:
Tested with websocket and GRPC. Seems to work fine.
I wonder whether we should send another code than 400, but did not find a fitting one on a first glance (definitely not 404, because we use that to indicate the path is no valid). 400 is fine, but generic, and probably used more often.  I am wondering if it is easier if client can easily detect this special condition (nothing published yet) by just looking at the number

SebastianSchildt:
Hmm wrt to 404 it seems VISS suggests it https://www.w3.org/TR/viss2-transport/#status-codes

nayakned:
@SebastianSchildt This exception will also be thrown when a entire branch is being read, and even if one of the leaves has not been previously set. E.g. getValue on Vehicle.OBD.* will throw this error as soon as it encounters the first datapoint that hasnt been set.

Would that be ok?

SebastianSchildt:
> @SebastianSchildt This exception will also be thrown when a entire branch is being read, and even if one of the leaves has not been previously set. E.g. getValue on Vehicle.OBD.* will throw this error as soon as it encounters the first datapoint that hasnt been set.
> 
> Would that be ok?

Guess that is ok. We never promised anything regarding that behaviour (it seems to be one of this "atomic or not" discussions. Most common usecase is querying single values, and if you DO query a branch, most likely it is required/expected to get all subelements to be useful at all (e.g. GPS position), so I think it is toally acceptable to bail out if not all of the requested data can be provided.

PR #304, opened 2022-07-22
Feature/update fix  Testclient and Python Package to spport GRPC

SebastianSchildt:
With the GRPC changes in #299, the client could only be executed  locally. Python package and Docker where broken. This should fix it. Expectation ist

 - testclient still works in (a pipenv) in a local source checkout
 - Python package can be build
    - build with `python setup.py bdist_wheel`
 - Docker container for testclient works
    - Build with `docker build -t testclient -f kuksa_viss_client/Dockerfile .` or wait for GH action and pull from ttl.sh before the image expires

Also, the python packaged related config files (setuptools stuff) have been moved into the kuksa_viss_client folder


nayakned (approved):
Tested all the three variants and found it to be working.
- Local
- Docker 
- Built package inside a docker container and installed it.

Can be merged.

PR #312, opened 2022-08-03
Feature/remove testclient

SebastianSchildt:
This removes the old CPP testclient, as it was essentially abandonware since we have the kuksa_viss_client.

While it still compiled, it was not updated to use the latest JSON schema, so it did not work as is. As it was basically a (bad) example how to code websocket in CPP, I opted to removed it completely, because for testing the kuksa-val-server the python testclient is much more useful


nayakned (approved):
Didnt break anything as far as I could see. :)
Other the old client of course.

PR #316, opened 2022-08-03
Feature/remove http

SebastianSchildt:
This removes the experimental REST interface, that has never seen much use, and was not very stable.

Websocket should (hopefully) still work

can you confirm with testclient/go sdk @nayakned  (also reinterpret cast still in 😁 , the code seems a bit fragile to me, and I did not want to touch too much at once)


doc/usage.md
```diff
@@ -16,16 +16,16 @@ OPTIONS:
...
-                                        vss = vss_release_2.2.json
+                                        vss = vss_release_2.1.json
                                         cert-path = . 
                                         log-level = ALL
                                         
   --vss arg                             [mandatory] Path to VSS data file 
                                         describing VSS data tree structure 
                                         which `kuksa-val-server` shall handle. 
-                                        Sample 'vss_release_2.2json' file can 
+                                        Sample 'vss_release_2.1.json' file can
```
nayakned, line 26: While I understand that this is just an example, any particular reason of changing the v2.2 to v2.1?

nayakned:
Units tests don't compile. _RestV1ApiHandlerTests.cpp_ needs to be removed I guess.

doc/usage.md
SebastianSchildt: Ah! The root cause was, I updated usage.md by copying the output form `kuksa-val-server --help`, it seems that was never updated, probably the doc was "manually" edited when upgrading to 2.2 and we forgot the actual compiled in usage notes. Should be fixed now

nayakned (approved):
Looks good. Can be merged.

Issue #308, opened 2022-07-27
Align databroker and kuksa-val GRPC


SebastianSchildt:
Use the same GRPCs interface in val-server and databroker. Enable GO&Python SDKs for both

SebastianSchildt:
Just a collector, and roadmap changed slightly. So closing this for now. Tasks remain

PR #320, opened 2022-08-05
First scratch of unified base proto

SebastianSchildt:
Hi guys,

too many meetings... I did a first horrible copy&paste orgy with some of the discussed things.

At this point it likely doesn't even run trough protoc compiler, however I think such a "fake PR" makes discussion easier.

Please feel free at this point to just push changes to this branch (of something comes out of it in the end, we can squash).

Idea is to have basic get/set/subscribe

Some comments in file. Still open points

 - Do we want the "Update" streaming also in base? It does seem useful for some feeders (even though can feeder is not using it atm)

 - We still need to think how/if at all we want to unify some of the metadata things. kuksa.databroker and kuksa.val-server differ a little bit here: val-server can do "everything"; but basically because it does not care at all, it just gives you verbatim JSON snippets of the underlying VSS model. databroker gives out structured data, but that of course then is limited to the field it supports directly. One important thing, that probably apps/feeders might want to know is at least querying the datatype of an entry. Also the "registering" of datapoints is almost but not quite entirely unlike the updateMetadata/tree in val-server and again, teaching the server new entries during runtime is a valid thing

works towards #308

### DIFF
--- a/proto/kuksa/val/v1/types.proto
+++ b/proto/kuksa/val/v1/types.proto
@@ -1,0 +1,285 @@
+// Copyright Robert Bosch GmbH, 2022. Part of the Eclipse Kuksa Project.
+//
+// All rights reserved. This configuration file is provided to you under the
+// terms and conditions of the Eclipse Distribution License v1.0 which
+// accompanies this distribution, and is available at
+// http://www.eclipse.org/org/documents/edl-v10.php
+
+syntax = "proto3";
+
+// I added V1 as in databroker. Is this good practice?
+package kuksa.val.v1;
+import "google/protobuf/timestamp.proto";
+
+
+// A Value. 
+message Value {
+  google.protobuf.Timestamp timestamp = 1;
+
+  oneof value {
+    string string            = 11;
+    bool bool                = 12;
+    sint32 int32             = 13;
+    sint64 int64             = 14;
+    uint32 uint32            = 15;
+    uint64 uint64            = 16;
+    float float              = 17;
+    double double            = 18;
+    StringArray string_array = 21;
+    BoolArray bool_array     = 22;
+    Int32Array int32_array   = 23;
+    Int64Array int64_array   = 24;
+    Uint32Array uint32_array = 25;
+    Uint64Array uint64_array = 26;
+    FloatArray float_array   = 27;
+    DoubleArray double_array = 28;
+  }
+}
+
+// Describes a VSS datapoint, includes everything there is
+// to know about it. Most fields optional
+message Datapoint {
+  // Defines the full VSS path of the datapoint.
+  string path = 1;
+
+  // The value of the datapoint
+  Value value = 2;
+
+  // The VSS data type of the entry.
+  //
+  // NOTE: protobuf doesn't have int8, int16, uint8 or uint16 so the actual
+  // value will be serialized as int32 and uint32 respectively
+  // (in the `value` field).
+  //
+  oneof optional_data_type {
+    DataType data_type = 3;
+  }
+
+  // Describes the meaning and content of the datapoint.
+  oneof optional_description {
+    string description = 4;
+  }
+
+  // [optional]
+  // A comment can be used to provide additional informal information
+  // on a datapoint.
+  oneof optional_comment {
+    string comment = 10;
+  }
+
+  // [optional]
+  // Whether this datapoint is deprecated.
+  oneof optional_deprecation {
+    string deprecation = 11;
+  }
+
+  // [optional]
+  // The unit of measurement
+  oneof optional_unit {
+    string unit = 12;
+  }
+
+  // [optional]
+  // Restrict which values are allowed.
+  // Only restrictions matching the DataType {datatype} above is valid/used.
+
+  ValueRestriction value_restriction = 13;
+
+  // Data entry type
+  // Adds specific fields
+  oneof entry_type {
+    Actuator actuator   = 30;
+    Sensor sensor       = 31;
+    Attribute attribute = 32;
+  }
+
+}
+
+///////////////////////
+// Actuator specific fields
+message Actuator {
+  // Target value of the actuator
+  // NOTE: This is what you set in order to command the the actuator
+  //       to move.
+  Value target = 30;
+}
+
+////////////////////////
+// Sensor specific
+message Sensor {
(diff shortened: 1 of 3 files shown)


proto/kuksabase.proto
```diff
@@ -0,0 +1,128 @@
...
+//some datapoints could be delivered we shall agree on another code
+//202? 206? Suggestions? 210?
+message SetRequest {
+  RequestType type = 1;
+  repeated Value values = 2;
+}
+
+//Sets shall have a global status only
+message SetResponse {
+  Status status = 1;
```
nayakned, line 72: Could we have an optional array of values to return the failures of the individual set requests?
If all the requests went through, we can use only the _global_ status. If not, then the ones which failed can be returned in the array of values along with their corresponding failures.

proto/kuksabase.proto
```diff
@@ -0,0 +1,128 @@
...
+See also https://github.com/eclipse/kuksa.val/blob/master/doc/jwt.md
+Upon reception of auth-token, server shall generate an auth_id in metadata
+that the client can use instead of auth_token in subsequent calls
+*/
+
+// The connecting service definition.
+service kuksa_val {
+  // Returns a list of requested data points.
+  rpc get (GetRequest) returns (GetResponse) {}
+  rpc set (SetRequest) returns (SetResponse) {}
```
nayakned, line 36: This means the feeders call the one RPC per datapoint and not use a streaming call. Right?

SebastianSchildt: Yes, see my comment in the original post. 
Not sure we should integrate streaming (it seems to make sense for some high-frequency, possibly single datapoint sources), or not (none of our open stuff uses it, and GRPC is pretty efficient anyway, does it make a difference)

SebastianSchildt: I vote: Lets keep this simple first. Set is setting ONE dp. We can always come back and optimize

SebastianSchildt:
I modifed it, and moved more stuff to the types.proto.

I also do hope it is more consistent now, and the protoc compiler does not complain

nayakned (approved):
Looks good for atleast now.

SebastianSchildt:
Added some input/inspiration from discussion and branches at https://github.com/boschglobal/kuksa.val/tree/feature/unified_proto and https://github.com/boschglobal/kuksa.val/tree/feature/unified_proto_fieldmask

Some rationale for decisions/changes
 - in the end decided to not use fieldmasks, for two reasons
     - I assume they would be needed often (I think most apps will want to restrict information they get/query specifically for what they need), and that is not the most efficient thing to do doing all thos handling of strings on the wire and in grpc libs
     - The approach now is defintely "less generic", and more tied to the VSS view of the world. I think that is good int his case
 - Modelled a Datapoint more completely following VSS
 - Used "oneof" pattern for optional fields
 - Did not add the "Streaming set" functions -> Currently we have literally almost no idea about our performance, or even the requirements, or if we are even the bottleneck. This can always be added later 
 - Did not differentiate between "Single Set/Get" and "Batch Set/Get", I think the functions now cover everything, and we can always do the "single without needing to create a list" sugar-coating in the SDK

proto/kuksa/val/v1/types.proto
```diff
@@ -0,0 +1,271 @@
...
+
+enum FieldType {
+  CURRENT_VALUE = 0;  // this is default, every implementation shall support this
+  TARGET_VALUE  = 1;  // this is a set point of an actuator
+  DATATYPE      = 2;  // VSS datatype of an element as string ("branch" in case of a VSS branch)
+  UNIT          = 3;   // VSS unit of an element as string
+  CHILDREN      = 4;   // Names of all direct children as string
+  ALL_METADATA  = 100; // All available metadata defined in VSS model
+  ALL           = 200; // All available data
+}
```
argerus, line 225: I think it would be better to have a 1:1 mapping between fields and this enum, i.e.

```
message Datapoint {
    Value value = 2;          // Field = VALUE
    DataType data_type = 3;   // Field = DATA_TYPE
    string description = 4;   // Field = DESCRIPTION
    string comment = 10;      // Field = COMMENT
    string deprecation = 11;  // Field = DEPRECATION
    string unit = 12;         // Field = UNIT
    ValueRestriction value_restriction = 13; // Field = VALUE_RESTRICTION
    Actuator actuator   = 30; // Field = ACTUATOR
    Sensor sensor       = 31; // Field = SENSOR
    Attribute attribute = 32; // Field = ATTRIBUTE
}

message Actuator {
    Value target = 1; // Field = ACTUATOR_TARGET
}
```

and

```suggestion
enum Field {
  VALUE = 0;  // this is default, every implementation shall support this
  ACTUATOR_TARGET = 1;  // this is a set point of an actuator
  DATA_TYPE      = 2;  // VSS datatype of an element as string ("branch" in case of a   VSS branch)
  UNIT          = 3;   // VSS unit of an element as string
  DESCRIPTION = 4;
  COMMENT = 5;
  DEPRECATION = 6;
  VALUE_RESTRICTION = 7;
  ACTUATOR = 8;
  SENSOR = 9;
  ATTRIBUTE = 10;

  ALL_METADATA  = 100; // All available metadata defined in VSS model
  ALL           = 200; // All available data

}
```

argerus (changes requested):

proto/kuksa/val/v1/base.proto
```diff
@@ -64,17 +64,16 @@ message GetRequest {
 
 // Status is overall status, there may be
```
erikbosch, line 65: Replace `Status` with `Error`?

proto/kuksa/val/v1/types.proto
```diff
@@ -214,28 +214,42 @@ enum DataType {
...
-  string reason             = 2;
-  string status_description = 3;
+// shall follow https://www.w3.org/TR/viss2-transport/#status-codes where it makes sense
+message Error {
+  uint32 code        = 1;
+  string reason      = 2;
+  string description = 3;
 }
 
 // used in get/set requests to optionally report status on indivusal paths
```
erikbosch, line 249: `indivusal` 

And remove all instances of status in the document.

SebastianSchildt:
Squashed it, and will merge. 
Any further changes we will consider "bugfixes". Let's start implementing this

PR #324, opened 2022-08-09
Update high level architecture intro

SebastianSchildt:
This updates the current high-level system architecture documents, doing the following

 - mention GRPC alongside VSS
 - Using the new official VSS logo
 - Making it databroker/val-server agnostic, just always talking about the "kuksa.val" component

In case the diff is not easy to read, here is how it would/will look like
https://github.com/boschresearch/kuksa.val/blob/feature/update_arch_pictures/doc/system-architecture.md

ECA check will probably fail because it was done using drawio, but I will squash accordingly

Please also give it a quick read through @argerus


argerus:
Looks good to me.

argerus (approved):

PR #329, opened 2022-08-16
Added a switch -t/--type for forcing gRPC client to use a particular datatype

nayakned:
You can force the kuksa client to set values using a specified datatype while using the gRPC interface.


nayakned:
@SebastianSchildt While the client can enforce the data type, the server checks for what it expects and reads from the corresponding field only, i.e., even if client sends an int32, the server reads the float field if the datapoint is supposed to be a float.

How do you intend to handle this in the kuksa-val-server.

SebastianSchildt:
Closing this, as this issue is solved as part of the new/upcoming kuksa-client work

PR #331, opened 2022-08-18
Include VSS 3.0 and use it as default

### LINKED ISSUE #322: Integrate VSS 3.0 release
VSS 3.0 https://github.com/COVESA/vehicle_signal_specification/releases/tag/v3.0

is released, we should 

- Add the JSON version to this repo in https://github.com/eclipse/kuksa.val/tree/master/data/vss-core
- Fix the cmake, docker and usage output of kuksa-val-server to point this file as default
- Make sure it is loaded "as default" when running databroker as well
  - to be defined. Include in Docker? Load upon startup or in the example Readme

erikbosch:
Fixes #322

Support for loading by default in databroker not yet added


data/vss-core/README.md
```diff
@@ -0,0 +1,33 @@
...
+* [VSS 2.2](https://github.com/COVESA/vehicle_signal_specification/releases/tag/v2.2)
+* [VSS 2.1](https://github.com/COVESA/vehicle_signal_specification/releases/tag/v2.1)
+* [VSS 2.0](https://github.com/COVESA/vehicle_signal_specification/releases/tag/v2.0)
+
+## Change process
+
+This is the process for introducing support for a new VSS version:
+
+* Copy the new json file to this folder
+* Add a new contants to `Vssdatabase_record.h` of the format `vss_X_Y_supported = true`
```
SebastianSchildt, line 21: I think that header is not needed anymore, so remove from docs

doc/usage.md
```diff
@@ -16,17 +16,17 @@ OPTIONS:
                                         easily (e.g. test and production 
                                         setup).Sample of configuration file 
                                         parameters looks like:
-                                        vss = vss_release_2.2.json
+                                        vss = vss_rel_3.0.json
```
SebastianSchildt, line 19: I think we should name the releases consistenly. Everything is called "release" and 3.0 only "rel". I suggest changin it to release as well.  Will not mark all instances of this

SebastianSchildt (changes requested):
Generally works for me. Also I would suggest to merge this and make databroker changes in a seperate PR. 
Other than that: Check comments

doc/usage.md
erikbosch: My original idea for changing was to use the same naming as in VSS, i.e. same name as in https://github.com/COVESA/vehicle_signal_specification/releases/download/v3.0/vss_rel_3.0.json, and keep the name for the others to do not cause troubles for anyone referring to them.

But I have no problem using the existing name pattern.

erikbosch:
Following scenarios now tested with VSS 3.0 update

- val-server with viss-client
- val-server with dbc2val
- val-server with gps2val
- val-server unittest
- databroker with cli
- databroker with dbc2val

doc/usage.md
SebastianSchildt: Let's use "release", looks nicer, and we do not need save bytes here

SebastianSchildt (approved):
Works.
lgtm 👯

PR #332, opened 2022-08-23
Add default VSS file to Dockerfile

erikbosch:
With this PR the databroker container will by default load the VSS 3.0 file

```
erik@debian3:~/kuksa.val$ docker run --rm -it  -p 55555:55555/tcp databroker
2022-08-23T08:45:05.334442Z  INFO databroker: Init logging from RUST_LOG (environment variable not found)
2022-08-23T08:45:05.334473Z  INFO databroker: Starting Kuksa Data Broker 0.17.0
2022-08-23T08:45:05.334480Z  INFO databroker: Populating metadata from file 'vss_release_3.0.json'
2022-08-23T08:45:05.339142Z  INFO databroker: Listening on 0.0.0.0:55555
2022-08-23T08:45:05.339172Z  INFO databroker::broker: Starting housekeeping task
^C2022-08-23T08:45:12.903682Z  INFO databroker: received SIGINT
2022-08-23T08:45:12.903722Z  INFO databroker::grpc_service: Shutting down

```

### DIFF
--- a/kuksa_databroker/Dockerfile
+++ b/kuksa_databroker/Dockerfile
@@ -27,9 +27,13 @@ FROM --platform=$TARGETPLATFORM debian:buster-slim
 
 COPY --from=builder workdir/databroker /app/databroker
 
+COPY ./data/vss-core/vss_release_3.0.json vss_release_3.0.json
+
 ENV KUKSA_DATA_BROKER_ADDR=0.0.0.0
 ENV KUKSA_DATA_BROKER_PORT=55555
+ENV KUKSA_DATA_BROKER_METADATA_FILE=vss_release_3.0.json
 
 EXPOSE $KUKSA_DATA_BROKER_PORT
 
 ENTRYPOINT [ "/app/databroker" ]
+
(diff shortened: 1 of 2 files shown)


argerus:
> So if arguments needs to be given then also the VSS 3.0 file needs to be included as argument

Instead of adding it as an argument in the docker file, you could set a default `ENV` for it.

That would make it:
1. possible to add additional arguments (without adding `--metadata` manually again).
2. easy to remove the default by manually setting an empty `KUKSA_DATA_BROKER_METADATA_FILE`.

kuksa_databroker/Dockerfile
```diff
@@ -27,9 +27,13 @@ FROM --platform=$TARGETPLATFORM debian:buster-slim
...
+COPY ./data/vss-core/vss_release_3.0.json vss_release_3.0.json
+
 ENV KUKSA_DATA_BROKER_ADDR=0.0.0.0
 ENV KUKSA_DATA_BROKER_PORT=55555
 
 EXPOSE $KUKSA_DATA_BROKER_PORT
 
 ENTRYPOINT [ "/app/databroker" ]
+
+CMD ["--metadata", "vss_release_3.0.json"]
```
argerus, line 39: Instead of adding it as an argument in the docker file, you could set a default ENV for it.

That would make it:

1. possible to add additional arguments (without adding --metadata manually again).
2. easy to remove the default by manually setting an empty KUKSA_DATA_BROKER_METADATA_FILE.

argerus (changes requested):

kuksa_databroker/Dockerfile
erikbosch: Changed accordingly

argerus (approved):

PR #335, opened 2022-08-26
Update kuksa.val.feeders version

### LINKED ISSUE #322: Integrate VSS 3.0 release

erikbosch:
To include VSS 3.0 adaptions

Fixes #322 


SebastianSchildt (approved):
👍

PR #336, opened 2022-08-26
GRPC proto cleanup

argerus:
1. #### Move metadata fields to `.metadata` in `Datapoint`.
    This makes it more clear what `VIEW_METADATA` refers to and it makes it possible to have `FIELD_METADATA`.

2. #### Add `fields` in addition to `view` in `Get` & `Subscribe`
    This makes it possible to subscribe to single fields. It's not really needed for `Get` but it seems consistent to add it to both.

    Because of this I also removed for example `VIEW_UNIT` since it's now equally easy to add `FIELD_METADATA_UNIT` to `fields`. In general I think `view` is a good solution to having a default view and a convenient way to request multiple fields. But perhaps not for specifying individual fields.

4. #### Include which fields have changed in subscription replies.
    In order for a client to tell which field has changed (e.g. if it subscribes to both `value` and `actuator_target`).

5. #### Various cleanup
    * Fix typos, prefix all enums, update documentation, removed references to `Status`
    * Add [field: FIELD_NAME] to all fields to make it super obvious which is which.

### Not included
* I still think there should be some way to specify that a value is not available. Compare with Android VHAL, that has [`VehiclePropertyStatus`](https://source.android.com/docs/devices/automotive/vhal/properties#property-status) for every `VehiclePropertyValue` which can be `AVAILABLE` `NOT AVAILABLE` or `ERROR`. And databroker which has the same concept embedded in the `oneof value`. The point is that a missing value isn't even necessarily an error and it should still be possible to at least `Get` and `Subscribe` to such a `Datapoint`.

### DIFF
--- a/proto/kuksa/val/v1/types.proto
+++ b/proto/kuksa/val/v1/types.proto
@@ -11,8 +11,10 @@ syntax = "proto3";
 package kuksa.val.v1;
 import "google/protobuf/timestamp.proto";
 
+// Should probably be changed to something like:
+// option go_package = "github.com/eclipse/kuksa.val/proto/v1";
+option go_package = "/kuksa_grpc_proto";
 
-// A Value. 
 message Value {
   google.protobuf.Timestamp timestamp = 1;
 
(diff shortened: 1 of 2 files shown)


SebastianSchildt (approved):
Argh, this was supposed to be already in. Must have messed sth up with git...

PR #341, opened 2022-08-31
Can parse defaults, but does not yet apply them

### LINKED ISSUE #339: Attritbutes not correctly populated in databroker when load VSS json
start current databroker

```
2022-08-31T07:24:50.689849Z  INFO databroker: Init logging from RUST_LOG (environment variable not found)
2022-08-31T07:24:50.689872Z  INFO databroker: Starting Kuksa Data Broker
2022-08-31T07:24:50.689930Z  INFO databroker: Populating metadata from file 'vss_release_3.0.json'
2022-08-31T07:24:50.693057Z  INFO databroker: Listening on 0.0.0.0:55555
2022-08-31T07:24:50.693074Z  INFO databroker::broker: Starting housekeeping task
```

query VSS version

```
scs2rng@RNG-C-001JT kuksa.val % ./target/release/databroker-cli
client> get Vehicle.VersionVSS.Major
-> Vehicle.VersionVSS.Major: ( NotAvailable )
client>
```
 Expected answer

```
-> Vehicle.VersionVSS.Major: 3
```

SebastianSchildt:
An _incomplete_ attempt to fix #339 as the Rust is weak within me.

What it currently does: Parsing default from JSON (keeping it as an "untyped" serde JSON value at that point), and during registering checking whether a default exist, and calling a function to apply it.

The challenge is, now the apply_default function would need a lot of code to convert the datatypes to the correct datapoints. Question @argerus , would that even be the right way to go? And if so, is now the right time, or will the new GRPC implementation change the required functions and structs a lot anyway?

### DIFF
--- a/kuksa_databroker/databroker/src/main.rs
+++ b/kuksa_databroker/databroker/src/main.rs
@@ -24,7 +24,6 @@ use clap::{Arg, Command};
 
 use databroker::grpc_service;
 use databroker::types;
-use databroker::broker::{DataBroker};
 use databroker::{broker, vss};
 
 // Hardcoded datapoints
@@ -104,22 +103,6 @@ fn init_logging() {
         .expect("Unable to install global logging subscriber");
 
     info!("{}", output);
-}
-
-async fn apply_default(id: i32, val: &serde_json::Value, broker : &DataBroker) {
-    let res = broker.set_datapoints(&[( //This need logic for all the types :-/
-        id,
-        broker::DataPoint {
-            ts:  std::time::SystemTime::now(), 
-            value: types::DataValue::String("he".to_string()),
-        },
-    )])
-    .await;
-
-    match res {
-        Ok(_t) => info!("Yeay"),
-        Err(_e) => info!("Error applying default"), //No easy way to print error
-    }
 }
 
 async fn shutdown_handler() {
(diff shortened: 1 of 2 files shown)


kuksa_databroker/databroker/src/main.rs
```diff
@@ -105,6 +106,22 @@ fn init_logging() {
     info!("{}", output);
 }
 
+async fn apply_default(id: i32, val: &serde_json::Value, broker : &DataBroker) {
```
SebastianSchildt, line 109: This function would need to get _a lot_ longer, getting access to metadata and creating the correct "DataValue"

SebastianSchildt:
@argerus  feel free to push into this branch, or steal form it, if you can fix it easily.... 

Not working on it currently

kuksa_databroker/databroker/src/main.rs
argerus: Moved the conversion `serde_json::Value` -> `DataValue` directly into the parser instead.

argerus:
> @argerus feel free to push into this branch, or steal form it, if you can fix it easily....
> 
> Not working on it currently

Should be good now.

SebastianSchildt:
Works. Lets merge this. One bug down/one feature up :)

Issue #334, opened 2022-08-25
No master image for databroker


SebastianSchildt:
We should make sure, databroker is is also pushing a master image to ghrc, i.e. for kuksa.val server we have

```
docker pull ghcr.io/eclipse/kuksa.val/kuksa-val:master
```

for databroker we do not have 

```
docker pull ghcr.io/eclipse/kuksa.val/databroker:master
```

Probably need some adaptation in GH action workflow/trigger

SebastianSchildt:
@argerus would assign you, if I could. PLs PR

SebastianSchildt:
Fixed

Issue #337, opened 2022-08-30
Things to do for next release


SebastianSchildt:
We shall have a new release after the unified API is implemented in databroker and clients/feeder support it (existing kuksa.val server userland without loosing support for server).

We will sync release version (i.e. not different versions for different releases, we'd rather make a new minor release if something important changes, and not which components didn't change)

As databroker is 0.1.7 currently and val-server at 0.2.1 , **next release shall be 0.3**

This is to list/track everything that needs to be done, before we can call it a release

- [x] Unified API implemented in all relevant components
   - [x] databroker
   - [x] Python SDK
   - [x] Go SDK
 - [x] Feeder 
   - [x] CAN Feeder
   - [x] GPS Feeder 
   - [x] (if exists) SomeIP Feeder work with val-server and databroker 
 - [x ] Python SDK renamed to kuksa-client #318 

 - [x] Release 0.3 tagged/released
    - [x]  0.3.0 tag exists
    - [x] docker pull ghcr.io/eclipse/kuksa.val/databroker:0.3.0 image exists and works
    - [x] docker pull ghcr.io/eclipse/kuksa.val/kuksa-val:0.3.0 image exists and works
    - [x] docker pull ghcr.io/eclipse/kuksa.val/kuksa-client:0.3.0 image exists and works
    - [x] Kuksa client package with version 0.3.0 exists in PyPi
    - [x] Created a "release" in GH, summarizing changes and links to the above mentioned images

SebastianSchildt:
Release done https://github.com/eclipse/kuksa.val/releases/tag/0.3.0

PR #344, opened 2022-09-01
Complete dockerfile that pushes directly for databroker.

SebastianSchildt:
Should also fix the issue that our current databroker images have no manifest and thus cannot be used in LEDA

Need to test inside this repo....


SebastianSchildt:
This follows the kuksa-val approach to naming/tagging, so it is also a prerequisite for #337 
Might do a an intermediate release on the 0.2. area, so Leda can have a "stable" tag, instead of master 
0.1.7 does not work for them due to missing mainfest in container image, but I won't overwrite that tag (as Velocitas is using it)

SebastianSchildt:
Will fix #334

PR #348, opened 2022-09-02
Limit contents copied when building kuksa-client container

erikbosch:
This is a minor improvement that could save some time for local builds of the kuksa-client docker container.
Without this change all content of current working directory is copied. That is typically not that much if starting on fresh clone, but if you have built e.g. databroker before you may copy a lot of unused data.

It has been verified that building and running containers works with this change.

COPY is typically preferred over DD, see https://docs.docker.com/develop/develop-images/dockerfile_best-practices/


SebastianSchildt (approved):
Tested, works. lgtm.

PR #349, opened 2022-09-02
Use Eclipse as organization for example certificates

erikbosch:
Contain following changes:

- Eclipse used instead of Bosch as "Issued By" in certification generation script. Fields contents taken from regular Eclipse certificates, except e-mail
- Regeneration of certificates to reflect the change in the script
- Minor change in script to regenerate CA.pem even if CA.key exists, and to support generation of Client/Server certificate without replacing keys.


SebastianSchildt:
Is is expected, that it can not connect to a previous version of kuksa.val server even though the keys themselves did not change.

had a current master kuksa.val server running, and built docker form this brach locally 

```
al % docker run -it --net=host kuksaclient:latest
Welcome to kuksa viss client version 0.17.0-88-dirty

                  `-:+o/shhhs+:`                  
                ./oo/+o/``.-:ohhs-                
              `/o+-  /o/  `..  :yho`              
              +o/    /o/  oho    ohy`             
             :o+     /o/`+hh.     sh+             
             +o:     /oo+o+`      /hy             
             +o:     /o+/oo-      +hs             
             .oo`    oho `oo-    .hh:             
              :oo.   oho  -+:   -hh/              
               .+o+-`oho     `:shy-               
                 ./o/ohy//+oyhho-                 
                    `-/+oo+/:.             

Default tokens directory: /kuksa_viss_client/kuksa_viss_client/../kuksa_certificates/jwt

connect to wss://127.0.0.1:8090
Disconnected!! [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: unable to get local issuer certificate (_ssl.c:997)
Error: Websocket could not be connected or the gRPC channel could not be created.
Server disconnected.
Test Client> 
```

erikbosch:
My understanding of certificate authentication is that is that OpenSSL in the client will check `Issuer` in the presented server certificate, and search for a local "root certificates" with matching `Subject`. So if the server presents a certificate issued by "Bosch", it will not check if the public key of a root certificate where subject contains "Eclipse" matches it.

We could theoretically - for a period of time - keep the old "BOSCH" CA certificate in parallel to the new one, but I do not  know if build scripts then would install both

SebastianSchildt:
Let's merge this after we have a last "0.2.x" release (then it will only break for 0.3 which is next larger release)
See #361 and #337

SebastianSchildt:
I think, now would be a good time to merge this. We had our 0.2.5 release, and are reworking our python lib & clients anyway

The effect will be that after the merge the default  certificates and tokens in master will not work with the released versions (e.g. you can not mix and match during runtime without manually supplying either side with the new or old certs)

Any opinions @erikbosch @romainletendart ?

erikbosch:
@SebastianSchildt - updated script (and generated CA.pem) so the CA.pem in this PR is valid to Oct 29, 2032 (3650 days)

SebastianSchildt (approved):
BXC is done, API work progressing, databroker still waiting for TLS. what could possibly go wrong.
Pressing button in 3,2,1 ...

PR #350, opened 2022-09-02
Improve kuksa-client documentation

### LINKED ISSUE #255: Hard to start local kuksa_viss client when an older version is installed in system
If a system has an installed version of kuksa_viss_client, and in a source directory you try to start it amnually doing

```
python3 __main__.py
```

You get (partially) the system installed version. I think this is due to to the search paths defaulting to system first.

A fix __seems__ to be to replace 

https://github.com/eclipse/kuksa.val/blob/c3739d5c6f1ef68a16cebca52c14670858ec48a8/kuksa_viss_client/__main__.py#L23

with 

```python
sys.path.insert(0,os.path.join(scriptDir, ".."))
```

Not sure this has negative implications when isntalled system wide

erikbosch:
Fixes #255

### DIFF
--- a/kuksa_viss_client/README.md
+++ b/kuksa_viss_client/README.md
@@ -4,26 +4,64 @@
 `kuksa.val` is a part of the opensource project [eclipse kuksa](https://www.eclipse.org/kuksa/).
 More about `kuksa.val` can be found under the root [README.md](../README.md).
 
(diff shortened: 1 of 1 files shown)


SebastianSchildt (approved):
👍 🥇

PR #358, opened 2022-09-06
Add kuksa.val.v1 to databroker + small update to .protos

argerus:
### Rational for .proto updates
- In VSS, the "thing" found at a specific VSS path is called a data entry.
- In VISS, the value + timestamp is grouped inside "dp", i.e. a datapoint.

The previous usage of "DataPoint" (in our .protos) to refer to the whole entry, led to confusion in variable naming when implementing it. To solve this, and in order to align with the terminology used by both VSS and VISS, the previous uses of datapoint ( or `Datapoint`) in the .protos was changed to data entry, `DataEntry` or simply entry.

Also part of this PR is the removal of all `oneof optional_* {...}` since it turned out that Rust (or the lib prost) actually did support the usage of `optional`. All occurrences have been replaced with `optional` instead.


proto/kuksa/val/v1/types.proto
```diff
@@ -15,7 +15,24 @@ import "google/protobuf/timestamp.proto";
...
+// be controlled by specifying either a `View` or a set of `Field`s.
+message DataEntry {
+  // Defines the full VSS path of the entry.
+  string path = 1;  // [field: FIELD_PATH]
+
+  // The value (datapoint)
+  Datapoint value = 2;  // [field: FIELD_VALUE]
+
+  // Actuator target (only used if the entry is an actuator)
+  Datapoint actuator_target = 3;  // [field: FIELD_ACTUATOR_TARGET]
```
SebastianSchildt, line 29: Schould this also be optional?

argerus: It would make sense intuitively, since the field is optional.

However, `optional` only makes sense for primitive types, e.g. `int32`, since the point of having it is to be able to track if the if a field has been set. With proto3, this is not possible for primitive types without adding `optional`.

But since `Datapoint` isn't a primitive type, it's possible to see if the field has been set anyway.

See https://github.com/protocolbuffers/protobuf/blob/v3.12.0/docs/field_presence.md

SebastianSchildt: Learned something new (or relearned something old?). Thanks, makes sense!

erikbosch:
I guess the databroker build error is because the proto files have not been added to the Dockerfile

`Error: Custom { kind: Other, error: "protoc failed: Could not make proto path relative: proto/kuksa/val/v1/val.proto: No such file or directory\n" }`

SebastianSchildt:
> I guess the databroker build error is because the proto files have not been added to the Dockerfile
> 
> `Error: Custom { kind: Other, error: "protoc failed: Could not make proto path relative: proto/kuksa/val/v1/val.proto: No such file or directory\n" }`

In that case, testing locally probably has faster roundtrips than waiting for CI. Just doing
```
docker build -t databroker:test -f kuksa_databroker/Dockerfile .
```
initiates a build of the databroker container (No need for buildx for local testing)

SebastianSchildt (approved):
Looking good, and stuff still works.Will merge after merging and tagging #361

PR #360, opened 2022-09-06
Fix license classifier for Pypi

SebastianSchildt:
Fixes the license classifier for Python, otherwise Pypi does not accept the Python package

Note: The used classifier DOES refer to the Apache 2.0 license even though the official string does omit "2.0".
See here for many examples for Apache 2.0 licensed python packages: https://pypi.org/search/?c=License+%3A%3A+OSI+Approved+%3A%3A+Apache+Software+License


eriksven (approved):
Works for in accordance with the PyPi documentation. It is indeed a bit confusing that PyPi omits the 2.0.

PR #361, opened 2022-09-06
Fix/python package

SebastianSchildt:
I was trying to build/push and updated python package for an intermediate release.

It seemed, the moving of the client lib when we added databroker, made the generated python wheel broken (mainly due to paths and what was included where).

This PR should fix that, hopefully without breaking the Docker build or local execution


kuksa_viss_client/README.md
```diff
@@ -1,12 +1,12 @@
 # Kuksa VISS Client
-![kuksa.val Logo](../doc/pictures/logo.png)
+![kuksa.val Logo](https://raw.githubusercontent.com/eclipse/kuksa.val/0.2.5/doc/pictures/logo.png)
```
erikbosch, line 2: Why/When is absolute path needed? Is it because the relative paths will not work if you just install Python source package? Why do we not refer to https://github.com/eclipse/kuksa.val/blob/master/doc/pictures/logo.png? Do we know for how long v0.2.5 will remain, having to update version number as part of every release is cumbersome, and can we actually update the reference before the release?

kuksa_viss_client/README.md
```diff
@@ -94,11 +94,11 @@ updateVISSTree      Update VISS Tree Entry
 
 This is an example showing how some of the commands can be used:
 
-![try kuksa_viss_client out](../doc/pictures/testclient_basic.gif "test client usage")
+![try kuksa_viss_client out](https://raw.githubusercontent.com/eclipse/kuksa.val/0.2.5/doc/pictures/testclient_basic.gif "test client usage")
 
 ### Updating VSS Structure
 
-Using the testclient, it is also possible to update and extend the VSS data structure. More details can be found [here](../doc/liveUpdateVSSTree.md).
+Using the testclient, it is also possible to update and extend the VSS data structure. More details can be found [here](https://raw.githubusercontent.com/eclipse/kuksa.val/doc/liveUpdateVSSTree.md).
```
erikbosch, line 101: Link does not work, no 0.2.5 used

kuksa_viss_client/README.md
SebastianSchildt: Absolute path do not work on Pypi, where the docs are also used compare old version
https://pypi.org/project/kuksa-viss-client/ where links are broken
and test
https://test.pypi.org/project/kuksa-viss-client/0.2.5.post2/

I try to refer to the tag where possible, so when we move stuff around, pages from older pypi version still look ok (and the tag should be there  to stay in GH)

erikbosch: > Absolute path do not work on Pypi

I assume you mean relative path?

Do we by the way have any release instructions for Kuksa.VAL, i.e. what to do when creating a new release? Concerning e.g. update of versions number. If so that checklist could mention that this number does not need to be updated for every new release, but it needs to be checked that the link still works.

SebastianSchildt: Jupp, and I we do not have release instructions. yet.

PR #364, opened 2022-09-09
[databroker] Implement kuksa.val.v1 service

argerus:
- Add `kuksa.val.v1` service implementation to databroker.
- Add `SetDatapoints` to `sdv.databroker.v1.Broker` interface.


erikbosch:
Hi @argerus 

I did a smoke test to verify that there are no regressions using dbc-feeder and it worked well.

What we lack - as I see it - is to update the README. A few ideas:

- Extend the [Interface section](https://github.com/eclipse/kuksa.val/blob/master/kuksa_databroker/README.md#interface)
- Mention that databroker support 2 interfaces, the legacy SDV Databroker Proto Interface and the new unified Kuksa.VAL Proto interface
- Mention that both are supported on the same port, a client can use whichever it like
- Mention that the databroker-cli (still) uses the legacy API (if I understand it correctly, otherwise state that it use the new API) 
- Ass a summary of the new methods in https://github.com/eclipse/kuksa.val/blob/master/kuksa_databroker/proto/sdv/databroker/v1/broker.proto, similar to what we already have for the legacy API

erikbosch (approved):
Looks good to me

erikbosch:
@romainletendart @int0x27 please review as well!

eriksven (approved):

PR #371, opened 2022-10-06
Fix websocket authorization

erikbosch:
The current kuksa_viss_client for websocket seems to be broken, at least if you use it without a token file, like dbcfeeder.
Then you get "None" as default by `authorize` in `__init__.py` and an exception, as `os.path.expanduser` does not like None.

Adapting to same structure as used for gRPC

This practically reverse [229a63c931065778e21b854a97e3c86f7659c3f9](https://github.com/eclipse/kuksa.val/commit/229a63c931065778e21b854a97e3c86f7659c3f9). (@SebastianSchildt  - could you please check the rationale for that commit, why it should behave differently than for gRPC)

Note that dbc/gps feeders in https://github.com/eclipse/kuksa.val.feeders currently rely on default authorization given by kuksa-viss-client, there is no way to manually override with another token file.

### DIFF
--- a/kuksa-client/kuksa_client/KuksaWsComm.py
+++ b/kuksa-client/kuksa_client/KuksaWsComm.py
@@ -36,6 +36,7 @@ class KuksaWsComm:
         self.cacertificate = config.get('cacertificate', os.path.join(scriptDir, "../kuksa_certificates/CA.pem"))
         self.certificate = config.get('certificate', os.path.join(scriptDir, "../kuksa_certificates/Client.pem"))
         self.keyfile = config.get('key', os.path.join(scriptDir, "../kuksa_certificates/Client.key"))
+        self.tokenfile = config.get('token', os.path.join(scriptDir, "../kuksa_certificates/jwt/all-read-write.json.token"))
         self.wsConnected = False
 
         self.subscriptionCallbacks = {}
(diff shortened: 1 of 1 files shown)


SebastianSchildt:
I do think the intention of that change was, that we did not want the library to magically pull out some token (that would not work in a "real" deployment anyway, where you are hopefully not using the example keys), in case you do not provide one. 

Maybe we should do you revert and make an issue for that?

The DBC feeder at least DOES support other tokens, it can be given in the config file the `[kuksa_val]`section https://github.com/eclipse/kuksa.val.feeders/blob/main/dbc2val/config/dbc_feeder.ini , which is provided verbatime as "config" to the python library. (but looking at the code also might require your revert.....)

aya, we might need more documentation....

erikbosch:
Ah, I was thinking mostly on the call to `authorize()` at https://github.com/eclipse/kuksa.val.feeders/blob/main/dbc2val/dbcfeeder.py#L212 where we do not provide any parameter even if possible, but youn are right, we might configs by the instantiation at https://github.com/eclipse/kuksa.val.feeders/blob/main/dbc2val/dbcfeeder.py#L210 as well.

We could very well argue that kuksa-viss-client (when used as a library) shall not provide any default tokens and certificates at all, they must be provided by the client. I.e. have no default for `self.tokenfile`, but then we should better change behavior so that we give a proper meaningful exception/error like "No token provided" or a warning like "No authorization will be performed".

But that likely require a multi-step approach, first adapting default configs for feeders so they provide a token to the client lib, then in a second phase make the config mandatory.

SebastianSchildt:
@romainletendart This is actually something you can take care of when changing the python lib (because we need to touch all our clients afterwards anyway):  The lib should not "magically" supply a random auth token form somewhere if none is given

erikbosch:
Can be closed when #416 is merged

SebastianSchildt:
#416 was merged

PR #378, opened 2022-10-12
NOT FOR MERGE: Testing build without Alpine

erikbosch:
This is an example pull request showing a possible to change to get rid of Alpine.

At least for kuksa-viss-client (Python) this is beneficial as then prebuilt versions of grpcio can be used.
For kuksa-val-server (C++) the difference is more subtle as we anyway need to build grpicio manually

### DIFF
--- a/kuksa-val-server/docker/Dockerfile
+++ b/kuksa-val-server/docker/Dockerfile
@@ -5,23 +5,14 @@
 # accompanies this distribution, and is available at
 # http://www.eclipse.org/org/documents/edl-v10.php
 
-FROM alpine:3.11 as build
+FROM python:3.10-slim-bullseye as build
 
 LABEL maintainer="Sebastian Schildt <sebastian.schildt@de.bosch.com>"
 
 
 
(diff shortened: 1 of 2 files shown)


SebastianSchildt:
If this goes through, compare sizes of the  container  image . Generally for target hardware, a longer build time once should be worth it, if the artifact gets (significantly) smaller that way.
(although then we might need to take a look at out CI again, currently not sure if the default runner uses a Pi or a Graviton instance to offload the ARM builds, in case it is the former, the latter should speed things up a little bit :))

erikbosch:
Did a size comparison. With slim the dockerfile seems to become 164 MB compared to around 94 MB with Alpine. So indeed a big change.

I do not know if there might be alternative solutions, like building grpcio on top of alpine and store/publish it so we easily could reuse it to avoid rebuilds.

SebastianSchildt:
Generally we should invest build time for smaller binaries. Alpine+GRPC is currently the most painful combination, but I think the better way to tackle this, is looking at our CI infrastructure

erikbosch:
Closing this one for now

PR #379, opened 2022-10-12
*: Rename kuksa-viss-client kuksa-client

### LINKED ISSUE #318: Rename kuksa-viss-client to kuksa-client
KUKSA currently supports VISS and GRPC, to reflect that, we should rename it. 
for this issue, rename the package/folder here. make sure testclient works, and CI still builds.
No need to adapt feeders

romainletendart:
Package directory and name shall use a dash as per https://pep8.org/#package-and-module-names.
Entry point shall also use a dash as it is common for UNIX programs. The imported python package has to use an underscore however otherwise we cannot import it (SyntaxError).

Fixes #318.


data/vss-core/README.md
```diff
@@ -32,28 +32,28 @@ This is the process for introducing support for a new VSS version:
...
+             :o+     /o/`+hh.     sh+
+             +o:     /oo+o+`      /hy
+             +o:     /o+/oo-      +hs
+             .oo`    oho `oo-    .hh:
+              :oo.   oho  -+:   -hh/
+               .+o+-`oho     `:shy-
+                 ./o/ohy//+oyhho-
+                    `-/+oo+/:.
 
 Default tokens directory: /home/erik/.local/lib/python3.9/site-packages/kuksa_certificates/jwt
```
SebastianSchildt, line 58: Not an issue of this PR, but we might want to change this later

doc/pictures/testclient_basic.cast
```diff
@@ -1,22 +1,22 @@
 {"version": 2, "width": 80, "height": 33, "ts": 1599119755, "env": {"SHELL": "/usr/bin/fish", "TERM": "rxvt-unicode-256color"}}
 [0.028694, "o", "\u001b]0;fish /home/wenwen/tools/kuksa.val"]
```
SebastianSchildt, line 2: If we change these can we at least also re-render them? See https://github.com/eclipse/kuksa.val/blob/master/doc/pictures/README.md

(shall we also redo them (e.g. record fresh .casts) maybe, but if not super easy maybe not in this PR)

SebastianSchildt (changes requested):
I tired the built docker and it still worked (the VISS part at least). Changes look good, I do think at the very least we should generate updated gifs from the modified .cast fiels (see comment above). Otherwise I'd rather merge this fast and "fail forward", because this PR touches a lot of files, and keeping it open too long will make things painful...

erikbosch (approved):
Looks good to me if what Sebastian points out is fixed.

(Maybe we should create some release instruction somewhere where it is mentioned that it shall be reviewed if the examples are still sufficiently relevant, or if they need to be regenerated)

doc/pictures/testclient_basic.cast
romainletendart: I re-rendered the GIFs.
Let's keep re-recording for later I would say.

PR #380, opened 2022-10-13
Remove old Jenkinsfile as al CI has been migrated to GH actions

SebastianSchildt:
As the title says.

Jenkins builds are redundant to GH actions and often fail due to our Jenkins setup and hdd-space starved master. 

This removes the Jenkinsfile, so we only rely on GH actions and can retire the Jenkins infrastructure


eriksven (approved):
makes sense to avoid duplicate infrastructure

PR #381, opened 2022-10-14
Fix some broken links in documentation

lukasmittag:
Like the titel says fix broken links:
* Update links to kuksa data feeders
* Remove non existing part: web-client
* Update links to VSS documentation


SebastianSchildt (approved):
Looking good 👍

PR #382, opened 2022-10-17
Make proto handling more robust, improve gRPC client 'connected' state and make project structure more pythonic

romainletendart:
This is mostly a preparation work before implementing the new kuksa.val.v1 protobuf API.
It addresses the following:

    General project strucuture: split the project in three packages: client, certificates and protobuf definitions.
    Misleading CLI feedback for the gRPC client stating that the client is connected when it's not.
    Using absolute imports instead of relative ones for generated python stubs as protoc does not handle well the latter.


SebastianSchildt:
I could install the package locally in a pipenv.

I am however seem to miss some information how I would use the cmdline client with databroker.

I did start databroker (port 55556 because OS X), and tried using the cleint

```
Test Client> setServerAddress 127.0.0.1 55556 -p grpc
Setting Server Address to 127.0.0.1:55556 with protocol grpc
Test Client> connect 
E1031 12:47:42.557452306      12 ssl_transport_security.cc:1495] Handshake failed with fatal error SSL_ERROR_SSL: error:100000f7:SSL routines:OPENSSL_internal:WRONG_VERSION_NUMBER.
Error: Websocket could not be connected or the gRPC channel could not be created.
gRPC channel disconnected.
E1031 12:47:43.554793473      12 ssl_transport_security.cc:1495] Handshake failed with fatal error SSL_ERROR_SSL: error:100000f7:SSL routines:OPENSSL_internal:WRONG_VERSION_NUMBER.
Test Client> E1031 12:47:45.206505099      15 ssl_transport_security.cc:1495] Handshake failed with fatal error SSL_ERROR_SSL: error:100000f7:SSL routines:OPENSSL_internal:WRONG_VERSION_NUMBER.
```

I may be missing some SSL configuration, but it is not obvious to me currently. `connect -i` does not help, the I just get slightly different errors

```
Exception: <AioRpcError of RPC that terminated with:
	status = StatusCode.UNIMPLEMENTED
	details = ""
	debug_error_string = "{"created":"@1667220486.667720526","description":"Error received from peer ipv4:127.0.0.1:55556","file":"src/core/lib/surface/call.cc","file_line":952,"grpc_message":"","grpc_status":12}"
>
E1031 12:48:07.037177651      12 ssl_transport_security.cc:1495] Handshake failed with fatal error SSL_ERROR_SSL: error:100000f7:SSL routines:OPENSSL_internal:WRONG_VERSION_NUMBER.
```

VISS works fine with latest val-server

SebastianSchildt (approved):
Have been given to understand this ist just restructuring, so the "new" databroker GRPC is not in yet :)
Can conform it still works with kuksa.val server

PR #383, opened 2022-10-17
[databroker] Add initial notification on subscription

argerus:
- Subscription will yield an initial notification containing current values of everything subscribed to.
- Return error if subscription request contains no entries.

### DIFF
--- a/kuksa_databroker/databroker/src/broker.rs
+++ b/kuksa_databroker/databroker/src/broker.rs
@@ -22,12 +22,11 @@ use tokio_stream::Stream;
 
 use std::collections::{HashMap, HashSet};
 use std::convert::TryFrom;
-use std::iter::FromIterator;
 use std::sync::atomic::{AtomicI32, Ordering};
 use std::sync::Arc;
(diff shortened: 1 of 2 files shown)


SebastianSchildt:
I can confirm databroker still works, but as @romainletendart  is working "on the other side" of the new  API , can you give a quick check/thumbs up whether this is working as you expect

romainletendart:
I was able to validate the following:
- Subscription will yield an initial notification containing current values of everything subscribed to.
- Return error if subscription request contains no entries.

However I think we are still missing details of the actual error that occurred.
When a RPC to subscribe() fails, all we get back is a gRPC status code e.g. StatusCode.INVALID_ARGUMENT(3) with details set to "NotFound".
SubscribeResponse does not yet have an Error field which means that only the gRPC error tells us whether something bad happened.
In order to have VISS-compatible error information, I think we would need to have a serialized-as-string Error message behind the 'details' attribute of the gRPC error (see https://github.com/grpc/grpc/tree/master/examples/python/errors).

PR #386, opened 2022-10-30
fix broken link of system-architecture.md

phongt:
Small fixing.


erikbosch (approved):
Looks good to me.

@lukasmittag - could you check so that this fix does not interfere with what you are doing

SebastianSchildt (approved):
Thanks, let's get it merged

(@lukasmittag PR has more "moving parts" and we might still need a day or two, so let's take the quick win first, and if needed repair a tiny conflict)

PR #387, opened 2022-10-31
kuksa_client: Implement new gRPC/proto API

romainletendart:
This PR allows kuksa-client to communicate with kuksa-databroker using latest gRPC/proto API (kuksa.val.v1).


kuksa-client/kuksa_client/cli_backend/grpc.py
```diff
@@ -0,0 +1,204 @@
...
+                entry = kuksa_client.grpc.DataEntry(
+                    path=path, metadata=kuksa_client.grpc.Metadata.from_dict(metadata_dict),
+                )
+            requestArgs = {'updates': (kuksa_client.grpc.EntryUpdate(entry=entry, fields=(field,)),)}
+            return self._sendReceiveMsg(("set", requestArgs), timeout)
+
+        return json.dumps({"error": "Invalid Attribute"})
+
+    # Function for authorization
+    def authorize(self, tokenfile=None, timeout=2):
```
erikbosch, line 110: Guess this is old code that has been moved, but do we know why we have a timeout of 2 here but 5 in most other cases?

erikbosch:
So I tried to do some tests following the README file.

Starting the docker file seems to fail, I assume `logging.ini` isn't included and needs to mentioned in `setup.cfg`

```
erik@debian3:~/kuksa.val/kuksa-client$ docker run --rm -it --net=host kuksa-client:latest
  File "/usr/local/lib/python3.10/runpy.py", line 196, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/usr/local/lib/python3.10/runpy.py", line 86, in _run_code
    exec(code, run_globals)
  File "/kuksa-client/kuksa_client/__main__.py", line 425, in <module>
    sys.exit(main())
  File "/kuksa-client/kuksa_client/__main__.py", line 420, in main
    logging.config.fileConfig(args.logging_config)
  File "/usr/local/lib/python3.10/logging/config.py", line 72, in fileConfig
    formatters = _create_formatters(cp)
  File "/usr/local/lib/python3.10/logging/config.py", line 105, in _create_formatters
    flist = cp["formatters"]["keys"]
  File "/usr/local/lib/python3.10/configparser.py", line 964, in __getitem__
    raise KeyError(key)
KeyError: 'formatters'
```

I also tried to build locally, but there the instructions seems to be outdated:

```
erik@debian3:~/kuksa.val/kuksa-client$ pipenv shell
Launching subshell in virtual environment...
 . /home/erik/.local/share/virtualenvs/kuksa-client-i_dVn3q4/bin/activate
erik@debian3:~/kuksa.val/kuksa-client$  . /home/erik/.local/share/virtualenvs/kuksa-client-i_dVn3q4/bin/activate
(kuksa-client) erik@debian3:~/kuksa.val/kuksa-client$ python __main__.py
python: can't open file '/home/erik/kuksa.val/kuksa-client/__main__.py': [Errno 2] No such file or directory
(kuksa-client) erik@debian3:~/kuksa.val/kuksa-client$ 
```
(This is I believe not related to this PR, but should better be fixed anyway. For both local build and docker run it could maybe be good in the README to show how to select grpc or viss when starting the client)

kuksa-client/kuksa_client/cli_backend/grpc.py
romainletendart: ¯\_(ツ)_/¯
I can set it to 5 to make it consistent with the rest.

erikbosch: Sound like a plan

romainletendart: Just to be sure, I dug into our history and found where that timeout got introduced: https://github.com/eclipse/kuksa.val/commit/2aa41f0332aed104c6ea9b2637c1dd26337aa760#diff-6c3010ee016b743e294b92ebc7fab8bca21b4a1735b194f9907516cba9ac0563R56

However that doesn't give us a clue on why this particular value.

erikbosch: Don't we all love magic numbers.

romainletendart:
@erikbosch I have normally addressed your different comments.
Things should work better now and should also be better documented. ;)

kuksa-client/kuksa_client/cli_backend/__init__.py
```diff
@@ -0,0 +1,46 @@
...
+
+class Backend:
+    def __init__(self, config):
+        self.serverIP = config.get('ip', "127.0.0.1")
+        self.serverPort = config.get('port', 8090)
+        try:
+            self.insecure = config.getboolean('insecure', False)
+        except AttributeError:
+            self.insecure = config.get('insecure', False)
+        with importlib.resources.path('kuksa_certificates', '') as default_cert_path:
```
erikbosch, line 29: At least for me this causes problems if using Python 3.9, but works well if using Python 3.10. Seems the implementation of importlib has changed a bit recently. Maybe not a problem if we do not work on Python 3.9, but do we need to mention it somewhere? (or has anyone else successfully run it on Python 3.9)

```
Default tokens directory: /home/erik/kuksa.val/kuksa_certificates/jwt

  File "/home/erik/kuksa.val/kuksa-client/kuksa_client/cli_backend/__init__.py", line 29, in __init__
    with importlib.resources.path('kuksa_certificates', '') as default_cert_path:
  File "/usr/lib/python3.9/contextlib.py", line 117, in __enter__
    return next(self.gen)
  File "/usr/lib/python3.9/importlib/resources.py", line 175, in _path_from_reader
    opener_reader = reader.open_resource(norm_resource)
  File "<frozen importlib._bootstrap_external>", line 995, in open_resource
IsADirectoryError: [Errno 21] Is a directory: '/home/erik/kuksa.val/kuksa-client/kuksa_certificates'

```

kuksa-client/Dockerfile
```diff
@@ -26,5 +26,7 @@ RUN apk add --no-cache libstdc++
 COPY --from=build /kuksa-client /kuksa-client
 ENV PYTHONUNBUFFERED=yes
 ENV GRPC_ENABLE_FORK_SUPPORT=false
+ENV PYTHONPATH=/kuksa-client
 WORKDIR /kuksa-client
-CMD ["/usr/local/bin/python", "-m", "kuksa_client"]
+CMD ["bin/kuksa-client"]
```
erikbosch, line 31: I think we shall remove CMD now when we have added ENTRYPOINT, otherwise we will get error if trying to start the docker container without giving any parameters

kuksa-client/kuksa_client/cli_backend/__init__.py
romainletendart: Interesting. The only documentation I can find on `importlib.resources.path` 's `resource` parameter reads:

>  resource is the name of the resource to open within package; it may not contain path separators and it may not have sub-resources (i.e. it cannot be a directory).

It basically tells nothing on the expected behaviour when `resource == ''`.
I will look for a documented way of achieving what I want.

kuksa-client/Dockerfile
SebastianSchildt: @erikbosch Did you use --net=host for databroker, and potentially also for the client, if only db is started with net==host note, that the IP you need to connect to is NOT 127.0.0.1 but rather the local ip of your computer, or the ip on the default docker bridge interface (often 172.17.0.1 in Linux). 

Also, in case of MAC: the databroker default port 55555 does not work there. Ever.

erikbosch: Thanks for the hints, I got it working (client in docker towards native broker) in my environment by using `host.docker.internal` . 

`docker run --rm -it kuksa-client:latest --ip host.docker.internal --protocol grpc`

SebastianSchildt (approved):
Lets move the new API forward :)

PR #396, opened 2022-11-07
Update cli auto completion

argerus:


SebastianSchildt (approved):
Makes debugging life easier :)

kuksa_databroker/databroker-cli/src/main.rs
```diff
@@ -574,24 +574,132 @@ fn code_to_text(code: &tonic::Code) -> &str {
...
     fn from_metadata(metadata: &[proto::v1::Metadata]) -> CliCompleter {
-        let mut properties = HashSet::<String>::new();
+        let mut root = PathPart::new();
         for entry in metadata {
-            properties.insert(entry.name.to_owned());
+            let mut parent = &mut root;
+            let mut parts = entry.name.split('.');
+            loop {
+                let path = parts.next();
+                match path {
```
flxo, line 610: I'm just curious - there's nothing wrong here, but isn't clippy complaining here that this could be a `while let`?

argerus: No clippy warning (wouldn't pass CI if there were), but you're absolutely right that using `while let` would be better.

PR #398, opened 2022-11-07
Add jemalloc for musl builds

argerus:
MUSL allocator is really slow. jemalloc is faster.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -176,6 +176,7 @@ dependencies = [
  "anyhow",
  "clap",
  "databroker-proto",
+ "jemallocator",
  "prost",
  "prost-types",
  "serde_json",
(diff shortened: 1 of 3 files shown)


SebastianSchildt:
Just a question: Do we currently have anything databroker side that is building/testing MUSL? I think our current databroker containers are glibc based.

Even if not, would be fine merging this, as it is conditional and should not influence the existing things.

Just wanted to double check, wether I need to test something for breakage :)

argerus:
> Do we currently have anything databroker side that is building/testing MUSL? I think our current databroker containers are glibc based.

Yes, this is correct. I just use the the musl build from time to time to impress myself with the total lack of dynamic dependencies :)

SebastianSchildt (approved):
Makes sense

lgtm 🍰

PR #399, opened 2022-11-08
kuksa-client: Add user-friendlier APIs to grpc.VSSClient with examples

romainletendart:


SebastianSchildt:
Will check tomorrow/later, but looking good. This will/would have made a lot of BCX hackers very happy 😀

SebastianSchildt (approved):
Works. Eventually we should have "real" documentation, but docstring helped :)
Looks much less cluttered for "common usecases"

lgtm 🦅

PR #400, opened 2022-11-08
BOM+licenses in databroker docker

SebastianSchildt:
This makes the databroker docker more compliant. When deploying an artifact (which a container is), the right thing(tm) to do is, including all licenses from your dependencies.

This PR achieves that for databroker, by first extracting a list of all dependencies used in the binary and then collecting all licenses and copying them over. 

Current design decisions
 - The processing/collecting of information is done by a small Python script (with the help of cargo), so we should be able to roll it out for other Rust components we might dockerize (looking at you databroker-cli)
 - It is executed during Docker build (and not as i.e. a separate GH action): I want to make sure it is not relying, on the correct "sequence" of things happening: When building the docker now you are guaranteed to get the correct BOM for that specific docker
 - The needed licenses texts (input for the Py script) are included in .gz form in the repo. I do hope it helps a little bit when other people are using "the usual tools" to scan our repo, that they do not flag having found a million licenses here
 - If the script encounters a previously unknown license (even if only as an additonal "OR" with a whitelisted one), it will abort. It is like a second safety net, so we do think about it when introducing dependencies with new licenses.
 - no complicated "updating of a previously known state": Script only runs when the output folder does not exist (dockerfile makes sure of that). I think it is better keeping it stateless, especially since the rust side allows us to programmatically get all required information


bs-jokri (approved):
LGTM

PR #402, opened 2022-11-10
Feature/stresstest

lukasmittag:
New stress test which uses the new python library from #387. Could be updated soon since #399 exists. 
The stress test can stress requesting to set values on KUKSA.val databroker and server. It provides the ability to stress test the subnscription to a topic too.


kuksa-client/kuksa_client/cli_backend/__init__.py
```diff
@@ -32,6 +32,7 @@ def __init__(self, config):
         self.cacertificate = config.get('cacertificate', str(self.default_cert_path / 'CA.pem'))
         self.certificate = config.get('certificate', str(self.default_cert_path / 'Client.pem'))
         self.keyfile = config.get('key', str(self.default_cert_path / 'Client.key'))
+        self.tokenfile = config.get('token', str(self.default_cert_path / 'jwt/all-read-write.json.token'))
```
SebastianSchildt, line 35: Why this is changing the client.
Also I think it is wrong to inject some "default" tokenfile.

@romainletendart  Any comments?

test/stresstest/broker.py
```diff
@@ -0,0 +1,18 @@
+import subprocess
```
SebastianSchildt, line 1: No license header

test/stresstest/client.py
```diff
@@ -0,0 +1,38 @@
+import kuksa_client
```
SebastianSchildt, line 1: Misses license header

test/stresstest/broker.py
```diff
@@ -0,0 +1,18 @@
...
+import config
+
+class Broker:
+
+    def start(self):
+        port = self.cfg.get("broker_port")
+        address = self.cfg.get("ip")
+        subprocess.run('cargo run --release --bin databroker -- --metadata vss_release_3.0.json --port %s --address %s' % (port, address), shell = True)
```
SebastianSchildt, line 9: Do we really want to have code in here to start databroker? Maybe we want. Generally speaking a Readme how all tis is intended to be used might help....

test/stresstest/config.py
```diff
@@ -0,0 +1,10 @@
+import configparser
```
SebastianSchildt, line 1: License header

test/stresstest/server.py
```diff
@@ -0,0 +1,26 @@
+import subprocess
```
SebastianSchildt, line 1: License header

test/stresstest/server.py
```diff
@@ -0,0 +1,26 @@
...
+import os.path
+import config
+
+class Server:
```
SebastianSchildt, line 5: Ah the equivalent to broker on top. I was wondering, might it also make sense - instead/in addtion to building the servers, to have way where they just pull (not build) the repsective containers and using them (then you could measure what is really released).

if they are started (assuming docker) with --net=host (=reachable by external testcleint without friuther forwarding) and --rm (do not leave old containers around) might be good

But this is not an "issue" with this PR, rather an idea :D

test/stresstest/stress.py
```diff
@@ -0,0 +1,23 @@
+import client
```
SebastianSchildt, line 1: Header

test/stresstest/stresstest.py
```diff
@@ -0,0 +1,35 @@
+import multiprocessing
```
SebastianSchildt, line 1: header

test/stresstest/subscriber.py
```diff
@@ -0,0 +1,18 @@
+import client
```
SebastianSchildt, line 1: header

test/stresstest/timer.py
```diff
@@ -0,0 +1,11 @@
+import time
```
SebastianSchildt, line 1: header

SebastianSchildt (changes requested):
See comments. Also, the vss_release file should not be replicated here. If it is needed, it can already be found in this repo in https://github.com/eclipse/kuksa.val/tree/master/data/vss-core

kuksa-client/kuksa_client/cli_backend/__init__.py
lukasmittag: I did this to be able to authorize the client to the server but yeah wasn't a good idea. Maybe there is a way to do this more generic.

test/stresstest/broker.py
lukasmittag: Is there a way to auto generate and add them? Or just manually copy them in?

test/stresstest/broker.py
lukasmittag: I did this for the stresstest, that it could be started fully automated. Maybe also by a git action :) later

test/stresstest/server.py
lukasmittag: This would be a good idea

kuksa-client/kuksa_client/cli_backend/__init__.py
erikbosch: We already set it for grpc right?, see https://github.com/boschglobal/kuksa.val/blob/feature/stresstest/kuksa-client/kuksa_client/cli_backend/grpc.py#L46 Seems similar to what I did in https://github.com/eclipse/kuksa.val/pull/371/files

romainletendart: Like Erik said, this is already done in the grpc backend.
If we need it on ws side as well, I think it makes sense to move grpc's code into the common class like you did Lukas.
Just remember to delete the code from the grpc implementation that does the same thing.
Regarding the righteousness of having a default token value, I think this needs to be addressed outside of this PR as we also provide a default private key which is also not secure.

test/stresstest/broker.py
erikbosch: I agree on that a small readme would be great, explaining the relationships between the *.py-files in this folder and how to start the tests? Just call `run_stresstest.sh`? 

Would also be good to write a line on the included *.json file. Is it just a copy of the already existing [VSS](https://github.com/boschglobal/kuksa.val/blob/feature/stresstest/data/vss-core/vss_release_3.0.json) file? If so, why is it duplicated. If just for convenience, maybe mention that it normally shall correspond to the latest version in the vss-core directory.

lukasmittag: The VSS file is just a copy. Was for convenience on my side - will change it. README's are always great :)

test/stresstest/client.py
```diff
@@ -0,0 +1,38 @@
...
+        if hasattr(self, "commThread"):
+            if self.commThread != None:
+                self.commThread.stop()
+                self.commThread = None
+        self.commThread = kuksa_client.KuksaClientThread(self.cfg)       #make new thread
+        self.commThread.start()                             #start thread
+
+    def __init__(self):
+        self.cfg = config.getConfig()
+        form = self.cfg.get("protocol")
```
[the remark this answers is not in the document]
lukasmittag: because we can not send a authorize request to the databroker but for server it is necessary or am I wrong?

test/stresstest/server.py
```diff
@@ -0,0 +1,26 @@
...
+
+    def build(self):
+        subprocess.run("mkdir ../../kuksa-val-server/build", shell = True)
+        subprocess.run('cd ../../kuksa-val-server/build && cmake .. && make', shell = True)
```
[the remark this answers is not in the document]
lukasmittag: Yeah probably best :)

test/stresstest/subscriber.py
```diff
@@ -0,0 +1,18 @@
...
+
+def msg_callback(msg):
+    # print("processing")
+    # print(msg)
+    # process msg
```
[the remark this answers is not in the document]
lukasmittag: If you want to see something on subscriber side you need this. I would like to keep them.

test/stresstest/stresstest.py
```diff
@@ -0,0 +1,35 @@
...
+    cfg = config.getConfig()
+    method = int(cfg.get("method"))
+    if method == 0:
+        _server = multiprocessing.Process(target=server.Server)
+        _server.start()
+    elif method == 1:
+        _broker = multiprocessing.Process(target=broker.Broker)
+        _broker.start()
+    else:
+        print("specify 0 or 1 | 0: server 1: broker")
```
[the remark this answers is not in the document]
lukasmittag: That's true, but I think the clients are started and when the databroker and server appear connect to it. I see here no problem. But generally yes would need some synchronization.

test/stresstest/timer.py
```diff
@@ -0,0 +1,11 @@
...
+
+class Timer():
+    def tic(self):
+        self.tstart_ns = time.time_ns()
+
+    def toc(self):
+        self.telapsed_ns = time.time_ns() - self.tstart_ns
+
+    def __init__(self):
+        self.tic()
```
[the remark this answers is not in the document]
lukasmittag: You mean for the timer? Just paste the code for tic and toc in the right places or what will be the solution from your side?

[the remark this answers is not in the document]
lukasmittag: Or just a function for tic() and a function for toc() without class Timer?

test/stresstest/config.ini
```diff
@@ -0,0 +1,11 @@
...
+
+ip=127.0.0.1
+port=55555
+insecure=true
+protocol=grpc
+broker_port=55555
+server_port=8090
+method=1
+stressSubscribe=true
+subNum=100
```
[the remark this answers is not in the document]
erikbosch: Should we possibly have a short comments for the configs, either here or in readme to give info on e.g. allowed values or corner cases - like is 0 a supported value for this config or not

`# Number of subscribers in stress test, 0 means no subscriptions will be used`

kuksa-client/kuksa_client/cli_backend/__init__.py
SebastianSchildt: Wanted to write a lot, instead put it in #406 

Wrt to this stresstest, it may be a seperate issue, but I do not see why this test can not just have a "clean" config file and give it to the library like a good citizen.
If the lib does not have authorize calls or other suitable hooks anymore, (@romainletendart  ?) maybe we should readd them, they could just do nothing in new GRPC currently

test/stresstest/subscriber.py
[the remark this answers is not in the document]
romainletendart: But they are commented out currently. If you want those prints to be optional maybe you could add a `verbose` option to your stress test or something?

test/stresstest/stresstest.py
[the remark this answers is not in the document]
romainletendart: If subscribers are started before the databroker, they just fail connecting and won't retry or wait until the databroker is there. I'd suggest you try running `cargo clean` and then `python stresstest.py` on your machine to see it happening.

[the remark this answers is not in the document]
lukasmittag: Ah okay, you're right. I will introduce some synchronisation

test/stresstest/subscriber.py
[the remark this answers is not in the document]
lukasmittag: Sounds good

test/stresstest/client.py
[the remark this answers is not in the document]
lukasmittag: I guess I got carried away by the german word Form :D

test/stresstest/server.py
lukasmittag: Should this be tackled in this PR or in a separate? I separated the stresstest script now from the scripts to run broker and server. I documented this in the README too. That you ahve to run the broker script and then run the stresstest script. In the run scripts for broker and server a flag could be set to distinguish between building from scratch and using the docker images.

test/stresstest/broker.py
```diff
@@ -0,0 +1,42 @@
...
+
+class Broker:
+    def handleSIGINT(self, signum, _):
+        if self.process != 0:
+            os.kill(self.process, signal.SIGTERM)
+    def start(self):
+        port = self.cfg.get("port")
+        address = self.cfg.get("ip")
+        kuksa_val_databroker_cmd = ['cargo', 'run', '--release', '--bin', 'databroker', '--', '--metadata', '../../data/vss-core/vss_release_3.0.json', '--port', f"{port}", '--address', f"{address}"]
+        self.process = subprocess.check_call(kuksa_val_databroker_cmd, stdin=None, stdout=None, stderr=None)
```
romainletendart, line 31: As databroker is expected to run indefinitely, `subprocess.check_call` will never return here.
If you start `python broker.py` and you then hit Ctrl-C,  `subprocess.check_call` will raise a `subprocess.CalledProcessError` which you don't catch here (which is fine as we **do** want to exit here anyway).
This means though `handleSIGINT` will never be called. As all that it does is terminating databroker's process anyway, I'd say we can get rid of custom signal handling here.

test/stresstest/broker.py
```diff
@@ -0,0 +1,42 @@
...
+    def __init__(self):
+        self.cfg = config.getConfig()
+        self.process = 0
+        signal.signal(signal.SIGINT, self.handleSIGINT)
```
romainletendart, line 35: See my other comment but we can get rid of signal handling here I'd say.

test/stresstest/broker.py
romainletendart: `check_call` already set `stdin=None, stdout=None, stderr=None` by default so I think you don't need to set them here.

test/stresstest/client.py
```diff
@@ -0,0 +1,48 @@
...
+
+    def start(self):
+        clientType = self.cfg.get("protocol")
+        if clientType == 'ws':
+            print('Starting VISS client')
+            try:
+                self.connect()
+                self.commThread.authorize(token=self.commThread.backend.tokenfile)
+            except:
+                raise Exception
```
romainletendart, line 32: What is the intent here?
What I read is: "Catch any exceptions with context and raise a new Exception without any context".
Couldn't you then let the original exception be raised without catching it?

test/stresstest/client.py
```diff
@@ -0,0 +1,48 @@
...
+        elif clientType == 'grpc':
+            print('Starting gRPC client')
+            try:
+                self.connect()
+            except:
+                raise Exception
+        else: 
+            raise ValueError
```
romainletendart, line 40:
```suggestion
            raise ValueError(f"Unknown client type {clientType}")
```
Context is always better when raising exceptions. Also, it's preferable to instantiate the exception type here i.e. `raise ValueError()` instead of `raise ValueError` (without `()`).

test/stresstest/config.py
```diff
@@ -0,0 +1,13 @@
...
+import os
+
+def getConfig():
+
+    config = configparser.ConfigParser()
+    home = os.path.expanduser("~")
+    config_path = home + '/kuksa.val/test/stresstest/config.ini'
```
romainletendart, line 8: I don't have that path on my machine.
I'd say it's better if you deduce that path from `config.py`'s path which you can get through `__file__`. Then you could do:
```python
config_path = pathlib.Path(__file__).resolve().parent
```
`resolve()` will resolve any symbolic links and turn your path into an absolute one.

test/stresstest/server.py
```diff
@@ -0,0 +1,58 @@
...
+import config
+import signal
+
+class Server:
+    def handleSIGINT(self, signum, _):
+        if self.process != 0:
+            os.kill(self.process, signal.SIGTERM)
+    def build(self):
+        home = os.path.expanduser("~")
+        build_path = home + '/kuksa.val/kuksa-val-server/build'
```
romainletendart, line 30: See my previous comment but we shouldn't use user-dependent paths here.

test/stresstest/server.py
```diff
@@ -0,0 +1,58 @@
...
+        print(os.getcwd())
+        src_path = build_path + '/src'
+        cmake_cmd = ["cmake",'..', build_path]
+        subprocess.check_call(cmake_cmd, stdin=None, stdout=None, stderr=None)
+        subprocess.check_call('make', cwd=src_path, stdin=None, stdout=None, stderr=None)
+    def start(self):
+        port = self.cfg.get('port')
+        address = self.cfg.get('ip')
+        home = os.path.expanduser("~")
+        build_path = home + '/kuksa.val/kuksa-val-server/build'
```
romainletendart, line 43: As you are defining/using the same `build_path` within `build()` and `start()`, you could define a `self.build_path` in `__init__()` and use the same one everywhere I think.

test/stresstest/stress.py
```diff
@@ -0,0 +1,47 @@
...
+import time
+import config
+
+def stressServer(num, path):
+    print('stress testing kuksa.val server with setValue')
+    Stressclient = client.Client()
+    Stressclient.start()
+    start = time.time_ns()
+    for count in range(0,num):
```
romainletendart, line 28:
```suggestion
    for count in range(0, num):
```

test/stresstest/stress.py
```diff
@@ -0,0 +1,47 @@
+########################################################################
+# Copyright (c) 2020 Robert Bosch GmbH
```
erikbosch, line 2: If we add new files, should not copyright year be current year? I.e. 2022?

test/stresstest/README.md
```diff
@@ -0,0 +1,11 @@
+The stresstest can be invoked by running: ''' python3 stresstest.py '''
+Before starting the stresstest run either: ''' python3 broker.py ''' or ''' python3 server.py '''
+For the configuration which kind of stresstest is executed , you can specify a config file like [config.ini](config.ini).
+In the configuration file you can specify:
+* port = port which gets used to connect the stress clients
+* ip = ip address which gets used to connect the stress clients
+* protocol = if the client should call websocket or gRPC's
```
erikbosch, line 7: Mention the actual string to use, maybe something like:

protocol = if the client should call websocket (`ws`, only supported by server)  or gRPC's (`grpc`)

test/stresstest/stress.py
lukasmittag: Of course :)

test/stresstest/server.py
lukasmittag: This will be tackled in a separate PR see #412

kuksa-client/kuksa_client/cli_backend/__init__.py
romainletendart: We still need to adapt https://github.com/eclipse/kuksa.val/blob/f1e1ce9f5f219a19048d150246bb6562cd969820/kuksa-client/kuksa_client/cli_backend/grpc.py#L45-L47 otherwise we declare twice `self.tokenfile` in grpc's case.
Please note that the one you introduced is a `str` and that grpc expects a `pathlib.Path`.

test/stresstest/config.ini
```diff
@@ -0,0 +1,11 @@
...
+
+    ip=127.0.0.1
+    port=55555
+    protocol=ws
+    insecure=False
+    subscriberCount=1
+    path=Vehicle.Speed
+    messageNum=1000
+    Debug=1
+    container=True
```
romainletendart, line 11:
```suggestion
[vss]
ip = 127.0.0.1
port = 55555
protocol = ws
insecure = False
subscriberCount = 1
path = Vehicle.Speed
messageNum = 1000
Debug = 1
container = True
```
`.ini` files usually:

- are unindented
- have no emtpy lines between a section title and its content but one empty line between sections (although you only have one section in this case)
- have spaces around equal signs.
For example: https://docs.python.org/3.8/library/configparser.html?highlight=configparser#quick-start

test/stresstest/README.md
```diff
@@ -0,0 +1,13 @@
+The stresstest can be invoked by running: ''' python3 stresstest.py '''
+Before starting the stresstest run either: ''' python3 broker.py ''' or ''' python3 server.py '''
```
romainletendart, line 2: I think you should switch lines 1 and 2 as starting a server is the first step to be done.

test/stresstest/README.md
```diff
@@ -0,0 +1,13 @@
...
+For the configuration which kind of stresstest is executed , you can specify a config file like [config.ini](config.ini).
```
romainletendart, line 3: How do I specify a config file?
When I read this I expect a CLI option that allows me to specify a config file.
Also, see my comment hereafter for config.py but it seems the config file is expected to be in the current working directory in order to work. That should be mentioned if it is expected here.

test/stresstest/README.md
```diff
@@ -0,0 +1,13 @@
...
+In the configuration file you can specify:
+* port = port which gets used to connect the stress clients
+* ip = ip address which gets used to connect the stress clients
+* protocol = if the client should call websocket (ws, only supported by server) or gRPC's (grpc)
+* insecure = Dis/Allows insecure connections to KUKSA.val server/databroker (False/True)
+* subscriberCount = amount of subscriber you want to spawn (0 is also allowed)
+* path = which VSS path gets used to stress
+* Debug = Display debug messages(1) else (0)
+
+The stresstest stresses setValue at the moment and builds everthing from scratch. Alternatively we should be able to pull docker images and run them and stresstest these instances. The second stresstest that could be performed is subscribing with a certain amount of subcribers. this checks how the bandwith and everything changes with 100 or more subscriber.
```
romainletendart, line 13: > 
```suggestion
The stresstest stresses setValue at the moment and builds everything from scratch. Alternatively we should be able to pull docker images and run them and stresstest these instances. The second stresstest that could be performed is subscribing with a certain amount of subscribers. this checks how the bandwidth and everything changes with 100 or more subscribers.
```

test/stresstest/config.py
```diff
@@ -0,0 +1,12 @@
+import configparser
+import pathlib
+
+def getConfig():
+
+    config = configparser.ConfigParser()
+    config_path = pathlib.Path("config.ini").resolve()
```
romainletendart, line 7: This means you expect "config.ini" to be in the current working directory. Is that what you intended?
Basically, I can't do `python3 test/stresstest/stresstest.py` otherwise I get:
```python
$ python test/stresstest/stresstest.py 
  File "/home/ler2so/Repo/kuksa.val/test/stresstest/stresstest.py", line 44, in <module>
    main()
  File "/home/ler2so/Repo/kuksa.val/test/stresstest/stresstest.py", line 38, in main
    cfg = config.getConfig()
  File "/home/ler2so/Repo/kuksa.val/test/stresstest/config.py", line 10, in getConfig
    vsscfg = config['vss']      #get data from config file
  File "/usr/lib/python3.10/configparser.py", line 964, in __getitem__
    raise KeyError(key)
KeyError: 'vss'
```

test/stresstest/server.py
```diff
@@ -0,0 +1,48 @@
...
+        isdir = os.path.isdir(self.build_path)
+        if not isdir:
+            subprocess.check_call(f"mkdir {self.build_path}")
+        cmake_cmd = ["cmake",'..', self.build_path]
+        subprocess.check_call(cmake_cmd)
+        subprocess.check_call('make', cwd=self.src_path)
+    def start(self):
+        port = self.cfg.get('port')
+        address = self.cfg.get('ip')
+        kuksa_val_server_cmd = ['./kuksa-val-server', '--log-level', 'ALL', '--address', f"{address}", '--port', f"{port}"]
```
romainletendart, line 35:
```suggestion
        kuksa_val_server_cmd = ['./kuksa-val-server', '--log-level', 'ALL', '--address', address, '--port', port]
```

test/stresstest/server.py
```diff
@@ -0,0 +1,48 @@
...
+import os.path
+import config
+import signal
+
+class Server:
+    def build(self):
+        isdir = os.path.isdir(self.build_path)
+        if not isdir:
+            subprocess.check_call(f"mkdir {self.build_path}")
+        cmake_cmd = ["cmake",'..', self.build_path]
```
romainletendart, line 29:
```suggestion
        cmake_cmd = ['cmake', '..', self.build_path]
```

test/stresstest/stress.py
```diff
@@ -0,0 +1,47 @@
...
+import client
+import time
+import config
+
+def stressServer(num, path):
+    print('stress testing kuksa.val server with setValue')
+    Stressclient = client.Client()
+    Stressclient.start()
+    start = time.time_ns()
+    for count in range(0,num):
```
romainletendart, line 28:
```suggestion
    for count in range(0, num):
```

test/stresstest/README.md
lukasmittag: Good point

kuksa-client/kuksa_client/cli_backend/__init__.py
romainletendart: The way you fixed the grpc module should do. 👍

test/stresstest/server.py
```diff
@@ -0,0 +1,54 @@
...
+import os.path
+import config
+import signal
+import argparse
+
+class Server:
+    def build(self):
+        isdir = os.path.isdir(self.build_path)
+        if not isdir:
+            subprocess.check_call(f"mkdir {self.build_path}")
```
romainletendart, line 29:
```suggestion
       os.makedirs(self.build_path,  exist_ok=True)
```
Does the same more or less without requiring a subprocess.

romainletendart (commented):
I only have minor comments left.
I would then rebase onto master, run kuksa_client's unit tests, run the stress test one last time and I think we are good for merging.

SebastianSchildt (approved):
👍

PR #403, opened 2022-11-11
Feature/databroker cli container

SebastianSchildt:
This PR includes a Dockerfile to build a container for databroker CLI.

 - It follows the exact same pattern as databroker: Simple debian-slim image.
 - It includes a BOM of thirdparty components + licenses

Not in scope of this PR
 - There might be more efficent ways going for MUSL builds and smaller/empty base image. Also cross compiling might integrated. (maybe for RUST we can find more efficient patterns than just "buildx"
 - No Action making use of it yet

To test, in kuksa.val root dir do

```
docker build -t cli -f kuksa_databroker/Dockerfile-cli .
```

and check it works 

```
docker run -it --rm  cli:latest  
```

Can you take a look @lukasmittag


kuksa_databroker/Dockerfile-cli
```diff
@@ -0,0 +1,47 @@
...
+# *
+# * This program and the accompanying materials are made available under the
+# * terms of the Apache License 2.0 which is available at
+# * http://www.apache.org/licenses/LICENSE-2.0
+# *
+# * SPDX-License-Identifier: Apache-2.0
+# ********************************************************************************/
+
+# This is expected to be executed in the kuksa.val top-level directory
+# This buuilds the databroker-cli
```
erikbosch, line 15: Minor spelling error "buuilds"

SebastianSchildt: Certainly true, and fixed now.

Would still be good to hear whether the container created by this Docker file actually  also _works_ :)

erikbosch:
I built locally and started it, but got an access denied when trying to set a value. It is not docker-related as I get the same error without docker. So the docker container seems to work.

(But we should better document in general for databroker-cli how you pass/set ip/port, as it is not obvious)

```
erik@debian3:~/kuksa.val$ docker run -it --net=host  --rm  cli:latest  
client> get Vehicle.Speed
-> Vehicle.Speed: ( NotAvailable )
client> set Vehicle.Speed 3
-> Error setting id Vehicle.Speed: AccessDenied
client> help
client commands:

  get             - Get a HAL property
  set             - Set a HAL property
  feed            - Set a HAL property (from vehicle side)
  connect         - Connect
  subscribe       - Subscribe to properties
  metadata        - Get datapoint metadata
  help            - You're looking at it
  quit            - Quit the demo

client> 
```

SebastianSchildt:
True, "passing port" is not even possible currently - it is hardcoded. I tended to see the databroker-cli rather as some test/hack tool for db development (because traditionally there was the kuksa cmdline client, but that had no support for db until recently). But it seems to be useful, so there are users, hence the container.
 
I could still see the python cmdline client being the "playing around" tool of choice later, but let's see how it works out in the long run.

Issue #391, opened 2022-11-03
Remove DAPR binding


SebastianSchildt:
The current DAPR binding is not used (and likely not even compatible with current DAPR). On the level of val we can not leverage much of the benefits

DAPR users can use Eclipse Velocitas DAPR bindings vor VSS data, as on the application it is more likely you can use it to interact with other components in the DAPR ecosystem.

SebastianSchildt:
Done in #413

PR #413, opened 2022-11-22
Remove DAPR support

SebastianSchildt:
This will solve #391 


erikbosch:
If we remove DAPR support here - is it still relevant to have some DAPR support in dbc2val feeder? (e.g. `VEHICLEDATABROKER_DAPR_APP_ID`). I.e. is the business reason for removing DAPR support here also valid for the feeders?

SebastianSchildt:
I do think it is not relevant in the feeder as well, as @romainletendart  is basing that on the new py lib (that has no DAPR dependencies), and my understanding is how you feed databroker has nothing to do with whether on top you use DAPR for other clients, but maybe @argerus  can confirm.

In any case, this PR should be fine, as it just removes DAPR support for kuksa.val server

erikbosch (approved):

erikbosch:
In dbc2val we have two DAPR-related configs. I can now understand the usecase for `VEHICLEDATABROKER_DAPR_APP_ID`, but I do not really see any reason for `DAPR_GRPC_PORT`, you could as well use the standard config mechanisms for address and port, right?

PR #423, opened 2022-11-29
Fix clippy warning (and compiler error)

argerus:
Rust 1.65 introduced some clippy warnings and a compiler error.

- Fix clippy warning.
- Use `while let Some(..)` instead of `loop` as suggested by @flxo
- Remove split-debuginfo


SebastianSchildt (approved):
Works

PR #427, opened 2022-11-30
kuksa-client: Document APIs with examples

romainletendart:


SebastianSchildt:
I see on grpc side you showcased the "simple" versions, should we also give a more lower level example, where you populate the DataEntry/fields/view by yourself?

I mean like this direction

```
            await client.set(
                updates=(EntryUpdate(entry=DataEntry(path='Vehicle.Speed', 
                value=Datapoint(value=200+count)), fields=(Field.VALUE,),
            ),)) 
```

SebastianSchildt (commented):
Nice. Just see comment above, and generally suggestion: I see you used all absolute links, so they will also work when the main readme is pushed to Pypi, but it is easier to maintain to just link to some general landing page ("See the docs for different (sync/async) ways to interact with kuksa.val". and then have small "landing .md" that branches out to the sync/async mds?

One indirection more, but might make navigation within one branch/tag easier

romainletendart:
> Nice. Just see comment above, and generally suggestion: I see you used all absolute links, so they will also work when the main readme is pushed to Pypi, but it is easier to maintain to just link to some general landing page ("See the docs for different (sync/async) ways to interact with kuksa.val". and then have small "landing .md" that branches out to the sync/async mds?
> 
> One indirection more, but might make navigation within one branch/tag easier

I've added the indirection you mentioned. It makes things easier but still not perfect as landing on README.md from a tag, will still point to main.md from the master branch. However once you reached main.md, you can select again the tag you're interested in and navigate from there.

romainletendart:
> Saw you sneaked in a bugfix/small improvement with protocol default value. Not sure if intended, but taking it :)

Could have been in a separate PR but I thought this made the threaded-API examples simpler so...

PR #432, opened 2022-12-07
kuksa-client: Simplify grpc subscribe*() and add synchronous implementations

romainletendart:
`grpc.aio.VSSClient.subscribe*()` now return async iterators and no longer accept callbacks.
This allows easier subscriptions as one can now directly iterate over asynchronous updates using an `async for` loop.
This also removes the need for having a function (callback) call for every update we get on our subscription.
This also allows a similar API to exist for the synchronous API i.e. `grpc.VSSClient.subscribe*()`.


SebastianSchildt (approved):
Looking good 🚗

PR #436, opened 2022-12-09
Add SIGTERM handler to databroker

### LINKED ISSUE #433: Databroker container not listening to interrupt signals
The databroker container (master as of 2022-12-09) does not seem to correctly handle system interrupt signals, such as SIGTERM, SIGINT.

Observed behavior:
- When shutting down the container runtime or stopping the container, the databroker container keeps running until the container runtime does a timeout and forcefully kills the container.
- Esp. for shutdown of the system, it is observable that eg systemd is waiting long time for Kanto-CM to shut down, as it waits for the databroker container to exit.

Expected behavior:
- databroker container listens for signals (SIGINT and SIGTERM) and gracefully shuts down (eg closing open channels)

How to test:
- Create a container with kuksa.val databroker: `docker run ghcr.io/eclipse/kuksa.val/databroker:master`
- Stop it with `docker stop --time 60 <container>`
- It should stop much faster than the 60 seconds

Notes:
There is a timeout setting for container runtime. So if a container is not properly handling signals, it will forcefully kill the process at some time. This is system dependent configuration and only meant as a fallback. Containers should behave nicely to also ensure a quick shutdown.

SebastianSchildt:
Might fix #433

### DIFF
--- a/kuksa_databroker/databroker/src/main.rs
+++ b/kuksa_databroker/databroker/src/main.rs
@@ -127,10 +127,13 @@ async fn shutdown_handler() {
     let mut sigint =
         signal(SignalKind::interrupt()).expect("failed to setup SIGINT signal handler");
     let mut sighup = signal(SignalKind::hangup()).expect("failed to setup SIGHUP signal handler");
+    let mut sigterm =
+        signal(SignalKind::terminate()).expect("failed to setup SIGTERM signal handler");
 
     select! {
         _ = sigint.recv() => info!("received SIGINT"),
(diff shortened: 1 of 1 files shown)


argerus (approved):

argerus:
Approved. Would have been fine to replace the occurance of SIGHUP with SIGTERM (i.e. init shutdown on SIGTERM/SIGINT instead of SIGHUP/SIGINT), but this works just as well.

If we wanna do stuff like reload configuration on SIGHUP in the future, we can just change it.

PR #438, opened 2022-12-13
kuksa-client: Make Ctrl-D stop kuksa-client

romainletendart:
Ctrl-D closes stdin which in turn makes cmdloop() return. do_quit() is not called in this case.
This meant that the client thread would still be lying around and the main thread would then wait forevever for its termination. Move the exit logic outside of do_quit() to make sure we stop the client thread no matter how we exited the command loop.


argerus (approved):

argerus:
Works for me

SebastianSchildt (approved):
👍

PR #439, opened 2022-12-14
kuksa-client: Introduce requirements pinning using pip-tools

romainletendart:


SebastianSchildt (approved):
Looks reasonable & works for me

PR #441, opened 2022-12-16
Container for rust CLI

SebastianSchildt:
This is fixing the problem that we do not have a container for the rust cli.

It has some bonus fixes, that we might roll out to other actions components

 - The CLI is shipped in a `FROM scratch` container. Small. 
 - There is some magic in there, that does extract the binaries from the container and provides them as GH Action artifacts (including bom. That may not technically be neccessary but is how things "should be done"(tm) if we were to include it in a release.  
 - In the GH Action summary page it will print the location of the temporary docker for testing (easier than scrolling through hundreds lines of code)

### DIFF
--- a/.github/workflows/kuksa_databroker-cli_build.yml
+++ b/.github/workflows/kuksa_databroker-cli_build.yml
@@ -1,0 +1,157 @@
+# /********************************************************************************
+# * Copyright (c) 2022 Contributors to the Eclipse Foundation
+# *
+# * See the NOTICE file(s) distributed with this work for additional
+# * information regarding copyright ownership.
+# *
+# * This program and the accompanying materials are made available under the
+# * terms of the Apache License 2.0 which is available at
+# * http://www.apache.org/licenses/LICENSE-2.0
+# *
+# * SPDX-License-Identifier: Apache-2.0
+# ********************************************************************************/
+
+name: kuksa_databroker-cli_build
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    paths:
+      - ".github/workflows/kuksa_databroker-cli_build.yml"
+      - "kuksa_databroker/**"
+      - "Cargo.*"
+      - "Cross.toml"
+  workflow_call:
+  workflow_dispatch:
+
+
+jobs:
+  lint:
+    name: Lint
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v3
+      - name: cargo fmt
+        working-directory: ${{github.workspace}}
+        run: cargo fmt -- --check
+      - name: cargo clippy
+        working-directory: ${{github.workspace}}
+        run: cargo clippy --all-targets -- -W warnings -D warnings
+
+
+  checkrights:
+    uses: ./.github/workflows/check_push_rights.yml
+    secrets: inherit
+
+# Run on selfhosted, because our runner has native ARM build in a remote
+# builder (no need for qemu)
+  build-container:
+    runs-on: [ self-hosted ]
+    needs: checkrights
+
+    steps:
+    - uses: actions/checkout@v3
+      with: 
+        submodules: recursive
+
+    - name: Docker meta
+      id: meta
+      uses: docker/metadata-action@v4
+      with:
+        # list of Docker images to use as base name for tags
+        images: |
+          ghcr.io/eclipse/kuksa.val/databroker-cli
+        # generate Docker tags based on the following events/attributes
+        tags: |
+          type=ref,event=branch
+          type=ref,event=pr
+          type=semver,pattern={{version}}
+          type=semver,pattern={{major}}.{{minor}}
+          type=semver,pattern={{major}}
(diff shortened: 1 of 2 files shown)


SebastianSchildt:
> * In the GH Action summary page it will print the location of the temporary docker for testing (easier than scrolling through hundreds lines of code)

https://github.com/eclipse/kuksa.val/actions/runs/3713953054

argerus (changes requested):
Rename all instances of `databrokercli` -> `databroker-cli` to have consistent naming.

Existing ghcr packages are named, e.g. `ghcr.io/eclipse/kuksa.val/kuksa-client` and not `ghcr.io/eclipse/kuksa.val/kuksaclient`.

The same goes for naming the binaries and github actions.

.github/workflows/kuksa_databrokercli_build.yml
```diff
@@ -0,0 +1,154 @@
...
+        echo -e "\nImages for testing temporarily available at \`ttl.sh/kuksa.val/kuksa-databrokercli-${{github.sha}}:1h\`" >> $GITHUB_STEP_SUMMARY
+        echo -e "\n\`\`\`\ndocker pull ttl.sh/kuksa.val/kuksa-databrokercli-${{github.sha}}:1h\n\`\`\`" >> $GITHUB_STEP_SUMMARY
+    
+    - name: Extracting AMD64 binaries
+      run: | 
+         docker rm cliamd || true
+         docker create --name cliamd --platform linux/amd64 ttl.sh/kuksa.val/kuksa-databrokercli-${{github.sha}}:1h
+         docker cp cliamd:/app ./  
+         docker rm cliamd
+         mv ./app ./databrokercli-amd64
```
argerus, line 133:
```suggestion
         mv ./app ./databrokercli-amd64
         tar cf databroker-cli-amd64.tar.gz -C databroker-cli-amd64 .
```

Explicitly tar the files instead, since file mode are lost if relying on github for zipping the directory (i.e. in upload-artifact below).

.github/workflows/kuksa_databrokercli_build.yml
```diff
@@ -0,0 +1,154 @@
...
+         docker create --name cliarm --platform linux/arm64 ttl.sh/kuksa.val/kuksa-databrokercli-${{github.sha}}:1h
+         docker cp cliarm:/app ./  
+         docker rm cliarm
+         mv ./app ./databrokercli-arm64
+         
+    - name: Archive  AMD64 binaries
+      uses: actions/upload-artifact@v3
+      with:
+        name: databrokercli-amd64
+        path: ./databrokercli-amd64
```
argerus, line 147:
```suggestion
        path: ./databroker-cli-amd64.tar.gz
```
Use created tar instead. File mode (executable) is lost when using upload-artifact for zipping the directory.

.github/workflows/kuksa_databrokercli_build.yml
```diff
@@ -0,0 +1,154 @@
...
+      uses: actions/upload-artifact@v3
+      with:
+        name: databrokercli-amd64
+        path: ./databrokercli-amd64 
+         
+    - name: Archive  ARM64 binaries
+      uses: actions/upload-artifact@v3
+      with:
+        name: databrokercli-arm64
+        path: ./databrokercli-arm64
```
argerus, line 153:
```suggestion
        path: ./databroker-cli-arm64.tar.gz
```

argerus (changes requested):

.github/workflows/kuksa_databrokercli_build.yml
```diff
@@ -0,0 +1,154 @@
...
+      - name: cargo clippy
+        working-directory: ${{github.workspace}}
+        run: cargo clippy --all-targets -- -W warnings -D warnings
+
+
+  checkrights:
+    uses: ./.github/workflows/check_push_rights.yml
+    secrets: inherit
+
+# Run on selfhosted, becasue our runner has native ARM build in a remote
```
erikbosch, line 48: Spelling error "becasue"

.github/workflows/kuksa_databrokercli_build.yml
```diff
@@ -0,0 +1,154 @@
...
+
+    - name: Log in to the Container registry
+      if: needs.checkrights.outputs.have_secrets == 'true'
+      uses: docker/login-action@v2
+      with:
+          registry: ${{ needs.checkrights.outputs.registry }}
+          username: ${{ github.repository_owner }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+    - name: Build kuksa.val databroker CLI container and push to ghcr.io
```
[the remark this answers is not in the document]
SebastianSchildt: Yes, I found it doesn't hurt, and makes it easier later when you know exactly where to extract the binaries without further logic. Will adapt the text

SebastianSchildt:
Everything should be modified according to your suggestions

argerus (approved):

Issue #440, opened 2022-12-16
Fix deprecation in check_push_rights


SebastianSchildt:
Currenly CI runs generate a warning:

```
[checkrights / check_push_rights](https://github.com/boschglobal/kuksa.val/actions/runs/3713425053/jobs/6296053935#step:2:14)
The `set-output` command is deprecated and will be disabled soon. Please upgrade to using Environment Files. For more information see: https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/
```

Should be fixed

SebastianSchildt:
done in #445 and #443

PR #445, opened 2023-01-07
check_push rights depreaction fix 

SebastianSchildt:
Use non-deprecated way to set outputs, removed unused output


SebastianSchildt:
will fix #440

erikbosch (approved):

PR #448, opened 2023-01-13
Feature/databroker cli version

SebastianSchildt:


kuksa_databroker/Dockerfile-cli
```diff
@@ -15,27 +15,59 @@
...
+# it seems on AMD build some dependency requires zlib headers
+# also the grpc proto compilation always falls back to
+# some binaries downloaded by cargo that require glibc
+#RUN apk add zlib-dev libc6-compat  libstdc++ gcompat
+
+# We need to specifiy target in build later to enable
+# static linking
+# See https://msfjarvis.dev/posts/building-static-rust-binaries-for-linux/
+# and https://github.com/rust-lang/rust/issues/78210
+ENV BUILDTARGET=x86_64-unknown-linux-musl
```
argerus, line 33: I didn't notice that you didn't specify `--target x86_64-unknown-linux-musl` when building before (i.e. in the databroker Docker file for example).

Wouldn't it make more sense to put all $TARGETARCH specific stuff in their respective `FROM base AS builder-$TARGETARCH`

```suggestion
ENV RUSTFLAGS='-C link-arg=-s'

FROM base AS builder-amd64
RUN echo "AMD64 quirks"
# it seems on AMD build some dependency requires zlib headers
# also the grpc proto compilation always falls back to
# some binaries downloaded by cargo that require glibc
#RUN apk add zlib-dev libc6-compat  libstdc++ gcompat
RUN cargo build  --bin databroker-cli --release --target x86_64-unknown-linux-musl
RUN mv /build/target/x86_64-unknown-linux-musl/release/databroker-cli /dist/
```

and in the final step
```
FROM  scratch

COPY --from=builder /dist/databroker-cli /app/databroker-cli
```


Also not sure why it works at all without adding:
`rustup target add aarch64-unknown-linux-musl`

when cross-compiling..

.github/workflows/kuksa_databroker-cli_build.yml
```diff
@@ -53,8 +53,19 @@ jobs:
...
-        submodules: recursive
+
+    # We need tags for versioning,
+    # Checkout action above would only get them 
+    # using fetch-depth=0 (fetch all)
+    # just fetching tags alone is not sufficient as we need to
+    # have their history for them to be visible by git describe
+    # As a compromise we only deepen the shallow checkout.
+    # See also  https://github.com/actions/checkout/issues/701
+    - name: Making GIT tags available
```
argerus, line 64: Since we are going to use the same version for all applications in the repo, I think it would make more sense to make the version globally available as an environment variable. That could be a simple github action running the equivalent of `git describe --tags --dirty=-dirty`.

Otherwise, we have to (re-)implement these extractions for C++ / python / Rust, and I'm not sure there's much benefit to doing that.

erikbosch:
@SebastianSchildt @argerus  - this PR is old. Is it still relevant, and if so should we better include it in KUKSA 0.4.0?

SebastianSchildt:
It is outdated, closing  it

PR #450, opened 2023-01-13
Support reading metadata from more than one file

sophokles73:
The "--metadata" option now accepts a comma-separated list of file paths. This is helpful when supporting custom datapoints in addition to the standard ones.

Also made sure that the port parameter value is in the valid port range.

### DIFF
--- a/kuksa_databroker/databroker/src/main.rs
+++ b/kuksa_databroker/databroker/src/main.rs
@@ -1,5 +1,5 @@
 /********************************************************************************
-* Copyright (c) 2022 Contributors to the Eclipse Foundation
+* Copyright (c) 2022-2023 Contributors to the Eclipse Foundation
 *
 * See the NOTICE file(s) distributed with this work for additional
 * information regarding copyright ownership.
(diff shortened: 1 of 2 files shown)


SebastianSchildt:
Nice, I like the feature, especially since it is one of those things the older val-server can do, but databroker not yet.

wrt to the actual implementation @argerus  will review it and might come back with some tipps.

I see the Linter is unhappy, in general, in case you are using VS Code `Format Document` in the command palette should fix that

kuksa_databroker/databroker/src/main.rs
```diff
@@ -183,16 +236,21 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
...
         .arg(
             Arg::new("metadata")
                 .display_order(4)
                 .long("metadata")
-                .help("Populate data broker with metadata from file")
+                .help("Populate data broker with metadata from (comma-separated) list of files")
                 .takes_value(true)
+                .use_value_delimiter(true)
+                .require_value_delimiter(true)
+                .multiple_values(true)
```
argerus, line 250: I think it would be beneficial to be able to specify `--metadata FILE` multiple times (in addition to a comma-separated list of files).
```suggestion
                .multiple_values(true)
                .action(clap::ArgAction::Append)
```

argerus (approved):

PR #452, opened 2023-01-16
Add Quickstart documentation

SebastianSchildt:
I think we are missing some quickstart instructions. This ones tires that, leaving lots of details out, but using the hopefully most important features in a way that will run on 95% of all systems,

@erikbosch can you take look (maybe add commits to the branch directly if you want to change something)

### DIFF
--- a/doc/quickstart.md
+++ b/doc/quickstart.md
@@ -1,0 +1,173 @@
+# KUKSA.val Quickstart
+
+The quickest possible way to get KUKSA.val up and running
+
+## Starting broker
+First we want to run KUKSA.val databroker
+
+```
+docker run -it --rm --net=host ghcr.io/eclipse/kuksa.val/databroker:master 
+```
+
+
+## Reading and Writing VSS data via CLI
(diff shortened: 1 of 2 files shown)


erikbosch (approved):
Looks good to me

doc/quickstart.md
```diff
@@ -0,0 +1,175 @@
...
+from kuksa_client.grpc import Datapoint
+
+import time
+
+with VSSClient('127.0.0.1', 55555) as client:
+    for speed in range(0,100):
+        client.set_current_values({
+        'Vehicle.Speed': Datapoint(speed),
+        })
+        print(f"Setting Vehicle.Speed to {speed}")
```
lukasmittag, line 88:
```suggestion
        print(f"Feeding Vehicle.Speed to {speed}")
```
Maybe underline the meaning of Feed and Set. Only problem if someone just copies the code and uses not a sensor path. So I leave it to you :) Instead looks good to me.

lukasmittag (approved):

PR #456, opened 2023-01-20
Update VSS parser

argerus:
Add support for parsing `min`, `max` & `allowed`.

### DIFF
--- a/kuksa_databroker/databroker/src/vss.rs
+++ b/kuksa_databroker/databroker/src/vss.rs
@@ -11,29 +11,131 @@
 * SPDX-License-Identifier: Apache-2.0
 ********************************************************************************/
 
+use std::collections::{BTreeMap, HashMap};
 use std::fmt;
+
+use serde::Deserialize;
 
 use crate::types;
 
-#[derive(Debug)]
-pub struct VssEntry {
-    pub name: String,
+#[derive(Debug, Deserialize)]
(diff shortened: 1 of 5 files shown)


kuksa_databroker/databroker/src/vss.rs
```diff
@@ -49,294 +151,325 @@ impl fmt::Display for Error {
...
+                .map_err(|err| err.into()),
+            types::DataType::Int16Array => serde_json::from_value::<Vec<i16>>(value)
+                .map(|array| {
+                    Some(types::DataValue::Int32Array(
+                        array.iter().map(|value| (*value).into()).collect(),
+                    ))
+                })
+                .map_err(|err| err.into()),
+            types::DataType::Int32Array => serde_json::from_value::<Vec<i32>>(value)
+                .map(|value| Some(types::DataValue::Int32Array(value)))
```
erikbosch, line 311: How come it says `| value|` here and on line 314? Error or something i do not understand?

argerus: That was pretty eagle-eyed of you :)

It's not an error, since it's only describes what to name the input variable in the closure for the `map` function.

It _is_ inconsistent, though, so I'll change it.

SebastianSchildt (approved):
"Works", or at least at this stage rather  does not see to break things to me

PR #467, opened 2023-01-25
Update Cargo files

lukasmittag:
Update the Cargo.toml's and update Cargo.lock file


SebastianSchildt (approved):
Works. Let's see it also makes dependabot a little happier

Issue #460, opened 2023-01-21
jemalloc maybe incompatible with some aarch64 CPUs


SebastianSchildt:
I can't run our ARM builds on a Cortex-A57

```
[/share/Download] # ./databroker/databroker 
<jemalloc>: Unsupported system page size
<jemalloc>: Unsupported system page size
memory allocation of 5 bytes failed
Aborted
[/share/Download] # cat /proc/cpuinfo 
processor	: 0
model name	: Annapurna Labs Alpine AL324 Quad-core ARM Cortex-A57 CPU @ 1.70GHz
Features	: fp asimd evtstrm aes pmull sha1 sha2 crc32
CPU implementer	: 0x41
CPU architecture: 8
CPU variant	: 0x1
CPU part	: 0xd07
CPU revision	: 3

processor	: 1
model name	: Annapurna Labs Alpine AL324 Quad-core ARM Cortex-A57 CPU @ 1.70GHz
Features	: fp asimd evtstrm aes pmull sha1 sha2 crc32
CPU implementer	: 0x41
CPU architecture: 8
CPU variant	: 0x1
CPU part	: 0xd07
CPU revision	: 3

processor	: 2
model name	: Annapurna Labs Alpine AL324 Quad-core ARM Cortex-A57 CPU @ 1.70GHz
Features	: fp asimd evtstrm aes pmull sha1 sha2 crc32
CPU implementer	: 0x41
CPU architecture: 8
CPU variant	: 0x1
CPU part	: 0xd07
CPU revision	: 3

processor	: 3
model name	: Annapurna Labs Alpine AL324 Quad-core ARM Cortex-A57 CPU @ 1.70GHz
Features	: fp asimd evtstrm aes pmull sha1 sha2 crc32
CPU implementer	: 0x41
CPU architecture: 8
CPU variant	: 0x1
CPU part	: 0xd07
CPU revision	: 3

```

SebastianSchildt:
Found this
https://github.com/jemalloc/jemalloc/issues/467
Need to investigate, however if there is a hard dependency on the build system, maybe making jemalloc optional is better.

I think we should not start shipping builds for different pagesizes, as most users would likely not even know what there's is

SebastianSchildt:
Fixed in #468

PR #468, opened 2023-01-25
Making jemalloc an optional feature

SebastianSchildt:
Should solve #460

### DIFF
--- a/kuksa_databroker/databroker/Cargo.toml
+++ b/kuksa_databroker/databroker/Cargo.toml
@@ -48,8 +48,12 @@ sqlparser = "0.16.0"
 serde = { version = "1.0", features = ["derive"] }
 serde_json = "1.0"
 
-[target.'cfg(all(target_env = "musl", target_pointer_width = "64"))'.dependencies.jemallocator]
-version = "0.5.0"
+jemallocator = { version =  "0.5.0", optional = true }
+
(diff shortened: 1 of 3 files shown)


argerus (changes requested):
Looks good, .
But can I suggest that we call the feature `jemalloc` instead.

kuksa_databroker/databroker/Cargo.toml
```diff
@@ -48,8 +48,8 @@ sqlparser = "0.16.0"
 serde = { version = "1.0", features = ["derive"] }
 serde_json = "1.0"
 
-[target.'cfg(all(target_env = "musl", target_pointer_width = "64"))'.dependencies.jemallocator]
-version = "0.5.0"
+# to enable jemmalloc use --features jemallocator
+jemallocator = { version =  "0.5.0", optional = true }
```
argerus, line 53:
```suggestion
[features]
jemalloc = ["dep:jemallocator"]

```

argerus (changes requested):

SebastianSchildt:
Fixed.
Probably not directly related, but my cargo was thinking serde_derive was missing from Cargo.lock, so I also added it, so at least version is pinned.

argerus (approved):

PR #469, opened 2023-01-26
Add dependencies as recommended by Eclipse process

SebastianSchildt:
Dependecies as created by dash for released artifacts


bs-jokri (approved):
@SebastianSchildt great to have that dependencies file now in the project. Hwe should think about putting the generation of the file into the build process. And we should move the file to the root folder of the repo as soon as the planned restructuring has happened.

PR #471, opened 2023-01-30
Authorization in KUKSA.VAL

argerus:
One design goal for the authorization in KUKSA.VAL is that it should be as independent as possible from the authorization infrastructure used around it. As a result, the preferred solution has been to use a signed JWT access token where everything needed for authorization is contained within. In kuksa-val-server, this information was contained in the claim `"kuksa-vss"` along with the separate `"admin"` and `"modifyTree"` claims.

```yaml
{
  ...
  "admin": true,
  "modifyTree": false,
  "kuksa-vss":  {
     "Vehicle.Drivetrain.Transmission.DriveType": "rw",
     "Vehicle.OBD.*": "r"
  }
}
 ```

One downside of this was that these authorization claims where not readily compatible with the (now) widely used [OAuth 2.0 Authorization Framework](#the-oauth-20-authorization-framework) and the accompanying [JWT Profile for OAuth 2.0 Access Tokens](#json-web-token-jwt-profile-for-oauth-20-access-tokens). Using these, the same functionality is instead provided by adding a scope to the request for access tokens from the authorization server which is then included in the issued JWT access token as the `"scope"` claim. Being compatible with this standard would make KUKSA.VAL compatible with a whole range of external authorization providers and libraries.

Another thing missing in the original design, was the ability to differentiate between the right to actuate signals (i.e. a form of write access) and the right to change the value of the signals (another form of write access).

This PR consists of a document ([Authorization in KUKSA.VAL](https://github.com/boschglobal/kuksa.val/blob/topic/authorization/doc/KUKSA.val_data_broker/authorization.md)) aiming to describe a new design for authorization which would address these shortcomings.

### CODE REVIEW

**Comment by SebastianSchildt** [comment]
"Readable" https://github.com/boschglobal/kuksa.val/blob/topic/authorization/doc/KUKSA.val_data_broker/authorization.md

**Comment by argerus** [comment]
Different takes at designs available as separate files now:
[Alternative A](https://github.com/boschglobal/kuksa.val/blob/topic/authorization/doc/KUKSA.val_data_broker/authorization_alt_a.md)
[Alternative B](https://github.com/boschglobal/kuksa.val/blob/topic/authorization/doc/KUKSA.val_data_broker/authorization_alt_b.md)
[Alternative C](https://github.com/boschglobal/kuksa.val/blob/topic/authorization/doc/KUKSA.val_data_broker/authorization_alt_c.md)
[Alternative D](https://github.com/boschglobal/kuksa.val/blob/topic/authorization/doc/KUKSA.val_data_broker/authorization_alt_d.md)

File: doc/KUKSA.val_data_broker/authorization_alt_d.md [hunk 1]
```diff
@@ -0,0 +1,315 @@
...
+issuing access tokens more restrictive than the authorization grant
+used to obtain them, as well as removing the resource server's need
+to understand a wide range of authentication methods.
+   
+
+
+#### Roles [[RFC 6749](https://www.rfc-editor.org/rfc/rfc6749#section-1.1)]
+
+OAuth defines four roles:
+
```
SebastianSchildt: What I like most about adopting OAuth wording, is that we can from an security architecture point of vielw clearly define the scope (pun intended :) ) and responsibilities of KUKSA.val. So to line out my current understanding

 * **resource owner**
 That depends on the use case, it may be the driver, the owner of a vehicle the OEM, etc. On a technical level we wouldn't know

* **resource server**
That is us. KUKSA.val databroker

* **client**
 Everything interacting with our APIs

* **authorization server**
The provider of tokens. Dependent on your application/deployment. It is a "technical component", that is outside the scope of KUKSA.val. Maybe some backend server that is triggered on the fly, if the user gives consent is some UI, maybe some packaging magic, bundling and signing application bundles/updates, etc.

doc/KUKSA.val_data_broker/authorization_alt_d.md
```diff
@@ -0,0 +1,315 @@
...
+This token contains everything the server needs to verify the authenticity of the claims
+contained therein. The structure is specified in the RFC as:
+
+#### JWT Access Token - Data Structure
+
+| Claim     | Description                                          |
+|-----------|------------------------------------------------------|
+| iss       | REQUIRED - as defined in Section 4.1.1 of [RFC7519]. |
+| exp       | REQUIRED - as defined in Section 4.1.4 of [RFC7519]. |
+| aud       | REQUIRED - as defined in Section 4.1.3 of [RFC7519]. See Section 3 for indications on how an authorization server should determine the value of "aud" depending on the request. |
```
SebastianSchildt, line 79: Maybe need to clarify whether this is about _our_ usage, or a recap of JWT in general. For example, while we shoud requrie defined "aud"  claims for databroker, if i read correctly in RFC7519 it is optional

doc/KUKSA.val_data_broker/authorization_alt_d.md
```diff
@@ -0,0 +1,358 @@
...
+
+#### Format
+
+The scopes used in KUKSA.VAL are "prefix scopes" that consist of a VSS path, prefixed
+with an action.
+
+| Scope                 | Description                 |
+|-----------------------|-----------------------------|
+|`<ACTION>:<PATH>`      | Allow ACTION for PATH       |
+|`!<ACTION>:<PATH>`     | Deny ACTION for PATH        |
```
SebastianSchildt, line 113: I still have a strong dislike for the negative scopes/"deny rules", I feel even more strongly so when using the scope

 - It just feels unnatural to me. I did try to find any wording in any  RFC that this is forbidden, but could not. None of the posted examples did this though
   - The Linus Torvalds argument here is maybe "this is so stupid, that nobody before us ever though about it"
 - It adds complication, i.e. you need add special handling and explanantion, such as the precedence below, that makes it less obvious for users when having situations like.  read:Vehicle.Body.Trunk !read:Vehicle:Body
 - If we ever extend/add scopes that include others, i.e. read can read "all", while there a specific scopes for certain (meta)data such as current values/QoS information , etc. it gets more complex
 - I think it is really not needed. The only _potential_(!) problem it solves are to large scope claims, because you want to include too many scattered datapoints that a negative list is smaller than a positive. We might (for a long time)never hit this. If we do there might be any number of other solutions for that (i.e. the tagging discussed below may be one such thing)
 - Leaving it out now, would not preclude a path of adding it later, as it is just another scope. But I really think we shouldn't.

doc/KUKSA.val_data_broker/authorization_alt_d.md
```diff
@@ -0,0 +1,358 @@
...
+
+All "deny" scopes have priority over any "allow" scope.
+
+Examples of KUKSA.VAL scope strings:
+
+| Scope string             | Access                                        |
+|--------------------------|-----------------------------------------------|
+|`read:Vehicle.Speed`      | Client can only read `Vehicle.Speed`          |
+|`read:Vehicle.ADAS`       | Client can read everything under Vehicle.ADAS |
+|`actuate:Vehicle.ADAS`    | Client can actuate all actuators under Vehicle.ADAS |
```
SebastianSchildt, line 123: I think this is interesting, what is a good initial set of scope-Actions-

`read` surely is. We might in the future want to add some more granular read rights as discussed below, but probably some for of general read permission is good

`actuate` Seems ok, but I am a little on the fence here:  So my understanding is we want to express, that we can write target value. the intent is to actuate,  but that does not mean "actuation" really happens, because there is a stack of software below that will be the judge of that. In that case write:target_value would be more exact

`provide`I do not see as a good term. Our terminology doc still pending,  let me just base this on "unwritten truths". So I assume we do use the term "Provider" for what has been a called a feeder so far. I would use the term provider for everything that connects to a "real" car, so also simple VSS services, i.e. a module that listens for some target value for opening a door, and then doing it. (for that he would not need a provide claim, just a read, so maybe if we don't want something along the lines of "write-current", this shall rather be feed.


 another way to go about this, would be, to take a look at the "Using "access_mode" + "fields" for more granularity" section, and define that first, and treat the comon things as shortcut, just examples (not fully thought through). The full form of valid scopes may be

 - `read:current:Vehicle.Body.Trunk`
 - `read:target:Vehicle.Body.Trunk`
 - `read:meta:Vehicle.Body.Trunk`
 - `write:current:Vehicle.Body.Trunk`
 - `write:target:Vehicle.Body.Trunk`
 - `write:meta:Vehicle.Body.Trunk`
 - `write:current:Vehicle.Body.Trunk`
While you would also allow
 - `read::Vehicle.Body.Trunk` (read everything there is about this tree/signal)
 - `actuate:Vehicle.Body.Trunk` could be defined as `actuate` being a shortcut for `write:target`
 - `feed:Vehicle.Body.Trunk` could be defined as `feed` being a shortcut for `write:current`

erikbosch: I think it would be good to specify if you get `read` access for free if you have `write`-access (or provide/actuate-access), or if you need to have explicit `read` access as well.

erikbosch:
I am thinking if it would be good to write something about how we intend to make sure that you only can access data for a specific vehicle.

I am thinking of a use-case where each vehicle (of a specific model) has a KUKSA.val server (broker or server). Tokens for both in-vehicle apps as well as remote apps are generated by a central Key Management System (KMS), either in the plant or later. All are signed by the same private key or cert chain. How do we make sure that I cannot miss-use the token intended for my vehicle to get access to data in some other vehicle. If KUKSA.val in all vehicles of that model has the same public key we cannot limit access just by how the token is signed.

Do we need to define a specific field in the token like `kuksa-server-id` that can be validated by the server/broker? The ID could for instance be the Vehicle VIN, i.e. that the server/broker needs to get the VIN in some way, and then compare it to the VIN in the token. For the main-branch, a dummy VIN can be provided in the server/broker config, or we could just have a config to ignore checking id. Real deployments would likely use some other mechanism.

Another topic which came to my mind (but not related to this PR as such). When we implement authorization support in broker, it is preferable if the authorization code/methods exists in a separate file so that if a customer wants to use a different token format they could do that by just refactoring/replacing a single file. Regardless of what format we propose here - it is quite possible that real deployments would use something else that fits their KMS infrastructure and their specific requirements.

LuLeRoemer:
Thanks for the concept! I really like the idea of using OAuth2, as it follows a standardized and well-known way to authorize users/clients. However, I would also expect some info about the different types of authorization flows that OAuth defines in section [1.3](https://www.rfc-editor.org/rfc/rfc6749#section-1.3).
Maybe that is out of scope, but I was wondering which of the grant flows we want to use for what and whether user interaction is still involved in the entire authorization process. Do we expect there to be clients that have a preconfigured scope, such as I imagine a feeder? Because then I guess it makes sense to use the `Client Credentials Grant` flow for that, where resource owner and client are basically the same.
For "normal" apps, I would expect there to be some sort of manual approval by the resource owner when authorizing for the first time, presenting the user with a list of VSS paths that we would then see in the scope claim. What flow would be used here?
Another question I have is where you expect the Authorization Server to be. In the vehicle or in the cloud? Maybe that is not relevant for now.

doc/KUKSA.val_data_broker/authorization_alt_d.md
argerus: Haha. I actually added the exact thing you wrote here (kinda) to the document without seeing your comment here.

PS
I really need to find a workable way to enable Github notifications so I don't miss comments.

doc/KUKSA.val_data_broker/authorization_alt_d.md
argerus: Yes, definitely. It's a bit unclear, but the whole "Access token" section (now renamed JWT access token) describes how [RFC 9068](https://datatracker.ietf.org/doc/html/rfc9068) defines a JWT access token.

Or put another way.
* There is one RFC defining JWT, [RFC 7519](https://www.rfc-editor.org/rfc/rfc7519).
* And there's another, newer one, defining JWT **access tokens**, or strictly speaking *JWT Profile for OAuth 2.0 Access Tokens*, [RFC 9068](https://datatracker.ietf.org/doc/html/rfc9068).

My proposal, I guess, is that it would be beneficial to follow, or at least be compatible with [RFC 9068](https://datatracker.ietf.org/doc/html/rfc9068).

doc/KUKSA.val_data_broker/authorization_alt_d.md
argerus: I fully agree with the sentiment.

> Leaving it out now, would not preclude a path of adding it later, as it is just another scope. But I really think we shouldn't.

Agreed. I'll move it to another "future possible extension" instead.

argerus: One thing that is rather different in our use case, in comparison to most examples that you can find out there, is that we rely exclusively on limiting access by what is provided in the access token.

I would say the common use case is that the subject usually issuing an access token has a number of access restrictions to begin with. And these have to be checked in addition to the scope by the resource server.

So if we are not adding additional ways to limit the access rights for a group of users, one use case for these "deny" scopes would be that the authorization server automatically adds them to all access tokens for a certain group of users.

But that's perhaps a misuse of the feature, and there are better ways to solve it.

doc/KUKSA.val_data_broker/authorization_alt_d.md
argerus: I think the general idea with `read`, `actuate` and `provide` is that they provide a "higher abstraction level" than read / write access to individual fields.

There are a couple of benefits to this.

For one, these actions convey _what_ the client is allowed to do, as opposed to the technical details of what that currently means in practice (i.e. write current value, read actuator target etc). And I agree with @erikbosch that this should be better specified.

#### `read`
There seems to be general consensus around what `read` should mean, i.e. someone with this scope should be allowed read everything for the specified paths. If more granularity is needed, a more specific way to describe that can be added.

#### `actuate`
For `actuate`, I think it would be natural if anyone allowed to actuate something would also be allowed to monitor the actuator as it moves along (i.e. read the value). It also seems natural that they are allowed to read the actuator target and metadata, so in short, `read` should be included in `actuate`. If more granularity is needed, a more specific way to describe that can be added.

> So my understanding is we want to express, that we can write target value. the intent is to actuate, but that does not mean "actuation" really happens, because there is a stack of software below that will be the judge of that.

Yes, the intent of `actuate` is that a client with this scope is allowed to actuate. I fail to see how what you are describing is relevant in terms of either intent or access rights. Sorry.

> In that case write:target_value would be more exact

Yes, this would be more "exact", if by that you mean it would more exactly mirror the current implementation of `kuksa.val.v1`. But I can easily imagine a future where KUKSA.VAL can return a response indicating success or failure when a client calls `actuate(..)` or equivalent (see my other way to long design doc :) ). Then `write:target_value` doesn't make sense anymore.

In contrast, I think `actuate` makes sense today, and would continue to make sense as long as it covers the same intent. These improvements in KUKSA.VAL are definitely not guarantied to happen, but by having "scope actions" that correspond to the "intent", this would be a non-issue in terms of authorization at least.

And as described above, specifying it like `write:target_value` wouldn't include the other access rights that should be included part of being allowed to actuate.

#### `provide`
> So I assume we do use the term "Provider" for what has been a called a feeder so far. I would use the term provider for everything that connects to a "real" car, so also simple VSS services, i.e. a module that listens for some target value for opening a door, and then doing it. (for that he would not need a provide claim, just a read, so maybe if we don't want something along the lines of "write-current", this shall rather be feed.

Again, I think this line of thinking is very tightly coupled to the current implementation of KUKSA.VAL. As an example, I think it is highly desirable if providers of the type you describe would have the ability to "present" themselves as providers to KUKSA.VAL instead of "anonymously" subscribing to `actuator_target`, as this would enable a number of features in KUKSA.VAL. One can also imagine a way for these providers to provide feedback of whether the actuation was successful or not (again see the other way to long "design doc" :) ).

The point is, the access rights needed for those things would be included in `provide`, even though we don't even have that functionality today. And those things would probably not map to any specific fields.

I don't see how your description of a provider that happens to not publish any values is
anything which makes `provide` less useful or misleading today.

#### Alternative for fields
> another way to go about this, would be, to take a look at the "Using "access_mode" + "fields" for more granularity" section, and define that first, and treat the comon things as shortcut, just examples (not fully thought through). The full form of valid scopes may be
> * `read:current:Vehicle.Body.Trunk`
> * `read:target:Vehicle.Body.Trunk`
> * `read:meta:Vehicle.Body.Trunk`
> * `write:current:Vehicle.Body.Trunk`
> * `write:target:Vehicle.Body.Trunk`
> * `write:meta:Vehicle.Body.Trunk`
> * `write:current:Vehicle.Body.Trunk`
>
> While you would also allow
> * `read::Vehicle.Body.Trunk` (read everything there is about this tree/signal)
> * `actuate:Vehicle.Body.Trunk` could be defined as `actuate` being a shortcut for `write:target`
> * `feed:Vehicle.Body.Trunk` could be defined as `feed` being a shortcut for `write:current`

Yes, this also crossed my mind, and I think the design of this "escape hatch" of specifying more granular access rights with access mode per field could be useful. But I think it belongs in the same pot as the "deny" scopes. It could be useful, but we should think it through before we include it. The proposed uses of these, I think focus way to much on _how_ a client is allowed to do something opposed to _what_ they are allowed to do. I think that's my main concern, not the ability to limit access to fields.

So the bottom line is that I think these "higher level abstraction" that cover _what_ the client is allowed to do, would serve us best both in terms of:

* **Ease of use**

  I think it's more intuitive with
  
  `"actuate:Vehicle.Body.Trunk"` & `"provide:Vehicle.Body.Trunk"`
  
  than
  
  `"write:current:Vehicle.Body.Trunk"` & `"write:target:Vehicle.Body.Trunk"`.
  
  And I actually had to double check that it wasn't `"write:value:Vehicle.Body.Trunk"` or `"write:actuator_target:Vehicle.Body.Trunk"` or `"write:target_value:Vehicle.Body.Trunk"`.

* **Future proof**

  By not coupling the access rights super tight to the current implementation of KUKSA.VAL, (or the current version of `kuksa.val.v1`), we have a lot more leeway in making KUKSA.VAL better going forward without the need to also change the design of the authorization.

argerus:
@SebastianSchildt
I tried to incorporate some of your thoughts around separating the authorization for providers of "actuation" and providers of "values" as part of the new section: [_Using "subactions" for more granularity_](https://github.com/boschglobal/kuksa.val/blob/topic/authorization/doc/KUKSA.val_data_broker/authorization.md#using-subactions-for-more-granularity).

@erikbosch 
> I am thinking if it would be good to write something about how we intend to make sure that you only can access data for a specific vehicle.

Good point, I'll add a section about this as well.

If the resource owner only has access rights for certain VIN:s, this information could be added as a separate claim, possibly as part of something called "entitlements"
```yaml
{
  ...
  "entitlements": [
    ...
    "vin:5ATCX26E13S015222"
    "vin:5GZCZ43D13S812715"
  ]
```
KUKSA.VAL would then always check that the VIN of the current vehicle matches one of the entitlements.

And if the resource owner wants to restrict access further (for a certain client / access token), it can just be included as another scope string:
```yaml
{
  ...
  "scope": "... vin:5GZCZ43D13S812715"
}
```

doc/KUKSA.val_data_broker/authorization_alt_d.md
lukasmittag: I agree with the actuate and provide mapping instead of write:target and write:current. We had the same issue again at the Hackathon. There where a lot of questions around this. In summary most of the people understood after explanation what the difference is, but I think actuate and provide makes it more clear. I think the shortcut idea is a very good compromise, but could also lead to more confusion E.g. if one is using the shortcuts and another the more granular description in their specification. And then they look at each other's stuff.

argerus:
@LuLeRoemer 
> Maybe that is out of scope, but I was wondering which of the grant flows we want to use

Good input. I think the aim is to fully ingest (and understand) the standards so that we know what we are talking about before answering :)

Could also be a good thing to add to the docs, i.e. how this fits into the context of KUKSA.VAL.

> Another question I have is where you expect the Authorization Server to be. In the vehicle or in the cloud? Maybe that is not relevant for now.

Shouldn't matter for now. I would assume it would be outside the vehicle (easier to secure).

SebastianSchildt:
> This seems to address just this. So KUKSA.VAL would then always have to check that the specified audience claim matches the vehicle it is running on.

I want to point out, that we should not have a hard dependency on it, as I see this only as one deployment option, i.e. there are use cases where you want to strongly tie an app/client to a specific vehicle, but there are also options where you may not want to do it:
 - very static: In mass production oyu likely only want to flash one software revision, and not parametrize too many "per-vehicle" things
 - very dynamic: In an app-store like scenario (i.e. how Android and iOS are doing it), you would likely also only "bundle" apps with metadata being valid for all (fleet/manufacturer/type) vehicles. Now of course, you  you could package and distribute such app/config bundles on a per vehicle basis, but I'd argue you wouldn't do that if not strictly necessary. The work seems small, but if you ned to distribute 100 million App updates, 100 Millions HTTP downloads from some (caching) CDN are way easier, than generating 100 Million configs, and distributing them individually



tldr; Yes good feature, we should definitely have it, but let's make sure there is also a "clean" way of not using it, for applications/deployments where it does not help

doc/KUKSA.val_data_broker/authorization.md
```diff
@@ -0,0 +1,505 @@
...
+
+The access token contains everything the server needs to verify the authenticity of the claims
+contained therein.
+
+Example of what's available in a JWT access token:
+
+Header:
+
+```yaml
+{
```
SebastianSchildt, line 131: If we do list the header, we probably should have "one-sentence",  maybe reference to spec,  what those are.
Especially wrt to alg would also be the place to say which one(s) is specifically supported by us

erikbosch:
> tldr; Yes good feature, we should definitely have it, but let's make sure there is also a "clean" way of not using it, for applications/deployments where it does not help

I assume there are two ways of making it "optional". One is to actually have a config flag in databroker that enables/disables the check, another is to always use it but say that if you do not want any extra checks, just use `aud=kuksa` for all your tokens and use the default aud in Databroker

doc/KUKSA.val_data_broker/authorization.md
```diff
@@ -0,0 +1,505 @@
...
+and including this in the `"aud"` claim, KUKSA.VAL can verify that the access token isn't meant
+for another resource.
+
+KUKSA.VAL also needs to verify that the access token is intended to be used for KUKSA.VAL.
+An `"aud"` claim that would address both these requirements could look something like this (TBD):
+
+```yaml
+{
+  "aud": [
+    "vehicle/5GZCZ43D13S812715/kuksa.val.v1"
```
SebastianSchildt, line 225: see my previous comment. It might not always be desired to have this limited to a simple vehicle. So we should define, 

1. How it looks when the token is valid independent of vehicle (of course signature need to be correct)
  - would we just leave the aud claim out?
  - would it be a special string, i.e. only  "kuksa.val"?
2. regarding "TBD" how it looks like WHEN restricting access, current example given is `vehicle/5GZCZ43D13S812715/kuksa.val.v1`. I think VIN is a good basis (and if a deployment uses that we would expect VIN to live at a well-defined place in our VSS tree? Or would we also add a commandline/runtime configuration for our "vin" indepdendet of loaded VSS -> but yeah, that is a techical quuestion, should not matter for the concept). If choosing VIN, maybe it is good to add some hint that we mean kuksa.val (in case other (SDV) systems also have VIN based tokens), but I think the example is to much: `vehicle`does not add much information in this case, and even specific API as in `kuksa.val.v1` because my assumption is the scopes we have also apply to e.g. the sdv api that is still there, and would also apply to a hypothetical `kuksa.val.v2`, so maybe `kuksa/5GZCZ43D13S812715` is fine?

doc/KUKSA.val_data_broker/authorization.md
```diff
@@ -0,0 +1,505 @@
...
+to limit what resources an access token is authorized to access. The `"aud"` claim seems
+to be a good way to do this.
+
+A possible addition to this would be to be able to limit access to certain VIN:s by also
+introducing a `"vin:<VIN>"` scope.
+
+```yaml
+{
+  ...
+  "scope": "... vin:5GZCZ43D13S812715"
```
SebastianSchildt, line 371: Not so sure, this is needed. So if combined with other read/etc scopes would it _restrict_ those further to only one VIN? Or does adding this scope imply "for this Vehicle "do whatever"

I feel if we define how aud works, how to use "aud" to not limit to a VIN, and how an "access whatever" scope (or then non-scope? See comment above) works this is not adding any clarity.

If we see this as an alternative, I like the `aud` better

doc/KUKSA.val_data_broker/authorization.md
argerus: Good points.

I think the document should definitely be updated to clarify this. Perhaps it is even a mistake to include any example of what the `"aud"` claim can contain since it will probably vary between all deployments.

After thinking some more, I think the correct way to view them are as opaque identifiers that should be verified to match what the resource server is configured to accept. The resource server could for example be configured like this:
```
[authorization]
accepted_audiences = ["asd8f98cv789", "234098vcxc"]
```
and it would be up to the infrastructure to configure the resource servers and either issue access tokens with an audience that is accepted by all instances in the fleet, or with unique identifiers for all vehicles in the fleet.

> would we just leave the aud claim out?

The way I understand it, the JWT access token should always include the `"aud"` if we are to follow [RFC9068](https://datatracker.ietf.org/doc/html/rfc9068).

doc/KUKSA.val_data_broker/authorization.md
argerus: > Not so sure, this is needed.

Me neither. Is one more "possible additions" meant to evaluate the extensibility of the design.

> So if combined with other read/etc scopes would it restrict those further to only one VIN?

Yes, that's the idea. Should be stated more clearly.

> If we see this as an alternative, I like the aud better

It's just an additional proposal and it's something that would be added in addition to the `"aud"` claim (as stated in the description of this scope addition).

**Example use case**

Let's say that a resource server has been configured to accept the following audiences in the access token (as proposed in a comment above):
```
[authorization]
accepted_audiences = ["asd8f98cv789", "234098vcxc"]
```

If 20 resource servers (vehicles) are all configured in this way, it would mean that an access token issued for one of those would be valid for all. If a resource owner wants to issue an access token valid for just one of them, there is no way to do this using the `"aud"` claim. It would be if all resource servers had at least one unique `accepted_audiences` entry.

The idea of adding it as a scope as well, is to allow the resource owner to restrict the access token to just one of those vehicles (even though they are all hard-coded to accept the same `"aud"`).

```
"scope": "... vin:5GZCZ43D13S812715"
```

This would act in addition to all the other scope:s specified. It wouldn't include any specific access, i.e. `read`. It would just mean that if `vin:5GZCZ43D13S812715` doesn't match the VIN of the vehicle, nothing else present in the scope would be allowed either.

But perhaps a better design would be to just make sure that all resource servers have at least one unique `accepted_audiences`. But as you said somewhere, this might not always be possible in a mass production scenario?

doc/KUKSA.val_data_broker/authorization.md
```diff
@@ -0,0 +1,505 @@
...
+A common (perhaps the most common) scenario when delegating access using OAuth2 is that the
+principal owner of a resource has certain access rights that needs to be verified by the
+resource server. This can be accomplished using the user id, group membership, roles or other entitlements present in the access token as claims.
+
+In addition to this, the resource owner can further limit what's authorized by an access
+token by specifying a scope in the request to the authorization server. The issued access
+token will then only grant this subset of authorization to the client (e.g. a third party app)
+using the token.
+
+There are currently no plans in KUKSA.VAL to separately verify the access rights of the
```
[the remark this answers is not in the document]
argerus: > If somebody is in possession of a valid token with no scope claim (or empty? is that the same?), he access everything.

Yes, I think that is the idea behind using scope.

It's "traditionally" the case that the resource owner will verify the access rights of the subject _in addition_ to any scope provided. If that is to be added at any point in the future, it would be better to interpret an empty scope as "allow everything" in the current implementation.

With that said, whether an empty scope can even exist in a valid access token is restricted by the authorization server:

>If the client omits the scope parameter when requesting authorization, the authorization server MUST either process the request using a pre-defined default value or fail the request indicating an invalid scope.  The authorization server SHOULD document its scope requirements and default value (if defined). [[RFC6749](https://www.rfc-editor.org/rfc/rfc6749#section-3.3)]

So, in short, this should be better explained in the document.

doc/KUKSA.val_data_broker/authorization.md
argerus: Agreed

doc/KUKSA.val_data_broker/authorization.md
SebastianSchildt: "works for me", Implementation-wise, it would imply adding an optiional (commandline?) paameter to databroker, specifying which "aud" it should feel responsible for, right?

doc/KUKSA.val_data_broker/authorization.md
SebastianSchildt: See comment above, maybe we can do two things
 - Allow a list of audiences to be configured when starting databroker
 - Without specifying any, just define a default audience, i.e. `kuksa.val`(not married to the string, could also be `DEADBEEF`or `AFFE`...? , which we can also use in example tokens

doc/KUKSA.val_data_broker/authorization.md
argerus: Yes, exactly.

doc/KUKSA.val_data_broker/authorization.md
argerus: Yes, I don't immediately see an issue with having a default audience.

doc/KUKSA.val_data_broker/authorization.md
```diff
@@ -0,0 +1,505 @@
...
+The scope(s) used in in KUKSA.VAL to limit access rights, utilize "prefix scopes" that
+consist of an action prefix followed by an (optional) VSS path. This provides a flexible
+way to limit for which VSS signals the action is allowed.
+
+| Scope                 | Description                 |
+|-----------------------|-----------------------------|
+|`<ACTION>[:<PATH>]`    | Allow ACTION for PATH       |
+
+Examples of KUKSA.VAL scope strings:
+
```
SebastianSchildt, line 244: Ok, as promised, "scopes", I feel this is the last "big thing" we need to align on. Understanding, down the road we can always add new scopes, what is the initial set of useful scopes.

let me mainly refer to what is in the current draft, and also the reply here https://github.com/eclipse/kuksa.val/pull/471#discussion_r1096407754

I sense a consensus, that `read:<something>` is a good idea and well understood.

from the previous comment
> I interpret most of the proposed changes as a desire for a more direct mapping between scope and the current implementation of kuksa.val.v1 where setting a field is the API.

I can follow this line of argument, and actually my first feeling on reading you proposal was, you intend to align on some "eventually-to-be" API wis provide, while I was going the more generic route with saying what to read/write. That was mainly influenced by checking i.e. the Github scopes our other Web APIs where, quite often scopes revolve a lot around "read:that" or "write:this". 

But I agree, this is about intent. So let's check the other intents of our users/clients: 

Another thing an application likely wants to do is "actuate", that is trying to influence the state of something in a vehicle. I would agree, that `actuate:something` is a good scope for that. My current understanding would be, that this implies, in terms of current API/data model, that a client is able to write the target value, and also read it (I would just assume restricting "read" here, does not make much sense), and possibly also being able to read the current value.


That leaves the "southbound things". We are converging on calling these things providers, see discussion in #479 and notes in https://github.com/boschglobal/kuksa.val/wiki/Terminology . There is a difference beween

 - Providing sensor values: E.g. that is what DBC feeder does, or really is the definition of what have been called "feeders" before. I think the scope must make that clear, and not only say provide. I would suggest `provide-sensor` (in the current version of this draft there seems no difference, the wiki page suggests `provide:value`, but syntax-wise, I think it is more elegant if ":" always separates Path from action and not "sub-actions". `provide-sensor` would imply - in current API terms - to be able to write the current value of a sensors. I would assume that also applies being able to read it, although I thin it does not matter so mich
 - Providing services to actuate things. I suggest calling the scope `provide-actuator` , the wiki suggests, `provide:actuation`, argument for ":" applies as above. And I feel using the exact term "actuator" as sensor in VSS makes things consistent, and what in fact you do is providing an actuator to the SDV world. Currently this is probably identical to "read", but having a separate scope, can be useful later, in case there will be a more synchronous/special API  to report on actuation success/errors etc.

I am not convinced we should have an aggregate `provide` scope for both aspects (now, could still be introduced later). Currently an example for a component needed both is our seat service example: It moves the seat (`provide-actuation`) and reports back (`provide-sensor`), however I would assume (can't prove it though :) ), that in real deployments both aspects would often be handled by different components, just due to the different safety requirements. maybe the actuation services calls a SomeIP service, or some exiting lower level API, while positions are on CAN/Ethernet somewhere anyway and can be parsed directly and cheaply in QM domain anyway.  Also it would leave the road open to other `provide-something`capabilites, although I can't think of any right now.

last point, which I would like to at least define in initial scopeset: Modification of VSS tree. While we may not implement fine-grained modification of every VSS metadata aspect via API in the short term, I do think the overlay concept, i.e. the ability to split a large VSS model into different parts is an important/powerful aspect of the VSS ecosystem. We already support loading several models during startup, I could see us extending this to a runtime API soonish, where you might instruct db to load additional JSONs via an API call (assume, some new/specific <del>feeder</del>provider is deployed for an app, it might want to "register" datapoinnts/extend the model). I think it wold be relatively straight forward, to limit which parts of the VSS tree an app is allowed to touch. For ease of use, maybe start with the `modify` mentioned below (I fell that "includes" create). I think just the term is a bit to generic, maybe `modify-model` is better

tldr; my suggested "initial actions" for scopes are
 - `read`
 - `provide-sensor`
 - `provide-actor`
 - `modify-model`

argerus: > I would suggest `provide-sensor`

I think this is a very problematic choice of wording.

> So provide-sensor might be better catching the intent and would imply - in current API terms - to be able to write the current value of a signal. I would assume that also applies being able to read it, although I think it does not matter so mich

On the contrary, I would say it _fails_ to catch the intent and is even misleading. Using `provide-sensor` fails to convey that this scope grants authorization to provide values for any type of VSS signal, be it an `Attribute`, `Actuator` or `Sensor`.

> I think -value may be accused of being to reliant on what the current API is. Just trying to use your arguments against you :)

The _value_ of a signal is established VSS terminology (right?), and not specific to the current API(s).

> in the current version of this draft there seems no difference, the wiki page suggests provide:value, but syntax-wise, I think it is more elegant if ":" always separates Path from action and not "sub-actions"

Did you see (or miss) this part of the draft?

>| Subactions for `read` | Description                        |
>|-----------------------|------------------------------------|
>| `read:field:<FIELD>`  | Allow client to only read a certain field for matching signals |
>
>| Subactions for `provide` | Description                     |
>|--------------------------|---------------------------------|
>| `provide:value` | Allow client to provide the value (but not respond to actuation requests) for matching signals. (includes `read`) |
>| `provide:actuation` | Allow client to respond to actuation requests (but not provide the value) for matching signals. (includes `read`) |
>
>| Subactions for `edit` | Description                     |
>|-----------------------|---------------------------------|
>| `edit:field:FIELD` | Allow client to edit metadata field FIELD for matching signals. (includes `read`) |
>
>| Subactions for `edit` | Description                     |
>|-----------------------|---------------------------------|
>| `edit:field:FIELD` | Allow client to edit metadata field FIELD for matching signals. (includes `read`) |

I will update the document with a better description of the thoughts behind structuring the scopes like this. In short, the `:` is used to describe a _hierarchical access right_ structure.

### Hierarchical Access Right
The core of the concept is that the top of the access rights hierarchy includes all the access rights below it.

So for example, `read` is the top (or root) in an access rights hierarchy including read access to everything. `read:Vehicle.Speed` is a child in that hierarchy and is included in both `read:Vehicle` and `read`. Furthermore, this can be expanded to include `read:field:FIELD` or `read:FIELD` which can be combined with either paths or tags (also separated by `:`) to further specify the specific child access right in this hierarchy.

Analogous to this, `provide` is the root in another access rights hierarchy. Both `provide:value` and `provide:actuation` is included in this root.

Your argument:

> I would assume (can't prove it though :) ), that in real deployments both aspects would often be handled by different components

This doesn't make an access rights hierarchy invalid or harder to use for the specified use case. Even though these aspects would _often_ be handled by separate components, using an hierarchy like this will make it easy to describe the access rights needed for components that provide **both** values and actuation.

argerus: > I suggest calling the scope provide-actuator , the wiki suggests, provide:actuation, argument for ":" applies as above. And I feel using the exact term "actuator" as sensor in VSS makes things consistent, and what in fact you do is providing an actuator to the SDV world.

Again, I think this terminology is very misleading.

Using the term _actuator_ to only refer to the actuation part is **not** consistent with VSS terminology.
Using the term _sensor_ to refer to the value of a signal is **not** consistent with VSS terminology.

An `Actuator` in VSS terminology includes _both_ the ability to set and read it. A provider providing only one of these features is not providing a complete actuator "to the SDV world".

SebastianSchildt: > The value of a signal is established VSS terminology (right?), and not specific to any current API.
I think that my be the point where I do not fully agree. 

We think about it in this terms in KUKSA.val (and VISS partially shares the same worldview). Sensors can be read, actuators can be, well, actuated. 
VSS itself however is just a data model, and does not say anything about applications. A signal has a name and a type. If you e.g. use this model in your data warehouse/backend, then something being an "actuator" might have some relevance there (maybe counting how often it has been used), but a (current) value might be not at all relevant here. Similar for sensors, maybe only statistical moments are stored. If VSS is used in an authorisation server, maybe "actuator" as VSS type is only a flag making sure, certain apps never get to use it.

(I know there are hints that might lead you otherwise e.g. _examples_ how you might want to use it here https://covesa.github.io/vehicle_signal_specification/rule_set/data_entry/sensor_actuator/ or this things even called Value restrictions "https://covesa.github.io/vehicle_signal_specification/rule_set/data_entry/allowed/").

## Suggestions
Therefore, even taking all explanatory text around VSS as normative, I feel `provide:value` to generic and I would have no good intuition what it means.

Would it be possible to make clear it `provide:currentvalue` or `provide-currentvalue` (wrt to ":" not yet sure if provide hierarchy has that much practical use, but I can agree it,  does not hurt)
 or alternatively in line with current  terminology revision `provide:state`
I would interpret all of these to be able to "write" the current value for sensors/actuators/attributes 

Also I think, in my shortlist I forgot  `actuate` for VSS consumers/users waning to "write" an target_value. It hink that is fine as is in the document.

That leaves the scope for providers providing actuation in a vehicle. As I just called this thing `actuation-provider` in terminology I would yield, and accept `provide:actuation` as the better solution. My understanding would be this will allow reading (at least) target_value, but likely everything as "read", but not allow writing "currentvalue".

So my new shortlist would be

```
read
actuate
provide:state or provide:currentvalue
provide:actuation
modify-model
```

argerus: > Therefore, even taking all explanatory text around VSS as normative, I feel provide:value to generic and I would have no good intuition what it means.

On that I completely agree.

In the terminology section, I am hesitant to use `Value provider` as it is, just like you say, to generic and the meaning not immediately obvious (it sounds more like something out of a PR slide). `Data provider` is proposed as an alternative there, so I guess `provide:data` could be another item to add to the list of alternatives to ponder. I even think that term was used spontaneously at FOSDEM (by some external party).

argerus: Regarding `modify-model`.

I haven't fully articulated why it could make sense to split this functionality into separate access rights (or actions). This is what it looks like in the draft at the moment.

| Action    | Description                                   |
|-----------|-----------------------------------------------|
| `edit`    | Allow client to edit metadata of a VSS entry. |
| `create`  | Allow client to create a VSS entry (under a certain path). If a VSS entry already exists, `edit` is needed to change it. |

The reason for differentiating between creating and editing entries is that creating an entry shouldn't have a huge effect on already running applications in the vehicle. Editing an entry on the other hand, could have a huge effect as it has the potential to invalidate the state of a lot of applications. That's why the right to create an entry doesn't automatically include the authorization to edit it.

I think this is an area where we need to think a bit to get a sensible implementation.

Regarding naming:
`create` seems intuitive enough, e.g. `create:Vehicle.Speed` or `create:Private.*`.

`edit` might be a bit to generic

SebastianSchildt: I fully agree with create as a way do allow "adding" things to the model. My assumption is that the create scope will not allow "overwriting" or changing existing parts in the model. That is useful if you are composing your model during runtime, to ensure "Nothing goes wrong"

I guess "modify-model" was my way to say "edit". One common approach VSS side could be, that you load a base model, for example the VSS 3.0 and on top of that load an overlay, that will set the "defaults" for all attributes and attributes like min/may value for a specific vehicle. That something like edit rights, and I was not sure if that _only_ allows editing existing leafs and branches or _also_ extending them.

I was searching for a scope that basically says "Vehicle.ADAS" is yours, do whatever. And as the most generic I was thinking it might be good to have implemented first.

But maybe the answer is, we do need to start with both
`create:` to _add_ but not change, `modify:` (or edit) to change/ammend existing nodes. 

Or build another hierarchy? 
 - `modify:create`  
 - `modify:amend`
 -  `modify:change`

argerus: I think `modify` and `create` are so different that it would cause more issues than benefits to try to combine them. The combination `modify:change` is perhaps a good example of this :)

Regarding `modify` / `edit` / `update` or whatever we want to call it, a few questions needs to be answered (I think) before we decide to add support for it:
* Would this scope allow changing the `entry type` of an already existing entry.
* Would this scope allow changing the `data type` of an already existing entry.

Maybe such changes should be disallowed completely. What other changes would be problematic?

These issues don't exist with regard to `create`, and my gut feeling is that a good way forward could be to start by adding `create`, if that is something which is needed.

SebastianSchildt: Yes, I think on create I think we have a consensus, that it is useful and what it means, so

```
read
actuate
provide:data
provide:actuation
create
```

wrt to "edit/modify" yes I think changing type/datatype/unit are "dangerous" operations (not implying we should not allow it, but maybe needs different scopes).

One near-term usecase I see for something where `create` is not enough is something which in VSS community is sometimes referred to "deployment information", most simple example: Your VSS model has attributes, most of them are by nature quite static (e.g. VIN), so you provide an overlay that modifies "default", i.e. the value for them (and defaults have been removed from VSS see https://github.com/COVESA/vehicle_signal_specification/pull/284)
The same goes for deployment specific information such as min/max

You can do this of course "pre-deployment", i.e. providing an adapted VSS file already, but there might be reasons somebody wants to do it "online", because that basically follows the pattern"config bits" you already have today in some ECUs enable/disable features or do certain settings.

I would have more (far-fetched) examples. Doing instances in vehicle would be an even more invasive operation.

My wish would be, that we do have some scope that might enable at least the first usecase. Either declaring it a "dangerous" operation and note you should use with care, or maybe limit changing "dangerous" things listed above. Any idea how to capture this in a obvious, self-explanatory, useful scope?

argerus: > Any idea how to capture this in a obvious, self-explanatory, useful scope?

Not spontaneously, no. My initial thought was:

What you are describing is just an indirect way to change the value of an attribute. That functionality is already covered by the `provide` scope, i.e. `provide:Vehicle.TheAttribute`.

> and defaults have been removed from VSS see https://github.com/COVESA/vehicle_signal_specification/pull/284

It's only removed for sensors & actuators, right? It's still there for attributes, as was the original intention, right.

Changing the default value wouldn't even be persisted in the current implementation. Maybe that's another dimension to think about (runtime changes vs persistant changes). And maybe the dangerous things wouldn't be so dangerous if they only applied to an offline "configuration" and where not applied immediately (to a currently running instance). I don't know, thinking out loud here.

doc/KUKSA.val_data_broker/authorization.md
SebastianSchildt: The section still does not say exactly how we are (intending) to implement it, there is unclear wording like  "needs to" "could" and "TBD"

It needs some "KUKSA.val **will** " 

I think we should state
 - if you do not specify otherwise, KUKSA.val will listen to the scope of `kuksa.val` (or whatever else generic term)
 - We can add some wording around "The choice and uniqueness of audiences is a deployment decision. The safest way is using individual audiences for individual Vehicles. Those could be based for example on the VIN. However, depending on the use case, audiences valid for a whole fleet or certain subsets/models of vehicles may be more useful."

doc/KUKSA.val_data_broker/authorization.md
SebastianSchildt: See comment above wrt to audiences

SebastianSchildt:
We have a section "Examples of other API scope implementations" right in the middle. I Siggest that is not a good fit, we can remove it, or put it somewhere down in references/Links

"Limit access (additional possibilities)"

I still vote to move this outside this document, I think the header, or maybe ""Limit access (potential future extnesions)" can remain, but then basically just linking to an issue where we copy the text (or alternatively a clearly named markdown)

I think our goal should be, that the document we are doing in this PR can be considered "normative" for KUKSA, so that  if we do not do something as described in here, it is a bug (in implementation or, well, in this document). So I think we should not clutter it with `#ifdef'  ideas, that we do not intend to implement immediately.

doc/KUKSA.val_data_broker/authorization.md
```diff
@@ -0,0 +1,561 @@
...
+have unique audiences for each instance (for example based on the VIN of a vehicle). However,
+depending on the use case, audiences valid for a whole fleet or certain subsets/models of vehicles
+may be the only option."
+
+Example of an `"aud"` claim:
+
+```yaml
+{
+  "aud": [
+    "5GZCZ43D13S812715/kuksa.val.v1",
```
erikbosch, line 254: On line 154 we use "vehicle/5GZCZ43D13S812715/kuksa.val.v1" as example. Any particular reason to have "vehicle" as prefix in one of them and not the other? If not possibly use it in both or none.

argerus: Yes, the reason is that I changed one of them but forgot the other :)

SebastianSchildt (approved):
I think this is largely aligned/fine now. Let's do further refinements if needed in separate PRs, so we can get this “blob" out of the way. I noted down some potential improvements in #491

PR #472, opened 2023-01-30
Token example draft

SebastianSchildt:
PR Just for the sake of discussion.

A try to describe & update the token format

To read it "rendered": https://github.com/boschglobal/kuksa.val/blob/feature/token_draft/doc/tokens.md

Things I sort of "randomly" decided here:
 - is it better to use an `*` if you really mean a subtree? Currently I assume addressing a branch means that
 - terms, is "access modifier" good?
 - verbosity: I went for `get_current` instead of inventing some shortcode such as `Rc`

### DIFF
--- a/doc/tokens.md
+++ b/doc/tokens.md
@@ -25,104 +25,113 @@ On a high level, these interactions with the VSS tree need to be considered
   "iss": "Eclipse KUKSA Dev",
   "iat": 1516239022,
   "exp": 1606239022,
-  "kuksa-vss":  {
-     "Vehicle.Speed": ["get_current"]
-  }
+  "vss-read":  [
+    "Vehicle.Speed"
+  ]
 }
 ``` 
 
-The authorisation is handled by the `kuksa-vss` claim. In KUKSA.val we only handle explicit access rights, i.e. a token directly encodes what the owner of said token can do. 
+The authorisation is handled by `vss-*` claims. In KUKSA.val we only handle explicit access rights, i.e. a token directly encodes what the owner of said token can do. 
 
 As KUKSA.val is a rather low level component, we are not making assumptions about the security model around your application. Depending on your deployment, you maybe have Role-baed access models, or you work with bearer tokens. You need to make sure, that such access rights are rolled out into explicit descriptions by an IAM/IDM or some other form of policy database in your system.
 
 
+## vss claims 
 
-## kuksa-vss claim format
+The following claims are supported.
 
-Entries in the `kuksa_vss` claim, follow the following format
+### vss-read
+This implies a client is able to read (get/subscribe) the current value for sensors, actuators and attributes, as well as the target value for actuators.
 
-```json
-"kuksa-vss":  {
-     "<VSS_PATH>": ["<ACCESS_MODIFIER_1>", "<ACCESS_MODIFIER_2>", "..."]
-  }
-```
+**Discussion: Might also be split into read-sensor, read-target? Should this also imply read-meta**
+
+### vss-provide-sensor:
+This implies a client is able to set the current value for sensors and actuators.
+
+**Discussion: Might alternatively be called just "provide", which would fit providers that are "feeders", but  currently in my mind a "provider" could also be of the service variety, and be a component that reacts on updated target values. Might also be jsut "write-current" which may be more obvious if you ar enot into VSS/KUKSA lingo**
+
+### vss-actuate
+This implies a client is able to set the desired target value for actuators.
+
+### vss-read-meta
+This implies a client is able to read VSS metadata, including names of children, for the given paths. Reading meta data without any additonal rights is useful, if you want to give a means of discoverability to a client, wihtout neccesarily granting rights to acces the data itself-
+
+### vss-modify-meta
+This implies a client is able to change metadata of a signal
+
+### vss-create-signals
+This implies a client is able to create any number of new signals (sensors, actuators, attributes) including a branching structure under the given paths.
+Create signals also implies modify-meta. **Discussion: or does it not?**
+
+
 
 ## VSS paths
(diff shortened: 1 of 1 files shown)


SebastianSchildt:
if you check this also check https://github.com/eclipse/kuksa.val/pull/471/files

SebastianSchildt:
Hah, interesting, it seems a key difference is, do we model the rights as `Path->applicable rights`, or `Action"->applicable paths`.  is this a random choice, or are there arguments for either?

argerus:
> Hah, interesting, it seems a key difference is, do we model the rights as Path->applicable rights, or Action"->applicable paths. is this a random choice, or are there arguments for either?

I added [Alternative B](https://github.com/boschglobal/kuksa.val/blob/topic/authorization/doc/KUKSA.val_data_broker/authorization.md#alternative-b) as well, which is neither (or both) :)

I think I prefer that over the others.

SebastianSchildt:
I modified this approach a bit. I also started with the claims.

I decided agains the "rule" alternative with "allow", "deny". "Expression-wise" it can probably express all the same, but so far discussion wise, it has always helped me to say "as a default you can do nothing, unless explicitely granted", I feel having `allow` and `deny`as main categorieds muddies the waters a little bit.

I have also not (yet?) added the similar approach with "!" form the first approach. My arguments are, the cost is implementation, but the gains are only: Makes writing _certain_ rulsets by hand easier (but then again, that is a once per application, and likely generated anyway if there is a more complex backend), and second it can express _certain_ rulesets shorter, but I am do not think that is an optimisation goal right now.

I tried giving clear definitions how `*`

Also I needed to make some choices regarding names, etc. Whereas they seem a little "random" I tried adding some discussion points in the text in *boldface*, that of course should not be in the final version. Please especially look at them

argerus:
> I decided agains the "rule" alternative with "allow", "deny". "Expression-wise" it can probably express all the same, but so far discussion wise, it has always helped me to say "as a default you can do nothing, unless explicitely granted", I feel having allow and denyas main categorieds muddies the waters a little bit.

I don't see anything preventing "deny by default" by introducing `allow` / `deny` rules (in one way or another).

But more importantly, if the goal is to somewhat model this on how OAuth etc works, the default is to grant everything. Remember that in order for "a user" to even _get_ a signed JWT token, they would needs all the access rights necessary to retrieve that from an authorization server (and can most probably request an "access to everything" token).

A JWT token granted without any scopes means that it isn't restricted to any subset of full access rights (i.e. it's not scoped). Are we not going to follow this practice?

> Makes writing certain rulsets by hand easier (but then again, that is a once per application, and likely generated anyway if there is a more complex backend), and second it can express certain rulesets shorter, but I am do not think that is an optimisation goal right now.

I don't think the "ease of writing it by hand" was any influencing factor in neither of the alternatives. Quite the opposite actually, since it's explicitly stated that the claims are generally generated based on the scopes provided to the auth server on issuance.

With regards to the size of the token. There's a hard limit on how much can be put into a HTTP header, so I think it should definitely be a priority to make the token as small as possible (but not smaller).

The way I see it, there are two or three major design considerations we should optimize for.

1. Clarity. I.e. how easy is it to both express and understand what permissions are granted. 
2. Expressibility / Extensibility. Is it possible to express all the permissions/restrictions users want to. If not, is there a clear path to adding it in the future.
3. Brevity / Size. As mentioned, there's a hard limit on HTTP header size. Brevity usually also helps with clarity, since it's easier to read (and understand) a smaller amount of information.

Maybe we should start the design by discussion what our actual design goals are here :)

argerus:
> A JWT token granted without any scopes means that it isn't restricted to any subset of full access rights (i.e. it's not scoped). Are we not going to follow this practice?

With that said, it's up to the authorization server to actually allow such an authorization token.

Relevant section from [RFC6749 - OAuth 2.0 Authorization Framework](https://www.rfc-editor.org/rfc/rfc6749#section-3.3), if that is relevant.
> If the client omits the scope parameter when requesting authorization, the authorization
server MUST either process the request using a pre-defined default value or fail the request
indicating an invalid scope.  The authorization server SHOULD document its scope requirements
and default value (if defined).

doc/tokens.md
```diff
@@ -0,0 +1,137 @@
...
+
+
+## vss claims 
+
+The following claims are supported.
+
+### vss-read
+This implies a client is able to read (get/subscribe) the current value for sensors, actuators and attributes, as well as the target value for actuators.
+
+**Discussion: Might also be split into read-sensor, read-target? Should this also imply read-meta**
```
argerus, line 46: I think the general concept of having an overall `read` which allows reading everything is practical. That would include all metadata.

If there is a need for more granular permissions, using the field name is seems to be the most workable way to do it (see [fields for granularity](https://github.com/boschglobal/kuksa.val/blob/topic/authorization/doc/KUKSA.val_data_broker/authorization_alt_d.md#possible-extension-using-access_mode--fields-for-more-granularity)).

The proposed alternatives doesn't seem to follow any consistent pattern.

Would `read-sensor` only allow reading sensors (i.e. not actuators & attributes)?

doc/tokens.md
```diff
@@ -0,0 +1,137 @@
...
+
+### vss-provide-sensor:
+This implies a client is able to set the current value for sensors and actuators.
+
+**Discussion: Might alternatively be called just "provide", which would fit providers that are "feeders", but  currently in my mind a "provider" could also be of the service variety, and be a component that reacts on updated target values. Might also be jsut "write-current" which may be more obvious if you ar enot into VSS/KUKSA lingo**
```
argerus, line 51: I think it's confusing with the name `vss-provide-sensor` if it includes permissions to provide attributes and actuators as well. `vss-provide` would make more sense.

Not convinced that `write-current` fits into any consistent naming scheme either.

If there is a need for more granular permissions, again I think it would be more workable to add an additional syntax where the individual fields are referenced along with access mode. But this should be in addition to the (probably) most common way of granting "provider rights" for certain paths regardless of whether they are attributes/sensors/actuators.

SebastianSchildt:
Discussion continues in #471

PR #474, opened 2023-02-09
Add timestamps to actuator target values

lukasmittag:


kuksa_databroker/databroker/src/broker.rs
```diff
@@ -52,10 +52,15 @@ pub struct Datapoint {
...
+#[derive(Debug, Clone)]
+pub struct ActuatorTarget {
+    pub target_datapoint: Datapoint,
+}
+
 #[derive(Debug, Clone)]
 pub struct Entry {
     pub datapoint: Datapoint,
-    pub actuator_target: Option<DataValue>,
+    pub actuator_target: Option<ActuatorTarget>,
```
argerus, line 63: This isn't wrong. But the data type of actuator target is `Datapoint` (in the `.proto` file) so an alternative would be to just use `Option<Datapoint>`. I think that would make it easier to reuse the conversion functions already in place for `Datapoint` (and would make the data type less nested).

lukasmittag: Yes, I thought about this too. Would it make definetly less nested. Then I would suggest to rename the ```datapoint``` in ```value``` or something else because otherwise I think the naming can be confusing.

argerus: Hmm, not so sure about that.

I think the general terminology in VSS (and elsewhere) is to call "value + timestamp" a datapoint. See for example VISS, [here](https://w3c.github.io/automotive/spec/VISSv2_Core.html#dfn-data-point) and [here](https://w3c.github.io/automotive/spec/VISSv2_Core.html#data-representation) for example.

Now. VSS doesn't have the concept of actuator targets, but if we want to introduce a new thing which is also a "value + timestamp", one way to look at it is that it also has the _data type_ `DataPoint`.

I suppose you see it as more of a _variable name_. Both ways to look at it makes sense I think. The alternative is to just keep it as you have already done, or maybe make it a [new type](https://doc.rust-lang.org/rust-by-example/generics/new_types.html) `ActuatorTarget(Datapoint)` instead.

lukasmittag: Okay, I see the point about ```Option<Datapoint>``` for the target value too. I will change it to this and keep the actuator_target variable name. Should be enough to understand.

argerus (commented):
Looks good to me. Rebase, and the build should succeed as well.

PR #475, opened 2023-02-10
Add initial authorization

argerus:
* Add support for authorization Bearer token
* Add argument for JWT public key
* Validate JWT access token when authorization enabled


kuksa_databroker/databroker/src/main.rs
```diff
@@ -247,6 +248,15 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
                 .value_parser(clap::builder::NonEmptyStringValueParser::new())
                 .required(false),
         )
+        .arg(
+            Arg::new("jwt-public-key")
+                .display_order(5)
+                .long("jwt-public-key")
```
lukasmittag, line 254: In the C++ version we use --insecure and then specifying the public key; same in kuksa client. I do not know if it is needed to have consistency. But in my first try I thought I need to use --insecure=false as 'always'.

kuksa_databroker/databroker/src/main.rs
```diff
@@ -319,9 +329,35 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
...
+    };
 
-    grpc::server::serve_with_shutdown(&addr, broker, shutdown_handler()).await?;
+    match jwt_public_key {
+        Some(pub_key) => {
+            let token_decoder = auth::token_decoder::TokenDecoder::new(pub_key)?;
+            grpc::server::serve_authorized(&addr, broker, token_decoder, shutdown_handler())
+                .await?;
+        }
+        None => {
```
lukasmittag, line 357: I see here a little bit of a problem maybe. If you forget to specify a key file you get the unauthorized version. This could be dangerous for guys that do not read the output from server and then just assume it is working. Maybe it is better to use the insecure flag. Like mentioned above. I'm not entirely sure it's better but just a thought.

kuksa_databroker/databroker/src/main.rs
argerus: `--insecure` specifies whether to allow non-encrypted communication, i.e. not using TLS. It's unrelated to authorization.

lukasmittag: Yeah, you're right. It would be:   '--cert-path arg (=".")                [mandatory] Directory path where 
                                        'Server.pem', 'Server.key' and 
                                        'jwt.key.pub' are located.' in the kuksa-val-server. I mingeld things up my bad.

kuksa_databroker/databroker/src/main.rs
argerus: > If you forget to specify a key file you get the unauthorized version. This could be dangerous for guys that do not read the output from server and then just assume it is working.

I agree. As far as I understand it isn't even possible to disable authorization in the C++ version. That is the more sensible option, which I think should be the default here as well. However, in order to not break existing clients, I think we should wait with making it the default. Perhaps even updating the warning message to that effect (i.e. the warning will turn to an error in the future).

And as I wrote in the comment above.
> `--insecure` specifies whether to allow non-encrypted communication, i.e. not using TLS. It's unrelated to authorization.

Once we add support for TLS, the default should eventually also be to disallow unencrypted communication. But by the same logic, there should be a transition period where it only generates a warning before actually breaking clients by making it the default.

SebastianSchildt: One sort-of future proof way might be to include a commandline parameter that describes the access to unauthorized clients, something like

```
--unauthorized_access [ALL,NONE,READ_ONLY,<custom_token>
```

and we could start with "ALL" as default, then sometime later print a warning, that it will change to "NONE" in a later version.

argerus: Well, is this type of unauthorized access a feature we want to have long term?

Regardless, I think it should be added as a separate feature in that case.

argerus: 1. Add authorization. It is optional and a warning is printed informing the user that authorization will be required (by default) in the future.
2. Add flag to allow unauthorized access (i.e. what you propose). The warning (above) is removed when this flag is used. Clients can start using this flag if they intend to continue allowing unauthorized access (when it becomes required by default).
3. Make authorization required by default.

SebastianSchildt: As discussed, this might be a separate feature moving forward. I put the ideas from this discussion here https://github.com/eclipse/kuksa.val/issues/487

SebastianSchildt:
@argerus can this be closed, because the functional changes had been included in #477  already?

argerus:
> @argerus can this be closed, because the functional changes had been included in #477 already?

Yes

PR #479, opened 2023-02-13
First draft of terminology

SebastianSchildt:
This is a first stab at defining the terminology.

Goals was to give definitions usable on "PPT level", that we can then hopfully use consistently accross our repos and documentations. I tried not going too deep on specific technical details and limitations, excpet one or tow sentences here and there (i.e. compared to the discussions in Wiki, I did not mention specific API or auth scopes).

I did a recap of the most important (hopefully) VSS terms. @erikbosch : You are likely a good authority to check here. I did a lot of copy & paste though :)

Some things I was not sure about is in **bold**. Main things is "Client"/"User"/"App"

One thing which I do suggest, is a differentiation between "sensor-provider" and "actuator-provider", because I feel it makes a big difference in system design "southbound" of us, even though for databroker not much difference. Rational here is, if we switch over to the term "provider", as a "better way of saying feeder", I think it would not be good to have a dbc-provider (as in current feeder), and a seat-provider (as in current service), without being able to have a clear/definend way of saying what they are doing (in that sense dbc would clearly be a sensor-provider, while seat would be an actuator-provider (if we use the VSS defintion of every actor is also a sensor) or both)


doc/terminology.md
```diff
@@ -0,0 +1,86 @@
+# Terminology
+
+This pages gives an overview about the terms we use, when talking about KUKSA.val components, architectures and the data handled by KUKSA.val.
```
erikbosch, line 3: I suggest we add a disclaimer/note that KUKSA.val Server is not included in the description

doc/terminology.md
```diff
@@ -0,0 +1,86 @@
...
+Everything we handle in KUKSA.val is defined as a using VSS. A signal in VSS is an entity identified by a dot-separated string in VSS, i.e. `Vehicle.Speed`.
+
+A signal can be of different types. VSS defines the following signal types:
+ - *Sensor*: Sensors are signals to read values of properties in a vehicle. Values of sensors typically change over time. Reading a sensor shall return the current actual value of the related property, e.g. the current speed of a vehicle. Example: `Vehicle.Speed`. [[Source]](https://covesa.github.io/vehicle_signal_specification/rule_set/data_entry/sensor_actuator/)
+ - *Actuator*: Actuators are signals that are  used to control the desired value of a property. _Every Actuator in VSS is also a Sensor_. Some properties in a vehicle may change instantly. A typical example is position of a seat. Reading a value of an actuator shall return the current actual value (i.e. the sensor trait of that actuator signal), e.g. the current position of the seat, rather than the wanted/desired position. Example: `set Vehicle.Cabin.Seat.Row1.Pos1.Position`. [[Source]](https://covesa.github.io/vehicle_signal_specification/rule_set/data_entry/sensor_actuator/)
```
argerus, line 71: > _Every Actuator in VSS is also a Sensor_

I think this wording is very problematic.

It confuses `Sensor` with the _value_ of signal. A signal is either an `Actuator`, `Attribute` or `Sensor`. It cannot be two at the same time.

An `Actuator` signal has a `value` just _like_ a sensor (and an `Attribute` for that matter). And an `Actuator` behaves just _like_ a sensor with the exception that it's also possible to set it. But it is not a `Sensor`.

I think the same confusing wording is present above with **sensor-provider** (but it could be a typo, I'm not sure).

doc/terminology.md
```diff
@@ -0,0 +1,86 @@
...
+ KUKSA.val fully supports actuators, and alwaysgives you full control whether you are interacting with the _target value_ or the _current value_ of an actuator signal.
```
argerus, line 72: Not sure what this sentence adds in terms of terminology, really.

argerus (changes requested):

doc/terminology.md
```diff
@@ -0,0 +1,86 @@
...
+ ## Vehicle Signal Specification (VSS)
+ KUKSA.val handles datapoints that are defined using the COVESA Vehicle Specification (VSS).
+
+ VSS introduces a domain taxonomy for vehicle signals. It defines
+ * A syntax for defining vehicle signals in a structured manner.
+ * A catalog of standard signals (VSs standard catalogue) related to vehicles.
+
+It can be used as standard in automotive applications to communicate information around the vehicle, which is semantically well defined. 
+
+See the [the official VSS documentation](https://covesa.github.io/vehicle_signal_specification/) for all details. Here we will summarizes the terms and concepts most relevant for KUKSA.val.
```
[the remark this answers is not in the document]
argerus: The `User` / `Provider` roles can also be seen as analogous to the more common `User` / `System` roles.

Providers are the system services.
Users are the, well, users.

doc/terminology.md
SebastianSchildt: Put some words to that effect in the beginning, when introducing (VSS resource) Server. I do hope most aspects in this document are generic, and try only to point to databroker as “example” or “further reading” how it is done by KUKSA.val

doc/terminology.md
```diff
@@ -0,0 +1,86 @@
...
+
+Databroker's API is GRPC-based, which means it can be accessed over the network by other programs or systems. From an architecture point of view, there are different components, with different roles interacting with databroker explained in the following. While a given system or binary might combine several of these roles, it is helpful to think in these categorizations when desgning a system.
+
+For more specific examples of components fullfilling these roles also check [System Architecture](./system-architecture.md).
+**todo: probably system architecture not a good name for that file, and I did not update it with the terms in this doc**
+
+
+### Clients
+**Discussion: App (may be to narrow)? User (i thinkw as called that in wiki. but clashes then with "end-user" or developer using KUKSA.val? On the other hand client may also be a good "overall" term for everything contacting our API**
```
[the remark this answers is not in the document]
SebastianSchildt: Good idea. Added those definitions. Slightly more verbose

doc/terminology.md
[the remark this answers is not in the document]
SebastianSchildt: In this revison I have chosen “Consumer” instead of (VSS) User. I think replacing every mention of “Consumer” with User” works as well. Reasons for going with consumer
 - It feels a nice counterpart for “provide”
 - No danger of clashes with human endusers
Potential reasons against
 - One might not feel comfortable with a “consumer” doing actuation. (I think it is ok, becasue as “consumer” of my power drill I am also free to make holes.)

But maybe somebody there is a term I haven’t thought about

I tried not to get too deeply technical with mentioning “subscribe” or“set”.  My thinking model was rather “digital twin” (without mentioning that particularly loaded word :) ) whereas my goal is to learn something  or changing something  about the true state via interacting with VSS. Thus, the role of Providers basically becomes syncing state.

Tell me if it worked.

doc/terminology.md
SebastianSchildt: The two types of providers are harder. I think that is also partially the point where we diverge in the authorisation PR. I don’t think “value” is defined/clear cut term in VSS. But I will try writing something about that in he other PR. 

For this PR let's see how to tackle it.

>> Every Actuator in VSS is also a Sensor
> I think this wording is very problematic.

I would stand by that statement, that an actuator can be seen as “sensor with benefits” in VSS,  I do agree that notion might not be obvious, shared by anyone. It is a poor choice here, if I expect a `sensor-provider` to also “feed” actuators and potentially attributes -> not obvious.

I was thinking about “signal-provider”, but then the question is, is that not the same as the “provide anything” in authorisation, i.e. it does not capture on an architecture level that I want to express, that this is a component that does not changes anything in the vehicle.

I think existing, “VSS terms” are not helping here, and introduced state-provider and actuation-provider as the two different classes of providers

doc/terminology.md
SebastianSchildt: Guess this refers to the "advertising language" -> removed

doc/terminology.md
argerus: > I don’t think “value” is defined/clear cut term in VSS.

I agree with this when used as "Value provider". It is to generic, and sounds more like something out of the PR department.

> [...] it does not capture on an architecture level that I want to express, that this is a component that does not changes anything in the vehicle. 
> I think existing, “VSS terms” are not helping here, and introduced state-provider and actuation-provider as the two different classes of providers

I think "state provider" is basically as generic as "value provider" (and sounds like a state (government) provider).

How do you feel about "data provider"? That should make it clear that it's not changing anything in the vehicle, and it feels a lot more natural to speak of in this context. I think someone at FOSDEM even spontaneously referred to them as "data providers".

doc/terminology.md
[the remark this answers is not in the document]
argerus: > I think it is ok, becasue as “consumer” of my power drill I am also free to make holes.

Haha :)
Yes, I think consumer gets the point across as well.

doc/terminology.md
```diff
@@ -0,0 +1,83 @@
...
+For specific examples of components fullfilling these roles also check [System Architecture](./system-architecture.md).
+**todo: probably system architecture not a good name for that file, and I did not update it with the terms in this doc**
+
+## Clients: VSS Consumers
+ VSS Consumers can also be seen as northbound clients of the VSS ressource server. A VSS consumer intends to interact with the Vehicle represented by the VSS model served by the server: A consumer typically wants to read sensors representing vehicle state available in VSS or affecting changes in the vehicle hardware by interacting with actuators available in VSS.
+
+ So typically A VSS consumer might be "apps" running on vehicle computers, the infotainment, or maybe even a user's personal device. 
+
+
+ ## Clients: VSS Providers
```
erikbosch, line 36: Is this supposed to be a heading - if so remove first blank

doc/terminology.md
erikbosch: I am quite happy with something like "data provider" and "actuation provider". I see a value in not using the term "sensor" as it is not always a "real sensor", it could for instance be a calculated/estimated value.

doc/terminology.md
SebastianSchildt: Done

doc/terminology.md
SebastianSchildt: Changed to data-provider now

SebastianSchildt:
I changed `state-provider`  to `data-provider` and integrated all other comments. Also fixed a number of typos. Content-wise  I would consider this ready now

doc/terminology.md
```diff
@@ -0,0 +1,106 @@
+# Terminology
+
+This pages gives an overview about the terms we use, when talking about interacting with VSS signals in-vehicle using KUKSA.val components.
```
argerus, line 3: This document aims to describe all the terminology used in KUKSA.VAL, not just the terminology related to VSS (although that will be a large part of it).

argerus (changes requested):

doc/terminology.md
```diff
@@ -0,0 +1,106 @@
...
+   * [Value](#value)
+   * [Metadata](#metadata)
+   * [Overlay](#overlay)
+   * [Datapoint](#datapoint)
+# VSS in-vehicle: Main components
+
+![In-vehicle VSS components](./pictures/terminology.svg)
+
+KUKSA.val helps you providing and interacting with VSS data in-vehicle. The figure above depicts the required components. We will describe each of their roles in the following sections. 
+## VSS Ressource Server
```
argerus, line 26:
```suggestion
# KUKSA.VAL Components

![In-vehicle VSS components](./pictures/terminology.svg)

KUKSA.val helps you provide and interact with VSS data in-vehicle. The figure above depicts the required components. We will describe each of their roles in the following sections. 
## Server
```

Just as we specify _Client_ below and go on to describe the different types etc, I think _Server_ is a better fit here (i.e. as top level heading). It could be good to specify what a "resource server" is also, especially pointing out the definition of it in OAuth2. I could also make sense to add another section for the authorization related terms, just like the _Vehicle Signal Specification_ section.

SebastianSchildt: I suggest we might decide to add the OAuth definitions in a later PR, if it seems to make sense.

doc/terminology.md
```diff
@@ -0,0 +1,110 @@
...
+ VSS introduces a domain taxonomy for vehicle signals. It defines
+ * A syntax for defining vehicle signals in a structured manner.
+ * A catalog of standard signals (VSs standard catalogue) related to vehicles.
+
+It can be used as standard in automotive applications to communicate information around the vehicle, which is semantically well defined. 
+
+See the [the official VSS documentation](https://covesa.github.io/vehicle_signal_specification/) for all details. Here we will summarizes the terms and concepts most relevant for KUKSA.
+
+## Signal
+Everything we handle in KUKSA is defined as a using VSS. A signal in VSS is an entity identified by a dot-separated string in VSS, i.e. `Vehicle.Speed`.
```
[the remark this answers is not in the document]
SebastianSchildt: I removed that sentence.Does not "break" the introduction for a VSS signal, and does not mix KUKSA into pure VSS defitniions

PR #480, opened 2023-02-14
Add ISC & ring licenses

argerus:
* Update createbom to handle custom license files
* Add ISC & ring licenses


SebastianSchildt (changes requested):
Every entry in thirdparty_components now has a `license_file": null` entry, can we either supress that, or point it to the license we copied, i.e. `./something.txt.gz`

(I have only tested for "standard licenses" so far, not yet with ring, but again, if there is license_file in there, make sure it points to whatever we copied in

SebastianSchildt (approved):
Looking good now.
I noticed license for ring in JSON being "none" but It hink that is fine (as it is not declared for the crate, and I am not sure what else to put, if they don't provide a good SPDX string we should not make one up). The license_file points to the correct place

Also still detected when I tried sneaking in a not-whitelisted license.

Will merge after CI is green, unfortunately it seems Eclipse org runners are overcommitted again today

SebastianSchildt (approved):
lgtm

PR #481, opened 2023-02-14
Change expiration of ttl.sh images

argerus:
Unavailability of build runners will sometimes cause the ttl.sh images to expire before the integration test has a chance to run them.


SebastianSchildt (approved):
Normally I'd say we should be nice citizens to ttl (it is free), so we are with the service a little while longer. But since databroker containers are so small, should not hurt (default now  is 24h).
I will merge the ISC license add PR  (including this), so only one build is triggered :D Not wasting all our minutes in the morning...

argerus:
Closing as it's already merged as part of #480

Issue #489, opened 2023-02-24
Facing kuala-grpc error with statusCode: UNIMPLEMENTED


mhatrepratik11:
**Hi Team,**
**I am facing a issue while connecting to kuksa-val-server using kuksa-client with grpc protocol.**

I have kuksa-val-server up and running and exposed ports please see in the below screenshots:

<img width="570" alt="Kuksa-Val-Server is running" src="https://user-images.githubusercontent.com/10961872/221299894-67f214b8-f85d-41d5-ae8b-0d44dab83a84.png">

See the exposed ports in the below screenshot:
<img width="1392" alt="Port exposed and forwarded" src="https://user-images.githubusercontent.com/10961872/221304371-2af17a1b-8945-45ef-a891-03506a1e8f2a.png">

I am trying to connect to kuala-val-server using kuala-client using VSSClient as you can see in the below screenshot.

**While trying to connect with python code:** 

```
from kuksa_client.grpc import VSSClient

with VSSClient('localhost', 50051) as client:
    current_values = client.get_current_values([
        'Vehicle.Speed',
    ])
    print(current_values['Vehicle.Speed'].value)
```

**_I am getting error with grpc statusCode: 12 (UNIMPLEMENTED) as shown below in the screenshot:
kuksa_client.grpc.VSSClientError: ({'code': 12, 'reason': 'unimplemented', 'message': ''}, [])_**

<img width="570" alt="Kuksa-grpc-unimpleted-error" src="https://user-images.githubusercontent.com/10961872/221301241-db85050a-52bc-4220-bfae-001da145e749.png">

**_NOTE: I am able to connect to kuksa-val-server through “ws” protocol by using code shown in the screenshot below:_**

<img width="570" alt="Kuksa-Client-WS-Connected" src="https://user-images.githubusercontent.com/10961872/221301868-d29f5563-fbbd-4e77-af13-c2fcc6855dad.png">

**Thanks in Advance !**

SebastianSchildt:
Hi, 

i understand this is a bit confusing currently. Basically  there are not all combinations between client and server that work.  Maybe this helps


| Server  | Client  |  Py API  | Comment
|--------|---|---|---|
[...]

PR #494, opened 2023-03-01
Describe Docker Desktop workarounds for Windows and Mac OS

SebastianSchildt:
Add hints how to use Docker Desktop.

This came up in #489 , so let's preserve the knowledge

### DIFF
--- a/doc/quickstart.md
+++ b/doc/quickstart.md
@@ -69,7 +69,7 @@ gRPC channel disconnected.
 
 ## Reading and Writing VSS data with code
 
-Most likely, to get productive with KUKSA.val you want to interact with it programmatically. Here is you you do it with our Python library.
+To realize your ideas with KUKSA.val you need to write programs that interact with its API. The easiest way to achieve this is using our Python library.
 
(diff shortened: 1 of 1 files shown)


doc/quickstart.md
```diff
@@ -69,7 +69,7 @@ gRPC channel disconnected.
 
 ## Reading and Writing VSS data with code
 
-Most likely, to get productive with KUKSA.val you want to interact with it programatically. Here is you you do it with our Python library.
+Most likely, to get productive with KUKSA.val you want to interact with it programmatically. Here is you you do it with our Python library.
```
lukasmittag, line 72:
```suggestion
Most likely you want to get productive with KUKSA.val in a developing way. The approach would then be to use our Python library.
```

lukasmittag (changes requested):

doc/quickstart.md
SebastianSchildt: I tried rewording this

lukasmittag (approved):

PR #497, opened 2023-03-03
Add missing mapping of errors

### LINKED ISSUE #496: Updating current value using wrong type
Given that I have registered a Data Entry of, say, type uint16.
Am I right in assuming that the Databroker should reject a gRPC request to set the current value to a boolean value?

sophokles73:
The layer that handles "set" requests has been extended with code that maps (internal) UpdateErrors to DataEntryErrors that are defined by the kuksa.val.v1 proto files.

Fixes #496


sophokles73:
@argerus is there any particular reason why the GitHub workflows do not run for this PR?

SebastianSchildt:
This are Github shared runners, Eclipse has booked none, or far to less. Wait a night. Will be queued a few ours.

For some older workflows that could not even finish in the 6 hour max runtime,  we have (faster) self-hosted runners, but they have their own issues in setting up, and need to be added on a per-repo basis. We might look into this again, if/once we move to our own orga (then we can have orga runners, not possible in eclipse orga).

But for now, waiting is the only options, there is just a serious shortage on runner minutes for Eclipse projects, which is why latest at lunch you get queued a long time

argerus:
> @argerus is there any particular reason why the GitHub workflows do not run for this PR?

No, but we've noticed that that seems to happen from time to time in the eclipse organization (?).

argerus:
Anyway, the patch looks good to me, and I've tried it locally. So I'm just waiting for the CI.

SebastianSchildt:
if you are happy @argerus , please give an "approve". In that case me or @lukasmittag  would "auto-merge" once CI is happy as well

argerus (approved):

argerus (changes requested):
Actually, it seems `cargo fmt` does change a few lines, so you should run that and update the PR first.

SebastianSchildt:
CLI might have been a transient error, I am rerunning the job

argerus (approved):

PR #501, opened 2023-03-08
Parse scope and convert to permissions

argerus:
* The scope is parsed and merged into `Permissions`.
* `Permissions` are available in all requests.


lukasmittag (approved):
Looks good. Codewise nothing I can see what is wrong behaviour.

PR #502, opened 2023-03-08
Enable enforcement of allowed values

lukasmittag:
This adds another validation step before one can feed/set a path.

E.g. If you want to allow only a few options for setting you can do something like:
Vehicle.Body.Windshield.Front.Wiping.Mode - only allows: "OFF", "SLOW", "MEDIUM", "FAST", "INTERVAL", "RAIN_SENSOR"
Databroker now checks if the value you specified is in this list (if it is specified)

### DIFF
--- a/kuksa_databroker/databroker/src/broker.rs
+++ b/kuksa_databroker/databroker/src/broker.rs
@@ -173,6 +173,7 @@ impl Entry {
         //                     .entry_type
         //                     .data_type
         //                     .description
+        //                     .allowed
 
         update
     }
@@ -1187,7 +1188,7 @@ async fn test_register_datapoint() {
             ChangeType::OnChange,
             EntryType::Sensor,
             "Test datapoint 1".to_owned(),
-            None,
+            Some(DataValue::BoolArray(Vec::from([true]))),
         )
         .await
         .expect("Register datapoint should succeed");
@@ -1199,6 +1200,10 @@ async fn test_register_datapoint() {
                 assert_eq!(entry.metadata.path, "test.datapoint1");
                 assert_eq!(entry.metadata.data_type, DataType::Bool);
                 assert_eq!(entry.metadata.description, "Test datapoint 1");
+                assert_eq!(
+                    entry.metadata.allowed,
+                    Some(DataValue::BoolArray(Vec::from([true])))
+                );
             }
             None => {
                 panic!("datapoint should exist");
@@ -1225,6 +1230,7 @@ async fn test_register_datapoint() {
                 assert_eq!(entry.metadata.path, "test.datapoint2");
                 assert_eq!(entry.metadata.data_type, DataType::String);
                 assert_eq!(entry.metadata.description, "Test datapoint 2");
+                assert_eq!(entry.metadata.allowed, None);
             }
             None => {
                 panic!("no metadata returned");
@@ -1245,6 +1251,23 @@ async fn test_register_datapoint() {
         .expect("Register datapoint should succeed");
 
     assert_eq!(id3, id1);
+
+    if datapoints
+        .add_entry(
+            "test.datapoint3".to_owned(),
+            DataType::String,
+            ChangeType::OnChange,
+            EntryType::Sensor,
+            "Test datapoint 3".to_owned(),
+            Some(DataValue::Int32Array(Vec::from([1, 2, 3, 4]))),
+        )
+        .await
+        .is_ok()
+    {
+        panic!("Datapoint should not be regsitered successfully");
+    } else {
+        // everything fine, should not succed to register because allowed is Int32Array and data_type is String
+    }
 }
 
 #[tokio::test]
@@ -1258,7 +1281,7 @@ async fn test_get_set_datapoint() {
             ChangeType::OnChange,
             EntryType::Sensor,
             "Test datapoint 1".to_owned(),
-            None,
+            Some(DataValue::Int32Array(vec![100])),
         )
         .await
         .expect("Register datapoint should succeed");
@@ -1270,6 +1293,7 @@ async fn test_get_set_datapoint() {
             ChangeType::OnChange,
             EntryType::Actuator,
             "Test datapoint 2".to_owned(),
+            None,
         )
         .await
         .expect("Register datapoint should succeed");
@@ -1309,6 +1333,7 @@ async fn test_get_set_datapoint() {
                 entry_type: None,
                 data_type: None,
                 description: None,
+                allowed: Some(Some(DataValue::Int32Array(vec![100]))),
             },
         )])
         .await
@@ -1338,7 +1363,7 @@ async fn test_get_set_datapoint() {
                 entry_type: None,
                 data_type: None,
                 description: None,
-                allowed: None,
+                allowed: Some(Some(DataValue::Int32Array(vec![100]))),
             },
         )])
         .await
@@ -1358,6 +1383,7 @@ async fn test_get_set_datapoint() {
                 entry_type: None,
                 data_type: None,
                 description: None,
+                allowed: None,
             },
         )])
         .await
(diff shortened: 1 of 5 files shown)


kuksa_databroker/databroker/src/types.rs
```diff
@@ -390,6 +390,20 @@ impl DataValue {
         }
     }
 
+    pub fn find(&self, search_value: &DataValue) -> Result<bool, CastError> {
```
argerus, line 393: And I think `contains` would be a better name for this function (`find` is usually used when the found element or it's index is returned by the function).

But since the functionality is only applicable to `DataValue`s that are arrays it could make more sense to not have this as a general function for `DataValue`s at all, but instead move the functionality directly into `validate_allowed` (as that's probably the only place it will be used anyway).

If keeping it as a generic function, a new error type should probably be created with `WrongType` and `NotAnArray` as possible error kinds, since CastError isn't that fitting (I think).

argerus (changes requested):

kuksa_databroker/databroker/src/broker.rs
```diff
@@ -175,14 +177,29 @@ impl Entry {
...
+            self.validate_allowed(value)?;
         }
         Ok(())
     }
 
-    fn validate_value(&self, value: &DataValue) -> Result<(), UpdateError> {
+    fn validate_allowed(&self, value: &DataValue) -> Result<(), UpdateError> {
+        // check if allowed value
+        if let Some(allowed_values) =  &self.metadata.allowed {
+            match allowed_values.find(value){
```
argerus, line 192: If the `value` is of array type (i.e. any of the `DataValue::*Array`) the correct behaviour is to check whether each element of the array is present in `allowed`. See https://covesa.github.io/vehicle_signal_specification/rule_set/data_entry/allowed/#allowed-values-for-array-types.

Hence the validation check needs to account for this.

argerus (changes requested):

kuksa_databroker/databroker/src/broker.rs
```diff
@@ -696,6 +713,7 @@ impl Entries {
         entry_type: EntryType,
         description: String,
         datapoint: Option<Datapoint>,
+        allowed: Option<types::DataValue>
```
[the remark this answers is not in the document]
lukasmittag: this needs to be validated in 'update_entries' too I guess.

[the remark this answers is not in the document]
argerus: Rather in `Entries::update(...)`.

```
impl Entries {
  pub fn update(...)
  pub fn add(...)
}
```

as those two ^ functions are the only way to update the entries in `Entries` (as the fields are private).

All other functions do it through those functions, of which `update_entries` is one.

kuksa_databroker/databroker/src/types.rs
argerus: `find` can be removed as it's no longer used.

kuksa_databroker/databroker/src/broker.rs
```diff
@@ -708,6 +880,28 @@ impl Entries {
...
+                (DataValue::Int32Array(_allowed_values), DataType::Int32Array) => (),
+                (DataValue::Int64Array(_allowed_values), DataType::Int64Array) => (),
+                (DataValue::Uint32Array(_allowed_values), DataType::Uint32Array) => (),
+                (DataValue::Uint64Array(_allowed_values), DataType::Uint64Array) => (),
+                (DataValue::FloatArray(_allowed_values), DataType::FloatArray) => (),
+                (DataValue::DoubleArray(_allowed_values), DataType::DoubleArray) => (),
+                _ => return Err(RegistrationError {}),
+            }
+        }
+
```
argerus, line 904: This looks very much like the `fn validate_allowed_type(..)` above.
You can probably reuse it by mapping the errors, i.e.
```
validate_allowed_type(..).map_err(|err| RegistrationsError {})?;
```

Maybe a `ValidationError` should be introduced since the errors are not only for updates anymore.

argerus (changes requested):

kuksa_databroker/databroker/src/broker.rs
lukasmittag: This then should be moved out of the implementation of Entry. Into DataValue makes not much sense in my opinion, because it is a corner case for a DataValue. Any suggestions where it could fit or just standalone?

argerus: > This then should be moved out of the implementation of Entry.

Why?

What I mean is that it makes sense to reuse validation logic regardless of if we're adding an entry or updating it. It should be validated the same, right?

lukasmittag: Right. Then the mechanism to update needs to be altered a bit. Okay, I see a way to do it. Thanks for the quick reply!

argerus (changes requested):

kuksa_databroker/databroker/src/broker.rs
```diff
@@ -708,28 +880,32 @@ impl Entries {
         // Map name -> id
         self.path_to_id.insert(name.clone(), id);
```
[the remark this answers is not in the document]
lukasmittag: Should maybe really investigate some time to firstly move validation one level deeper (e.g. to ```Metadata``` and ```Datapoint``` implementation). And maybe therefore have a look what to do with ```id``` trait of ```Metadata```

kuksa_databroker/databroker/src/broker.rs
```diff
@@ -374,6 +540,10 @@ impl Entry {
             changed.insert(Field::ActuatorTarget);
         }
 
+        if update.allowed != self.metadata.allowed {
+            self.metadata.allowed = update.allowed;
+        }
```
argerus, line 545: Alright testing the actual functionality now and noticed that enforcing allowed values only works the first time.

The reason is the code added here.

It is changing `metadata.allowed` to `None` when the update doesn't contain any `allowed` (i.e. the common case). All subsequent requests are then allowed regardless of value since `allowed` is now set to `None`.

argerus: When #507 is merged, this would actually have been prevented anyway, since changing metadata isn't allowed. But it's still a bug.

So one way to fix it is just removing this code for the time being, until we add support for updating metadata (and not just allowed values).

argerus: This is handled for `actuator_target` by wrapping it in an additional `Option<>`

```
    // Actuator target is wrapped in an additional Option<> in
    // order to be able to convey "update it to None" which would
    // mean setting it to `Some(None)`.
    pub actuator_target: Option<Option<Datapoint>>, // only for actuators
```

kuksa_databroker/databroker/src/broker.rs
```diff
@@ -696,40 +874,49 @@ impl Entries {
...
+            .validate_allowed_type(&new_entry.metadata.allowed)
+            .map_err(|_err| RegistrationError {})?;
+
         // Get next id (and bump it)
         let id = self.next_id.fetch_add(1, Ordering::SeqCst);
 
         // Map name -> id
         self.path_to_id.insert(name.clone(), id);
 
+        new_entry.metadata.id = id;
```
argerus, line 916: Doesn't pass clippy since this `clone()` isn't needed anymore..

clippy error / warning
```rust
error: redundant clone
   --> kuksa_databroker/databroker/src/broker.rs:914:36
    |
914 |         self.path_to_id.insert(name.clone(), id);
    |                                    ^^^^^^^^ help: remove this
    |
```

```suggestion

        // Get next id (and bump it)
        let id = self.next_id.fetch_add(1, Ordering::SeqCst);

        // Map name -> id
        self.path_to_id.insert(name, id);

        new_entry.metadata.id = id;
```

argerus (changes requested):

kuksa_databroker/databroker/src/broker.rs
argerus: Otherwise, I think this PR looks fine now.

argerus (approved):

argerus (changes requested):

kuksa_databroker/databroker/src/main.rs
```diff
@@ -310,6 +313,7 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
                                 entry_type: None,
                                 data_type: None,
                                 description: None,
+                                allowed: Some(None),
```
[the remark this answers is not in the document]
lukasmittag: Since, it would be the same for actuator target.I will create an issue for EntryUpdate::builder() issue/annotation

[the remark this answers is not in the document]
argerus: Sure. But just to confirm, you'll fix the bug regarding `allowed` in this PR right.

There is no corresponding bug for actuator_target, and adding a `builder()` isn't a requirement, that's just a way to make it a bit harder to get it wrong in the future.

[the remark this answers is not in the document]
lukasmittag: #515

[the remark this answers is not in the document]
argerus: I.e. this one, `allowed: Some(None)` on line 316 still seems to be there. It should not.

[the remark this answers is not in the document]
argerus: (on the other hand... it doesn't really matter since it's a hardcoded entry)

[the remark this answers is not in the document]
lukasmittag: Yeah, removed it.

argerus (approved):

PR #503, opened 2023-03-08
Catch exception for the client when incorrect types supplied as value.

### LINKED ISSUE #498: kuksa-client doesnt recover from exception
While working with the databroker and kuksa-client using gRPC protocol, if any interaction results in an exception, the client cannot be recovered. I need to quit the client and restart from scratch. 

To reproduce the issue:
1. Start the databroker: `docker run -it --rm --net=host ghcr.io/eclipse/kuksa.val/databroker:master`
2. Start the client: `docker run -it --rm --net=host ghcr.io/eclipse/kuksa.val/kuksa-client:master --port 55555 --protocol grpc --insecure`
3. Execute following command: `setValue Vehicle.Speed 0x100`. This results in an exception.
4. Any subsequent commands result in timeouts only.
exception

nayakned:
Handling ValueError exception in the event of incorrect values. Should fix #498

### DIFF
--- a/kuksa-client/kuksa_client/cli_backend/grpc.py
+++ b/kuksa-client/kuksa_client/cli_backend/grpc.py
@@ -200,6 +200,8 @@ class Backend(cli_backend.Backend):
                 responseQueue.put((resp, None))
             except kuksa_client.grpc.VSSClientError as exc:
                 responseQueue.put((None, exc.to_dict()))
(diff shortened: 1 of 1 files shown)


SebastianSchildt (approved):
Thank you, this works.
Approving, as the failed kuksa-client cli is just related to a buildx/containerd upstream issue

PR #504, opened 2023-03-09
Add Cucumber/Gherkin based integration tests

sophokles73:
I have implemented a few basic integration tests against a running Databroker.
The tests have been implemented using the [Cucumber framework](https://cucumber.io/)
which allows specifying test scenarios using natural language following
[Gherkin](https://cucumber.io/docs/gherkin/) syntax.

The general idea is to implement basic functionality for interacting with the
system under test (the Databroker) and asserting outcomes. Based on this
functionality, one can then easily add more tests by simply writing additional
test scenarios.


SebastianSchildt:
Cucumber looks interesting, though I never  looked into it any deeper. Could you maybe present what you did (and what would need to be done to extend the scope of this tests) at the next KUKSA community call (Thursday 16th 13 CET, see here https://eclipse.github.io/kuksa.website/resources/ for link)?

kuksa_databroker/createbom/bomutil/maplicensefile.py
```diff
@@ -21,6 +21,7 @@
 
 MAP = {
     "Apache-2.0": "Apache-2.0.txt.gz",
+    "BlueOak-1.0.0": "BlueOak-1.0.0.md.gz",
```
SebastianSchildt, line 24: Where is this license coming from? I see the cucumber create added below, but that is just APACHE2 or MIT according to crates.io . Is this from some transitive dependency? 

In case you know, is this a dependency that is used only during build time or is it a runtime dependency? Or only at runtime for tests?

It "looks" harmless (https://spdx.org/licenses/BlueOak-1.0.0.html) but in any case, we need to clear it with Eclipse. But first we need to know where it comes from. If you can tell us @sophokles73 , maybe @bs-jokri  can try pushing it through dash, to see whether it is maybe already cleared

sophokles73: Yes, it is introduced by a transitive dependency of cucumber. As such, it is only required for building/testing but not during runtime.

SebastianSchildt:
It seems like a pretty cool concept to me. Any opinions by other guys? I already "got the tour" from @sophokles73 , I think he is happy to show anyone else as well. 

Speaking of this PR specifically, if we want to move forward with this, the main two open points are
 - getting the new license/dependency cleared (I can work on that)
 - an opinion, maybe from @argerus w.r.t to the changes done in this PR to Cargo and the "invasiveness" of the code in databroker. Looks not bad to me. Feels fine to ne, but you have a better overview. Obviously, when hacking databroker, tests might fail, or when we change som design or apsect of the API, tests need to be adapted, but that is sort of the point :)

SebastianSchildt:
Ping @argerus: Should we move forward with this? It has been open quite some time (and meanwhile needs rebasing). We should decide, and if we want it, rebase & merge.

I "like" it, but that is not a very technical argument. Any opinions?

argerus:
> Should we move forward with this?

I thought we where waiting for the license approval (from Eclipse), since this PR adds a license which we haven't whitelisted yet...

SebastianSchildt:
Ah good point, I need to follow up on that. Assuming it is fine, we want this?

argerus:
Yes, I've postponed reviewing it in detail (given that we're waiting for license approval).
But in general it looks good, and it's definitely something that would benefit the project when merged.

sophokles73:
@argerus @SebastianSchildt I have rebased to current HEAD ...

SebastianSchildt:
license-wise I think we can merge this. BlueOak "looks fine to me", and we have a (relatively fresh) CQ open Eclipse side. This is "only" a test dependency, and worst case can be pulled. We have some CQs from other dependencies already "in" open for > 1 month now. So let's get this up to date and merged based on our best understanding, and react on anything that might come out of the CQs

sophokles73:
Then I can rebase again (one last time?) ;-)

argerus (approved):

SebastianSchildt (approved):

PR #510, opened 2023-03-16
Test cross-compile base image

sophokles73:
This PR is just for testing a change to the Dockerfile and how it works on GitHub Actions. No need to review nor merge.


sophokles73:
the approach tested here seems to work properly and I have now incorporated it into #506

Issue #406, opened 2022-11-16
Do not have "magic" default tokens in Python library 


SebastianSchildt:
See also discussion in #402 and #371

The python library sometimes tries to guess a usable token file with lines like

```python
self.tokenfile = config.get('token', str(self.default_cert_path / 'jwt/all-read-write.json.token'))
```

That is an uglyiness we inherited from the previous version of the client. This is wrong, because it is unexpected behaviour.

It is different from shipping default certs/and tokens for testing, that is fine, because it should be transparent when and where they are used, e.g. if we modify our clients that in their _configuration_ file it refers to a default token, that is fine, I just think we should not have code "deep in the lib" that goes like "hmm, no token given, let's use some secret path instead then"

@romainletendart i guess up to you cleaning the lib - or having opinions wrt to this

erikbosch:
This applies to `*.pem` / `*.key` files used for TLS-communication as well, like in the example below, or? I.e. we shall not make any assumption on where they are stored or what they are called?

https://github.com/eclipse/kuksa.val/blob/master/kuksa-client/kuksa_client/cli_backend/__init__.py#L31

```
        self.default_cert_path = pathlib.Path(kuksa_certificates.__path__[0])
        self.cacertificate = config.get('cacertificate', str(self.default_cert_path / 'CA.pem'))
        self.certificate = config.get('certificate', str(self.default_cert_path / 'Client.pem'))
        self.keyfile = config.get('key', str(self.default_cert_path / 'Client.key'))
```

SebastianSchildt:
Not sure  is exactly the same, or depends on usage. Assuming there is a "default path" for certs,  I think it might be ok 
 - for the CLI client (not necessarily the lib) to load default certificates unless otherwise instructed
 - for the library part it might still be ok to load certificates for checking TLS server cert from a default location, if not overiden (could be discussed, but for practical purposes I think there will not be many installations where the endpoint is really signed by an "official" system CA). But might want to print a waring "Using builtin default CA" or something like that 

A little like when your browser/app communicates with facebook or tesla, etc.: Of course things required for the TLS part can be checked without any additional config, but of course  there will be no JWT to post crap or unlock your car, until provided by the user/requested via AP before

So I think the JSON token as mentioned above is the critical thing to get out here, but we should take a good look at the other things as well and think clearly what is needed as default here and how it can be made obvious to the user..

In the example I am not sure what applies to which i.e. is Client.key even used at all currently?


[...]

PR #517, opened 2023-03-23
Support authorization in kuksa client/library

lukasmittag:
Implement authorization for kuksa python library and use it in the kuksa client. It includes a cli command too.

### DIFF
--- a/kuksa-client/kuksa_client/grpc/__init__.py
+++ b/kuksa-client/kuksa_client/grpc/__init__.py
@@ -44,67 +44,67 @@ logger = logging.getLogger(__name__)
 
 
 class DataType(enum.IntEnum):
-    UNSPECIFIED     = types_pb2.DATA_TYPE_UNSPECIFIED
-    STRING          = types_pb2.DATA_TYPE_STRING
-    BOOLEAN         = types_pb2.DATA_TYPE_BOOLEAN
-    INT8            = types_pb2.DATA_TYPE_INT8
-    INT16           = types_pb2.DATA_TYPE_INT16
-    INT32           = types_pb2.DATA_TYPE_INT32
-    INT64           = types_pb2.DATA_TYPE_INT64
-    UINT8           = types_pb2.DATA_TYPE_UINT8
-    UINT16          = types_pb2.DATA_TYPE_UINT16
-    UINT32          = types_pb2.DATA_TYPE_UINT32
-    UINT64          = types_pb2.DATA_TYPE_UINT64
-    FLOAT           = types_pb2.DATA_TYPE_FLOAT
-    DOUBLE          = types_pb2.DATA_TYPE_DOUBLE
-    TIMESTAMP       = types_pb2.DATA_TYPE_TIMESTAMP
-    STRING_ARRAY    = types_pb2.DATA_TYPE_STRING_ARRAY
-    BOOLEAN_ARRAY   = types_pb2.DATA_TYPE_BOOLEAN_ARRAY
-    INT8_ARRAY      = types_pb2.DATA_TYPE_INT8_ARRAY
-    INT16_ARRAY     = types_pb2.DATA_TYPE_INT16_ARRAY
-    INT32_ARRAY     = types_pb2.DATA_TYPE_INT32_ARRAY
-    INT64_ARRAY     = types_pb2.DATA_TYPE_INT64_ARRAY
-    UINT8_ARRAY     = types_pb2.DATA_TYPE_UINT8_ARRAY
-    UINT16_ARRAY    = types_pb2.DATA_TYPE_UINT16_ARRAY
-    UINT32_ARRAY    = types_pb2.DATA_TYPE_UINT32_ARRAY
-    UINT64_ARRAY    = types_pb2.DATA_TYPE_UINT64_ARRAY
-    FLOAT_ARRAY     = types_pb2.DATA_TYPE_FLOAT_ARRAY
-    DOUBLE_ARRAY    = types_pb2.DATA_TYPE_DOUBLE_ARRAY
+    UNSPECIFIED = types_pb2.DATA_TYPE_UNSPECIFIED
+    STRING = types_pb2.DATA_TYPE_STRING
+    BOOLEAN = types_pb2.DATA_TYPE_BOOLEAN
+    INT8 = types_pb2.DATA_TYPE_INT8
+    INT16 = types_pb2.DATA_TYPE_INT16
+    INT32 = types_pb2.DATA_TYPE_INT32
+    INT64 = types_pb2.DATA_TYPE_INT64
+    UINT8 = types_pb2.DATA_TYPE_UINT8
+    UINT16 = types_pb2.DATA_TYPE_UINT16
+    UINT32 = types_pb2.DATA_TYPE_UINT32
+    UINT64 = types_pb2.DATA_TYPE_UINT64
+    FLOAT = types_pb2.DATA_TYPE_FLOAT
+    DOUBLE = types_pb2.DATA_TYPE_DOUBLE
+    TIMESTAMP = types_pb2.DATA_TYPE_TIMESTAMP
+    STRING_ARRAY = types_pb2.DATA_TYPE_STRING_ARRAY
+    BOOLEAN_ARRAY = types_pb2.DATA_TYPE_BOOLEAN_ARRAY
+    INT8_ARRAY = types_pb2.DATA_TYPE_INT8_ARRAY
+    INT16_ARRAY = types_pb2.DATA_TYPE_INT16_ARRAY
+    INT32_ARRAY = types_pb2.DATA_TYPE_INT32_ARRAY
+    INT64_ARRAY = types_pb2.DATA_TYPE_INT64_ARRAY
+    UINT8_ARRAY = types_pb2.DATA_TYPE_UINT8_ARRAY
+    UINT16_ARRAY = types_pb2.DATA_TYPE_UINT16_ARRAY
+    UINT32_ARRAY = types_pb2.DATA_TYPE_UINT32_ARRAY
+    UINT64_ARRAY = types_pb2.DATA_TYPE_UINT64_ARRAY
+    FLOAT_ARRAY = types_pb2.DATA_TYPE_FLOAT_ARRAY
+    DOUBLE_ARRAY = types_pb2.DATA_TYPE_DOUBLE_ARRAY
     TIMESTAMP_ARRAY = types_pb2.DATA_TYPE_TIMESTAMP_ARRAY
 
 
 class EntryType(enum.IntEnum):
     UNSPECIFIED = types_pb2.ENTRY_TYPE_UNSPECIFIED
-    ATTRIBUTE   = types_pb2.ENTRY_TYPE_ATTRIBUTE
-    SENSOR      = types_pb2.ENTRY_TYPE_SENSOR
-    ACTUATOR    = types_pb2.ENTRY_TYPE_ACTUATOR
+    ATTRIBUTE = types_pb2.ENTRY_TYPE_ATTRIBUTE
+    SENSOR = types_pb2.ENTRY_TYPE_SENSOR
+    ACTUATOR = types_pb2.ENTRY_TYPE_ACTUATOR
 
 
 class View(enum.IntEnum):
-    UNSPECIFIED   = types_pb2.VIEW_UNSPECIFIED
+    UNSPECIFIED = types_pb2.VIEW_UNSPECIFIED
     CURRENT_VALUE = types_pb2.VIEW_CURRENT_VALUE
-    TARGET_VALUE  = types_pb2.VIEW_TARGET_VALUE
-    METADATA      = types_pb2.VIEW_METADATA
-    FIELDS        = types_pb2.VIEW_FIELDS
-    ALL           = types_pb2.VIEW_ALL
+    TARGET_VALUE = types_pb2.VIEW_TARGET_VALUE
+    METADATA = types_pb2.VIEW_METADATA
+    FIELDS = types_pb2.VIEW_FIELDS
+    ALL = types_pb2.VIEW_ALL
 
 
 class Field(enum.IntEnum):
-    UNSPECIFIED                = types_pb2.FIELD_UNSPECIFIED
-    PATH                       = types_pb2.FIELD_PATH
-    VALUE                      = types_pb2.FIELD_VALUE
-    ACTUATOR_TARGET            = types_pb2.FIELD_ACTUATOR_TARGET
-    METADATA                   = types_pb2.FIELD_METADATA
-    METADATA_DATA_TYPE         = types_pb2.FIELD_METADATA_DATA_TYPE
-    METADATA_DESCRIPTION       = types_pb2.FIELD_METADATA_DESCRIPTION
-    METADATA_ENTRY_TYPE        = types_pb2.FIELD_METADATA_ENTRY_TYPE
-    METADATA_COMMENT           = types_pb2.FIELD_METADATA_COMMENT
-    METADATA_DEPRECATION       = types_pb2.FIELD_METADATA_DEPRECATION
-    METADATA_UNIT              = types_pb2.FIELD_METADATA_UNIT
+    UNSPECIFIED = types_pb2.FIELD_UNSPECIFIED
+    PATH = types_pb2.FIELD_PATH
+    VALUE = types_pb2.FIELD_VALUE
+    ACTUATOR_TARGET = types_pb2.FIELD_ACTUATOR_TARGET
+    METADATA = types_pb2.FIELD_METADATA
+    METADATA_DATA_TYPE = types_pb2.FIELD_METADATA_DATA_TYPE
+    METADATA_DESCRIPTION = types_pb2.FIELD_METADATA_DESCRIPTION
+    METADATA_ENTRY_TYPE = types_pb2.FIELD_METADATA_ENTRY_TYPE
+    METADATA_COMMENT = types_pb2.FIELD_METADATA_COMMENT
+    METADATA_DEPRECATION = types_pb2.FIELD_METADATA_DEPRECATION
+    METADATA_UNIT = types_pb2.FIELD_METADATA_UNIT
     METADATA_VALUE_RESTRICTION = types_pb2.FIELD_METADATA_VALUE_RESTRICTION
-    METADATA_ACTUATOR          = types_pb2.FIELD_METADATA_ACTUATOR
-    METADATA_SENSOR            = types_pb2.FIELD_METADATA_SENSOR
-    METADATA_ATTRIBUTE         = types_pb2.FIELD_METADATA_ATTRIBUTE
+    METADATA_ACTUATOR = types_pb2.FIELD_METADATA_ACTUATOR
+    METADATA_SENSOR = types_pb2.FIELD_METADATA_SENSOR
+    METADATA_ATTRIBUTE = types_pb2.FIELD_METADATA_ATTRIBUTE
 
 
 class MetadataField(enum.Enum):
@@ -129,7 +129,7 @@ class VSSClientError(Exception):
 
     @classmethod
     def from_grpc_error(cls, error: RpcError):
-        grpc_code, grpc_reason  = error.code().value
+        grpc_code, grpc_reason = error.code().value
         # TODO: Maybe details could hold an actual Error and/or repeated DataEntryError protobuf messages.
         #       This would allow 'code' to be an actual HTTP/VISS status code not a gRPC one.
         return cls(error={'code': grpc_code, 'reason': grpc_reason, 'message': error.details()}, errors=[])
@@ -158,23 +158,28 @@ class Metadata:
 
     @classmethod
     def from_message(cls, message: types_pb2.Metadata):
-        metadata = cls(data_type=DataType(message.data_type), entry_type=EntryType(message.entry_type))
+        metadata = cls(data_type=DataType(message.data_type),
+                       entry_type=EntryType(message.entry_type))
         for field in ('description', 'comment', 'deprecation', 'unit'):
             if message.HasField(field):
                 setattr(metadata, field, getattr(message, field))
         if message.HasField('value_restriction'):
-            value_restriction = getattr(message.value_restriction, message.value_restriction.WhichOneof('type'))
+            value_restriction = getattr(
+                message.value_restriction, message.value_restriction.WhichOneof('type'))
             metadata.value_restriction = ValueRestriction()
             for field in ('min', 'max'):
                 if value_restriction.HasField(field):
-                    setattr(metadata.value_restriction, field, getattr(value_restriction, field))
+                    setattr(metadata.value_restriction, field,
+                            getattr(value_restriction, field))
             if value_restriction.allowed_values:
-                metadata.value_restriction.allowed_values = list(value_restriction.allowed_values)
+                metadata.value_restriction.allowed_values = list(
+                    value_restriction.allowed_values)
         return metadata
 
     # pylint: disable=too-many-branches
     def to_message(self, value_type: DataType = DataType.UNSPECIFIED) -> types_pb2.Metadata:
-        message = types_pb2.Metadata(data_type=self.data_type.value, entry_type=self.entry_type.value)
+        message = types_pb2.Metadata(
+            data_type=self.data_type.value, entry_type=self.entry_type.value)
         for field in ('description', 'comment', 'deprecation', 'unit'):
             field_value = getattr(self, field, None)
             if field_value is not None:
@@ -191,12 +196,15 @@ class Metadata:
                 DataType.INT64_ARRAY,
             ):
                 if self.value_restriction.min is not None:
-                    message.value_restriction.signed.min = int(self.value_restriction.min)
+                    message.value_restriction.signed.min = int(
+                        self.value_restriction.min)
                 if self.value_restriction.max is not None:
-                    message.value_restriction.signed.max = int(self.value_restriction.max)
+                    message.value_restriction.signed.max = int(
+                        self.value_restriction.max)
                 if self.value_restriction.allowed_values is not None:
                     message.value_restriction.signed.allowed_values.extend(
-                        (int(value) for value in self.value_restriction.allowed_values),
+                        (int(value)
+                         for value in self.value_restriction.allowed_values),
                     )
             elif value_type in (
                 DataType.UINT8,
@@ -209,12 +217,15 @@ class Metadata:
                 DataType.UINT64_ARRAY,
             ):
                 if self.value_restriction.min is not None:
-                    message.value_restriction.unsigned.min = int(self.value_restriction.min)
+                    message.value_restriction.unsigned.min = int(
+                        self.value_restriction.min)
                 if self.value_restriction.max is not None:
-                    message.value_restriction.unsigned.max = int(self.value_restriction.max)
+                    message.value_restriction.unsigned.max = int(
+                        self.value_restriction.max)
                 if self.value_restriction.allowed_values is not None:
                     message.value_restriction.unsigned.allowed_values.extend(
-                        (int(value) for value in self.value_restriction.allowed_values),
+                        (int(value)
+                         for value in self.value_restriction.allowed_values),
                     )
             elif value_type in (
                 DataType.FLOAT,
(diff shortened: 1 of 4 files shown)


kuksa-client/kuksa_client/grpc/__init__.py
```diff
@@ -452,12 +452,14 @@ def __init__(
         self,
         host: str,
         port: int,
+        
         root_certificates: Optional[Path] = None,
         private_key: Optional[Path] = None,
         certificate_chain: Optional[Path] = None,
         *,
+        token: Optional[str] = None,
```
argerus, line 460: This doesn't seem to be used.

kuksa-client/kuksa_client/grpc/aio.py
```diff
@@ -251,19 +255,23 @@ async def set(self, *, updates: Collection[EntryUpdate], **rpc_kwargs) -> None:
...
         try:
             resp = await self.client_stub.Set(req, **rpc_kwargs)
         except AioRpcError as exc:
             raise VSSClientError.from_grpc_error(exc) from exc
         self._process_set_response(resp)
 
-    async def authorize(self, *, token: str) -> str:
-        raise NotImplementedError('"authorize" is not yet implemented')
+    async def authorize(self, *, token: Dict) -> None:
+        self.token = "Bearer " + token
```
argerus, line 274: Calling it `self.token` is confusing as it doesn't actually store the token.
It's fine to construct the authorization header here (instead of doing it repeatedly later), but in that case it should probably be called `self.authorization_header` or something like that instead.

kuksa-client/kuksa_client/grpc/aio.py
```diff
@@ -251,19 +255,23 @@ async def set(self, *, updates: Collection[EntryUpdate], **rpc_kwargs) -> None:
...
         paths_with_required_type = self._get_paths_with_required_type(updates)
+        
         paths_without_type = [
             path for path, data_type in paths_with_required_type.items() if data_type is DataType.UNSPECIFIED
         ]
         paths_with_required_type.update(await self.get_value_types(paths_without_type, **rpc_kwargs))
         req = self._prepare_set_request(updates, paths_with_required_type)
+        metadata = []
+        metadata.append(("authorization", self.token))
+        rpc_kwargs.__setitem__("metadata", metadata)
```
argerus, line 266: Instead of unconditionally adding/replacing the `metadata` argument, `"authorization"` should be added to it if exists, otherwise it should be created. Otherwise, other metadata will be overwritten if it exists.

argerus (changes requested):

kuksa-client/kuksa_client/grpc/aio.py
argerus: > But maybe then it would be an idea to create `authorization_header` in `authorize()`

That's the case already (except for the naming).

>  then just adding it to the rpc_kwargs if the authorization_header is not empty

That's also fine.

But the point here is that "metadata" shouldn't be overwritten if it exists. If it exists, then "authorization" should be added to it instead.

kuksa-client/kuksa_client/grpc/__init__.py
```diff
@@ -457,7 +457,7 @@ def __init__(
         certificate_chain: Optional[Path] = None,
         *,
         ensure_startup_connection: bool = True,
-    ):
+):
```
argerus, line 460: to much :(

Don't you have an automatic formatter in your IDE?

kuksa-client/kuksa_client/grpc/aio.py
```diff
@@ -251,19 +253,22 @@ async def set(self, *, updates: Collection[EntryUpdate], **rpc_kwargs) -> None:
...
             resp = await self.client_stub.Set(req, **rpc_kwargs)
         except AioRpcError as exc:
             raise VSSClientError.from_grpc_error(exc) from exc
         self._process_set_response(resp)
 
-    async def authorize(self, *, token: str) -> str:
-        raise NotImplementedError('"authorize" is not yet implemented')
+    async def authorize(self, *, token: Dict) -> None:
+        token = "Bearer " + token
+        self.authorization_header.append(("authorization", token))
```
argerus, line 271: This is not right.
It will keep appending more and more tuples to the list when authorize is called more than once.

kuksa-client/kuksa_client/grpc/aio.py
lukasmittag: You mean if something else is written in metadata it should not be overwritten only appended with the authorization header. Now I get it.

kuksa-client/kuksa_client/grpc/aio.py
```diff
@@ -55,6 +55,7 @@ def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
         self.channel = None
         self.exit_stack = contextlib.AsyncExitStack()
+        self.authorization_header = []
```
argerus, line 58: There will only be one authorization header (not a list).

argerus (changes requested):

erikbosch:
I tested it and it works. But before making a real PR we need to:

- Document the differences in authorization handling for server vs. broker, i.e. describe which tokens that are default for each app
- Decided which tokens we want to include in the python package and the Docker image. As of today we include only the "old" for server. Possibly we want the new ones as well.
- If we need different tokens, then likely default handling at https://github.com/eclipse/kuksa.val/blob/master/kuksa-client/kuksa_client/cli_backend/__init__.py#L35 needs to be changed. For instance first checking protocol, then setting default token file?

lukasmittag:
What I saw just now is. That I might move the authorize function to the BaseVSSClient, so asynchronous and synchronous kuksa python lib/client can use authorize fucntion and the parameters.

kuksa-client/kuksa_client/grpc/aio.py
```diff
@@ -274,6 +286,7 @@ async def subscribe(self, *,
             rpc_kwargs
                 grpc.*MultiCallable kwargs e.g. timeout, metadata, credentials.
         """
+        rpc_kwargs.__setitem__("metadata", self.rpc_metadata)
```
argerus, line 289: This is still overwriting the "metadata" key in `rpc_kwargs` just like before.

Maybe I need to expand on what I mean here.

Ok, so I don't know if you're familiar with the `*args` & `**kwargs` concept in python. Anyway, what the `*` and `**` means in practice is that this is a way to pack the arguments that are passed into a function in python into a single argument. A single `*` refers to packed positional arguments, and a double `**` refers to the packed keyword arguments. The same syntax is also used to unpack variables into function arguments when calling a function. But let's start with the packing.

In this case we have the following function exposed in the library:
```
async def subscribe(self, *, entries: Iterable[SubscribeEntry], **rpc_kwargs)
```

What this means, is that if someone calls this function and want to supply some extra metadata, for example to enable connecting to databroker using a DAPR gRPC proxy. they would call it like this:
```
client.subscribe(entries=[entry], metadata=[("dapr-app-id", "server")])
```

Now, how does this work? `metadata` isn't an argument in the function. That's where pythons argument packing comes into play. Inside the `subscribe` function `**rpc_kwargs` will contain the packed _keyword arguments_ like this:

```
{
  "metadata": [("dapr-app-id", "server")]
}
```

which will subsequently be supplied to the actual call to the stub like this:
```
resp_stream = self.client_stub.Subscribe(req, **rpc_kwargs)
```

which is the same as calling it like this:

```
resp_stream = self.client_stub.Subscribe(req, metadata=[("dapr-app-id", "server")])
```

So this is what I mean when I say that you shouldn't overwrite "metadata" if it's already present in `**rpc_kwargs`. _If_ "metadata" is already present, another key, "authorization" should be added to it without overwriting "dapr-app-id" (taking the example provided above).

kuksa-client/kuksa_client/grpc/aio.py
```diff
@@ -262,8 +267,15 @@ async def set(self, *, updates: Collection[EntryUpdate], **rpc_kwargs) -> None:
...
-        raise NotImplementedError('"authorize" is not yet implemented')
+    async def authorize(self, *, token: Dict) -> None:
+        self.authorization_header = "Bearer " + token
+        if self.rpc_metadata != None:
+            metadata = dict(self.rpc_metadata)
+        else: 
+            metadata = dict()
+        metadata = metadata | {"authorization": self.authorization_header}
+        self.rpc_metadata = list(metadata.items())
+
```
argerus, line 278: I'm not sure this change is resolving anything. The function most probably should continue accepting a token of type `str` and not a `Dict`.

I'm also not sure why `self.rpc_metadata` is needed, or that it even makes sense. See my other comment below.

argerus (changes requested):

kuksa-client/kuksa_client/grpc/aio.py
lukasmittag: I agree, why I used this is. If we provide the string from the other function the string looks like ```{'token', 'ajsjdksjdaksdj'}``` so in order to get the token out I used a ```Dict```. Else it would be needed to split the string or something like that or fix it in the fucntion that gets the token from the file. I didn't change the function for getting token, because I thought Romaine had some genius thoughts behind it :)

lukasmittag: I intorduced ```self.rpc_metadata``` because I thought we do not need to check the metadata in every call just when setting a new token. Like you described I was wrong.

kuksa-client/kuksa_client/grpc/__init__.py
lukasmittag: Now I have

kuksa-client/kuksa_client/grpc/aio.py
```diff
@@ -262,18 +271,27 @@ async def set(self, *, updates: Collection[EntryUpdate], **rpc_kwargs) -> None:
...
+    async def authorize(self, *, token: Dict) -> None:
+        self.authorization_header = "Bearer " + token
+
+    async def generate_metadata_header(self, metadata: list) -> list:
+        if metadata != None:
+            metadata = dict(metadata)
+        else:
+            metadata = dict()
+        metadata = metadata | {"authorization": self.authorization_header}
+        return list(metadata.items())
```
argerus, line 283: This doesn't work for me
```
TypeError: unsupported operand type(s) for |: 'dict' and 'dict'
```

```suggestion
        if metadata:
            metadata = dict(metadata)
        else:
            metadata = dict()
        metadata["authorization"] = self.authorization_header
        return list(metadata.items())
```

argerus (changes requested):

kuksa-client/kuksa_client/grpc/aio.py
argerus: > I intorduced self.rpc_metadata because I thought we do not need to check the metadata in every call just when setting a new token.

Ah, now I understand (I think).

So yes, if no `self.authorization_header` is set, no header should be added to the request. What I couldn't understand was all the extra code & variables added. A simple:

```
if self.authorization_header:
    // Add authorization header to request
```

would accomplish that.

Also change the `__init__` function to:

```
self.authorization_header = None
```

kuksa-client/kuksa_client/grpc/aio.py
lukasmittag: Interesting. It's a merge operator. Which python version are you using?

argerus: Ok, haven't seen it before.

I'm running python 3.8.10 (Ubuntu LTS 20.04)

argerus: Either way, it's not even needed here as we're not merging dicts.

kuksa-client/kuksa_client/grpc/aio.py
```diff
@@ -237,6 +244,7 @@ async def get(self, *, entries: Iterable[EntryRequest], **rpc_kwargs) -> List[Da
             rpc_kwargs
                 grpc.*MultiCallable kwargs e.g. timeout, metadata, credentials.
         """
+        rpc_kwargs.__setitem__("metadata", await self.generate_metadata_header(rpc_kwargs.get("metadata")))
```
argerus, line 247: I don't see any reason to call the dunder method explicitly.
```
rpc_kwargs["metadata"] = ...
```
is both shorter and IMO more idiomatic.

Applies to all the instances of this.

lukasmittag:
> 

Are there any opinions on the default tokens. I think if we keep the current approach we should supply an all-read-write token in the container. Plus maybe document hoe to load files in the docker container (and document how to create new tokens).However I think it the token not get set by default. I would suggest the container starts in the state unauthenticated and you need to make a cli command call ```authorize <TOKEN> ```I do not know what happens if you authorize to an instance with no authorization enforced. Maybe nothing. Other opinions are welcome :) Furthermore I have seen that databroker-cli has ```token-file``` and ```token```as cli commands. Should we provide these two options in kuksa client as well? E.g. you can authorize via pasting token string or via specifying token file (currently only token file is supported in kuksa client)

erikbosch:
> 
> For the default tokens. Yeah the feeders maybe rely on them, but I think it is not a "good" way to use some default tokens. I would rather like supplying them in a config file or some environment variables you have to set. For user-friendliness I understand, but you might want to know that you need that stuff if you want to develop something. So I see I'm a bit split with my opinion on this.

I think we need to discuss how we want to use and distribute default tokens in general. So far they are in most cases directly or indirectly included in deliverables (pypi packages, docker containers, ...) so even if the user need to explicitly mention them they exist within the package/container. For the Python feeders currently indirectly included by the dependency to kuksa-client. I think that sort of makes sense. Otherwise we need to find an alternative solution for the user to download the tokens (and certificates) - like releasing them in a zip file that the user can download.

But maybe we shall ignore feeders in this PR and focus on the the command line client instead. Do we expect tokens to be included in the python package and docker container so that the user can do as written in documentation, i.e. call `printTokenDir` and based on that get the default location for broker or server? Or do we expect the user to download by some other mechanism from Github? Regardless of what we choose, I think the documentation needs to be updated to show how to authorize against a "default databroker". I.e. which token do you need to use and how shall you get the token.

kuksa-client/kuksa_client/grpc/aio.py
```diff
@@ -283,6 +292,9 @@ async def subscribe(self, *,
         except AioRpcError as exc:
             raise VSSClientError.from_grpc_error(exc) from exc
 
+    async def authorize(self, *, token: Dict) -> None:
```
argerus, line 295:
```suggestion
    async def authorize(self, *, token: str) -> None:
```

argerus (changes requested):

lukasmittag:
> > > 
> > 
> > 
> > Are there any opinions on the default tokens. I think if we keep the current approach we should supply an all-read-write token in the container. Plus maybe document hoe to load files in the docker container (and document how to create new tokens).However I think it the token not get set by default. I would suggest the container starts in the state unauthenticated and you need to make a cli command call `authorize <TOKEN> `I do not know what happens if you authorize to an instance with no authorization enforced. Maybe nothing. Other opinions are welcome :) Furthermore I have seen that databroker-cli has `token-file` and `token`as cli commands. Should we provide these two options in kuksa client as well? E.g. you can authorize via pasting token string or via specifying token file (currently only token file is supported in kuksa client)
> 
> Sounds reasonable to me. And in one way renaming the `authorize` command to for example `token-file` makes sense as we actually do not perform any authorization when `authorize` command is given. Like in the example below, the `authorize` command only seems to check that the file exist, it does not care about content.
> 
> ```
> Test Client> authorize ~/tmp.txt
> OK
> 
> Test Client> getValue Vehicle.Speed
> {
>     "error": {
>         "code": 16,
>         "reason": "unauthenticated",
>         "message": "Invalid auth token: DecodeError(\"InvalidToken\")"
>     },
>     "errors": []
> }
> 
> Test Client> 
> ```

Should we keep the functions behind cli command? Because VSS server uses authorize too and just sets the token from file. It chekcs back with KUKSA.val server, but authorization is handled different. I think both solutions have potential for confusion by the users. Whats your opinion @erikbosch @argerus

lukasmittag:
> > > 
> > 
> > 
> > Are there any opinions on the default tokens. I think if we keep the current approach we should supply an all-read-write token in the container. Plus maybe document hoe to load files in the docker container (and document how to create new tokens).However I think it the token not get set by default. I would suggest the container starts in the state unauthenticated and you need to make a cli command call `authorize <TOKEN> `I do not know what happens if you authorize to an instance with no authorization enforced. Maybe nothing. Other opinions are welcome :) Furthermore I have seen that databroker-cli has `token-file` and `token`as cli commands. Should we provide these two options in kuksa client as well? E.g. you can authorize via pasting token string or via specifying token file (currently only token file is supported in kuksa client)
> 
> Sounds reasonable to me. And in one way renaming the `authorize` command to for example `token-file` makes sense as we actually do not perform any authorization when `authorize` command is given. Like in the example below, the `authorize` command only seems to check that the file exist, it does not care about content.
> 
> ```
> Test Client> authorize ~/tmp.txt
> OK
> 
> Test Client> getValue Vehicle.Speed
> {
>     "error": {
>         "code": 16,
>         "reason": "unauthenticated",
>         "message": "Invalid auth token: DecodeError(\"InvalidToken\")"
>     },
>     "errors": []
> }
> 
> Test Client> 
> ```

So I thought about this and had the idea (with help of John) to send a get server info call to databroker. Then you will se if it works and authorize command can be kept. I'm not sure if the error message which will be currently provided is the best. 
```
"error": {

        "code": 16,

        "reason": "unauthenticated",

        "message": "Invalid auth token: DecodeError(\"InvalidToken\")"

    },

    "errors": []
```
Should we map this in a way or just do something like ``` raise("Authorization unsuccesful")``` Which would be less transparent to what is the issue.

erikbosch:
I noticed that we use the "old" server tokens as default tokens directory (see https://github.com/eclipse/kuksa.val/blob/master/kuksa-client/kuksa_client/__main__.py#L484). Should we possibly adapt that so we select the correct ones depending on protocol (well rather if we use server or broker).

lukasmittag:
> I noticed that we use the "old" server tokens as default tokens directory (see https://github.com/eclipse/kuksa.val/blob/master/kuksa-client/kuksa_client/__main__.py#L484). Should we possibly adapt that so we select the correct ones depending on protocol (well rather if we use server or broker).

I think this needs some rework, yes. I will do it while adding the default tokens in the docker container.

SebastianSchildt:
The fault token _directory_ as presented by the _client_ is one thing, I would however also remind us of this bug https://github.com/eclipse/kuksa.val/issues/406 

If this is still in, I think now is the time to get rid of it. I do think the _library_ should not bring any default crypto material at all.

Client could be a different story.

Also it would not bite us in the consumers/providers yet, as databroker default is still no security at all. _With_ security by default databroker side, I think it still would be a matter of clear error messages, or bundling some default stuff with the clients, but _not_ somehwere hidden in the library

kuksa-client/kuksa_client/grpc/aio.py
```diff
@@ -55,6 +55,7 @@ def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
         self.channel = None
         self.exit_stack = contextlib.AsyncExitStack()
+        self.token = ""
```
erikbosch, line 58: Why do we want to overwrite whatever we got as part of `*args` and was set in `__init__` in BaseVSSClient?

kuksa-client/kuksa_client/grpc/__init__.py
erikbosch, line 460: Why do we have a token argument if we anyway ignore it?

kuksa-client/kuksa_client/grpc/aio.py
erikbosch, line 58: This overwrites was what set in VSSBaseClient, or?

SebastianSchildt:
I would expect them to not "contain" the material (as copying it would later create consistency issues, and extracting during build from "somewhere" also sees not easy.

I could suggest something, where the default configuration (for now disabled,  as databroker is without default security anyway) something like

```
# Add the JWT access token if you have security enabled. You can find the default testing tokens at "somehwere"
#token=
```

I mean not that exact words, but "somehwere" might refer to a Github path or an artifact, in case we want one.

Later when security becomes default, maybe then the config looks like

```
# blabla
token=
```

and if you start it unmodified, the client might complain to you

`No access token given, please check xy.config to supply token or disable authentication"

Similar pattern for any TLS root cert

lukasmittag:
> The fault token _directory_ as presented by the _client_ is one thing, I would however also remind us of this bug #406
> 
> If this is still in, I think now is the time to get rid of it. I do think the _library_ should not bring any default crypto material at all.
> 
> Client could be a different story.
> 
> Also it would not bite us in the consumers/providers yet, as databroker default is still no security at all. _With_ security by default databroker side, I think it still would be a matter of clear error messages, or bundling some default stuff with the clients, but _not_ somehwere hidden in the library

This bug refers to the client backend so it is not in the library

erikbosch:
- The good thing: I can get dbcfeeder to work this
- The bad thing: Some manual changes needed

I will try to describe them here. What cause problems is the `logger.debug` statement below in https://github.com/boschglobal/kuksa.val/blob/feature/kuksa_client_authorization/kuksa-client/kuksa_client/grpc/__init__.py

```
    def __enter__(self):
        creds = self._load_creds()
        if creds is not None:
            channel = grpc.secure_channel(self.target_host, creds)
        else:
            channel = grpc.insecure_channel(self.target_host)
        self.channel = self.exit_stack.enter_context(channel)
        self.client_stub = val_pb2_grpc.VALStub(self.channel)
        if self.ensure_startup_connection:
            logger.debug("Connected to server: %s", self.get_server_info())
        return self
```

When this is called we have not yet had time to call `authorize`, so `self.authorization_header` is not set. But that does not matter as we do not add `self.authorization_header` in `get_server_info` and you get a `VSSClientError` exception.

- Do we not want to call `self.generate_metadata_header` also in `get_server_info()`?

Some possible changes below. I do not know hos "standardized" the `Bearer`keyword is, or if it is better to have `token` as class argument add and `Bearer` in constructor.

https://github.com/boschglobal/kuksa.val/compare/feature/kuksa_client_authorization...erikbosch:kuksa.val:pr517?expand=1

lukasmittag:
Unit tests for authorization function are missing as of now too.

lukasmittag:
> * The good thing: I can get dbcfeeder to work this
> 
>     * The bad thing: Some manual changes needed
> 
> 
> I will try to describe them here. What cause problems is the `logger.debug` statement below in https://github.com/boschglobal/kuksa.val/blob/feature/kuksa_client_authorization/kuksa-client/kuksa_client/grpc/__init__.py
> 
> ```
>     def __enter__(self):
>         creds = self._load_creds()
>         if creds is not None:
>             channel = grpc.secure_channel(self.target_host, creds)
>         else:
>             channel = grpc.insecure_channel(self.target_host)
>         self.channel = self.exit_stack.enter_context(channel)
>         self.client_stub = val_pb2_grpc.VALStub(self.channel)
>         if self.ensure_startup_connection:
>             logger.debug("Connected to server: %s", self.get_server_info())
>         return self
> ```
> 
> When this is called we have not yet had time to call `authorize`, so `self.authorization_header` is not set. But that does not matter as we do not add `self.authorization_header` in `get_server_info` and you get a `VSSClientError` exception.
> 
>     * Do we not want to call `self.generate_metadata_header` also in `get_server_info()`?
> 
> 
> Some possible changes below. I do not know hos "standardized" the `Bearer`keyword is, or if it is better to have `token` as class argument add and `Bearer` in constructor.
> 
> https://github.com/boschglobal/kuksa.val/compare/feature/kuksa_client_authorization...erikbosch:kuksa.val:pr517?expand=1

Personally I do not think it is necessary to add ```self.generate_metadata_header()``` to the get server info. Because this should work without authorization too. Correct me if I'm wrong but I see no issue why it should be protected to say that the server is databroker and version is bla bla. The standardized way to get the authorization header is a good idea.

lukasmittag:
What I find strange too is that at least for me there should not be logger things in the python library. But this would mean a lot of refactoring. Do not know how you guys see this @erikbosch @SebastianSchildt @argerus

kuksa-client/kuksa_client/grpc/__init__.py
```diff
@@ -562,7 +606,10 @@ def __enter__(self):
             channel = grpc.insecure_channel(self.target_host)
         self.channel = self.exit_stack.enter_context(channel)
         self.client_stub = val_pb2_grpc.VALStub(self.channel)
-        if self.ensure_startup_connection:
+        if self.authorization_header is None:
+            logger.debug(
+                "Can not ensure startup connection without token to authorize")
+        elif self.ensure_startup_connection:
             logger.debug("Connected to server: %s", self.get_server_info())
```
erikbosch, line 613: If you end up here you will still get a `'No auth token provided'` error as `get_server_info()` does not care if `self.authorization_header` is defined. So if you request a client like this:

```
self._grpc_client = self._exit_stack.enter_context(kuksa_client.grpc.VSSClient(
                 host=self._ip,
                 port=self._port,
                 token = token_content,
                 #ensure_startup_connection = False
            ))
```

I.e. with token defined and `self.ensure_startup_connection` set.

If you add:

```
rpc_kwargs["metadata"] = self.generate_metadata_header(
            rpc_kwargs.get("metadata"))
```

... to  `get_server_info()` that use-case seems to work as well.

Then I can sort of agree - does  `get_server_info()` really need to request a token file. Should that not be possible to run without? But as long as it request it we need to provide it as I see it.

erikbosch:
> What I find strange too is that at least for me there should not be logger things in the python library. But this would mean a lot of refactoring. Do not know how you guys see this @erikbosch @SebastianSchildt @argerus

I have no real problem with logging printouts in libraries, as long as they normally never are seen. I.e. debug is OK, and warnings/errors are OK if they indicate that you most likely are using the library in the wrong way or have found a bug, but you do not think it is worth the effort to throw an exception or add an error code, at least not know.

lukasmittag:
For getting this merged I know added ```generate_metadata_header()``` to ```get_server_info()```. Now I will add some unit tests for ```authorization()```. What I've encountered is that if there are some line breaks at the end of file for authorization token the grpc call fails so in the cli backend I did a ```rstrip()```. Could be done directly in authorization function too maybe better if I thin about it now.

erikbosch (approved):
I have verified that this fulfills the dbcfeeder needs, tested by the draft PR in https://github.com/eclipse/kuksa.val.feeders/pull/74

If we have a lot of more to do on this, like testing, we should possibly anyway consider merging/releasing this one as it would be difficult to verify the dbcfeeder docker builds unless we have a released version including this change

kuksa-client/tests/test_grpc.py
```diff
@@ -875,9 +986,20 @@ async def test_set_nonexistent_entries(self, unused_tcp_port, val_servicer):
...
+            name='test_server', version='1.2.3')
+        async with VSSClient('127.0.0.1', unused_tcp_port, ensure_startup_connection=False) as client:
+            success = await client.authorize(token='eyJ0eXAiOiJKV1QiLCJhbGciOiJSUzI1NiJ9.eyJzdWIiOiJsb2NhbCBkZXYiLCJpc3MiOiJjcmVhdGVUb2tlbi5weSIsImF1ZCI6WyJrdWtzYS52YWwiXSwiaWF0IjoxNTE2MjM5MDIyLCJleHAiOjE3NjcyMjU1OTksInNjb3BlIjoicmVhZDpWZWhpY2xlLldpZHRoIHByb3ZpZGU6VmVoaWNsZS5TcGVlZCJ9.w2c8xrYwBVgMav3f0Se6E8H8E36Nd03rJiSS2A8s-CL3GtlwB7wVanjXHhppNsCdWym3tK4JwgslQdMQF-UL4hd7vzdtt-Mx6VjH_jO9mDxz4Z0Uzw7aJtbtQSpi2h6kwceTVTllkbLRF7WRHWIpwzXFF9yZolX6lH-BE9xf1AB62d6icd9SKxFnVvYs3MVK5D1xNmDNOmm-Fr0d2K604MmIIXGW5kPZJYIvBKO4NYRLklhJe47It_lGo3gnh1ppmzTOIo1kB4sDe55hplUCbTCJVricpyQSgTYsf7aFRPK51XMRwwwJ8kShWeaTggMLKpv1W-9dhVWDk4isC8BxsOjaVloArausMmjLmTz6KwAsfARgfXtaCrMsESUBNXi5KIdAyHVXZpmERvc9yeYPcaWlknVFrFsHbV6bw4nwqBX-0Ubuga0NGNQDFKmyTKQrbuZmQ3L9iipxY8_BOSCkdiYtWbE3lpplxpS_PaZl10KAaMmUfbcF9aYZunDEzEtoJgJe2EeGu3XDBtbyXVUKruImdSEdjaImfUGQIWl5bMbVH4N4zK5jE45wT5FJiRUcA5pMN5wNmDYJJzgbxWNpYW40KZYPFc_7XUH8EZ2Cs69wDHam3ArkOs1qMgMIoEPWVzHakjlVJfrPR9zQKxfirBtNNENIoHsBjJ_P4FEJCN4')
```
erikbosch, line 993: Here I think a comment specifying the source of this token would be useful, as we sometime in the future may need to replace it. I.e. refer to the file containing this token, or the code snippet used to create it.

erikbosch (changes requested):
A minor comment on source of token

erikbosch (approved):

PR #518, opened 2023-03-25
Add TLS support

argerus:
Basic functionality is there, but more work is needed in terms of documentation, ease of use.


int0x27 (changes requested):
It would be nice to have some warnings if pub keys, ca-cert, etc. does not start with expected lines
```-----BEGIN XXX-----```

File: kuksa_databroker/databroker/src/main.rs [hunk 1]
```diff
@@ -342,12 +369,40 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
...
+    let tls_config = if args.is_present("insecure") {
+        ServerTLS::Disabled
+    } else {
+        let cert_file = args.get_one::<String>("tls-cert");
+        let key_file = args.get_one::<String>("tls-private-key");
+        match (cert_file, key_file) {
+            (Some(cert_file), Some(key_file)) => {
+                let cert = std::fs::read(cert_file)?;
+                let key = std::fs::read(key_file)?;
+                let identity = tonic::transport::Identity::from_pem(cert, key);
```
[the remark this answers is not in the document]
argerus: It doesn't seem possible using this library..

int0x27 (approved):
looks good

kuksa_databroker/databroker/src/main.rs
```diff
@@ -247,14 +247,15 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
                 .value_name("PORT")
                 .required(false)
                 .env("KUKSA_DATA_BROKER_PORT")
-                .value_parser(clap::value_parser!(u16).range(1..))
+                .value_parser(clap::value_parser!(u16))
```
sophokles73, line 250: my understanding is that the port number `0` can be used to bind to *any free port*? If so, then I think this is worth mentioning in the property's description both in the source code as well as in the README

argerus: Yes, that is correct (and the reason I removed the limit of > 0).

With that said, it doesn't really do much good at the moment as databroker isn't printing which port it binds to (when using port 0) as it's just printing the port provided verbatim..

I have a plan to fix it, but as it's a bit involved I didn't do it as part of this PR.
Regarding documenting this behaviour, I mean sure it doesn't hurt, but it's standard unix behaviour, right?

SebastianSchildt (approved):

Issue #476, opened 2023-02-10
Python lib examples only working with "with"


SebastianSchildt:
Current python lib uses "context managers" to enable usage in `with` statements. The canonical example from the documentation is

```python
with VSSClient('127.0.0.1', 55556) as client:
    current_values = client.get_current_values([
        'Vehicle.Speed',
    ])
    print(current_values)
```

the following code does not work

```python
databroker=VSSClient('127.0.0.1', 55556)
current_values2 = databroker.get_current_values([
    'Vehicle.Speed',
])
print(current_values2)
```

this is unexpected and not how it works for example with file io in python. here

```python
file = open("some.file", "r")
```

can achieve the same as

```python
with open("some.file", "r") as file:
```

The root cause seems to be, that without `with``__enter__`is not called, see the error message for the second example

```
Traceback (most recent call last):  File "/Users/scs2rng/Documents/Dev/kuksa-prometheus/test.py", line 11, in <module>    current_values2 = databroker.get_current_values([  File "/Users/scs2rng/.local/share/virtualenvs/kuksa-prometheus-n8PS3LmC/lib/python3.10/site-packages/kuksa_client/grpc/__init__.py", line 586, in get_current_values    entries = self.get(  File "/Users/scs2rng/.local/share/virtualenvs/kuksa-prometheus-n8PS3LmC/lib/python3.10/site-packages/kuksa_client/grpc/__init__.py", line 738, in get    resp = self.client_stub.Get(req, **rpc_kwargs)AttributeError: 'NoneType' object has no attribute 'Get'
```

It seems a bit unclear what the best python (coding) practice to achieve a better behavior is

erikbosch:
If we change something we must make sure feeders, still work. As an example dbcfeeder creates client with:

```
 vss_client = self._exit_stack.enter_context(kuksa_client.grpc.VSSClient(
                host=self._kuksa_client_config['ip'],
                port=self._kuksa_client_config['port'],
            ))
```

(Which results in that `__enter__`method is called)

SebastianSchildt:
Hah, that is likely the "workaround" if you want a vss_client object. 
I have seen code where `__enter__` basically just returns ´self´, and the initialisation/acquiring of resources is already done in `__init__`. Not sure that is how context managers are intended to be used, but it still gives the advantage of cleanup via `__exit__`  should something go wrong, and would probably be compatible with existing code.

[...]

PR #526, opened 2023-04-03
Use kuksa-client as non context-manager

lukasmittag:
This solves #476

### DIFF
--- a/kuksa-client/kuksa_client/__init__.py
+++ b/kuksa-client/kuksa_client/__init__.py
@@ -35,6 +35,7 @@ from kuksa_client._metadata import *
 
 from . import cli_backend
 
+
 class KuksaClientThread(threading.Thread):
 
     # Constructor
@@ -86,16 +87,22 @@ class KuksaClientThread(threading.Thread):
     # The given callback function will be called then, if the given path is updated:
     #   updateMessage = await webSocket.recv()
     #   callback(updateMessage)
-    def subscribe(self, path, callback, attribute = "value", timeout=5):
+    def subscribe(self, path, callback, attribute="value", timeout=5):
         return self.backend.subscribe(path, callback, attribute, timeout)
 
-    def subscribeMultiple(self, paths, callback, attribute = "value", timeout=5):
+    def subscribeMultiple(self, paths, callback, attribute="value", timeout=5):
         return self.backend.subscribeMultiple(paths, callback, attribute, timeout)
 
     # Unsubscribe value changes of to a given path.
     # The subscription id from the response of the corresponding subscription request will be required
     def unsubscribe(self, sub_id, timeout=5):
         return self.backend.unsubscribe(sub_id, timeout)
+
+    def disconnect(self, timeout=5):
+        return self.backend.disconnect(timeout)
+
+    def connect(self, timeout=5):
+        return self.backend.connect(timeout)
 
     # Thread function: Start the asyncio loop
     def run(self):
--- a/kuksa-client/kuksa_client/__main__.py
+++ b/kuksa-client/kuksa_client/__main__.py
@@ -138,8 +138,10 @@ class TestClient(Cmd):
 
     ap_getServerAddr = argparse.ArgumentParser()
     ap_connect = argparse.ArgumentParser()
-    ap_connect.add_argument('-i', "--insecure", default=False, action="store_true", help='Connect in insecure mode')
     ap_disconnect = argparse.ArgumentParser()
+    ap_connectThread = argparse.ArgumentParser()
+    ap_connectThread.add_argument('-i', "--insecure", default=False, action="store_true", help='Connect in insecure mode')
+    ap_disconnectThread = argparse.ArgumentParser()
     ap_authorize = argparse.ArgumentParser()
     tokenfile_completer_method = functools.partial(Cmd.path_complete,
         path_filter=lambda path: (os.path.isdir(path) or path.endswith(".token")))
@@ -414,13 +416,19 @@ class TestClient(Cmd):
 
 
     @with_category(COMM_SETUP_COMMANDS)
-    @with_argparser(ap_disconnect)
-    def do_disconnect(self, _args):
+    @with_argparser(ap_disconnectThread)
+    def do_disconnectThread(self, _args):
         """Disconnect from the VISS/gRPC Server"""
         if hasattr(self, "commThread"):
             if self.commThread is not None:
                 self.commThread.stop()
                 self.commThread = None
+    
+    @with_category(COMM_SETUP_COMMANDS)
+    @with_argparser(ap_disconnect)
+    def do_disconnect(self, _args):
+        """Disconnect from the VISS/gRPC Server"""
+        self.commThread.disconnect()
 
     def checkConnection(self):
         if self.commThread is None or not self.commThread.checkConnection():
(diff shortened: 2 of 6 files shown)


SebastianSchildt (changes requested):
Calling quit in kuksa-client yields exception:

```
Test Client> quit
  File "/Users/scs2rng/.local/share/virtualenvs/testkc-dPiaaAKY/bin/kuksa-client", line 8, in <module>
    sys.exit(main())
  File "/Users/scs2rng/Documents/Dev/kuksa.val/kuksa-client/kuksa_client/__main__.py", line 550, in main
    clientApp.stop()
  File "/Users/scs2rng/Documents/Dev/kuksa.val/kuksa-client/kuksa_client/__main__.py", line 384, in stop
    self.commThread.stop()
  File "/Users/scs2rng/Documents/Dev/kuksa.val/kuksa-client/kuksa_client/__init__.py", line 51, in stop
    self.backend.stop()
AttributeError: 'Backend' object has no attribute 'stop'
```

SebastianSchildt:
Started some testing on async side, first example (shortened), works when chaign from 

```python
async with VSSClient('127.0.0.1', 55557) as client:
        current_values = await client.get_current_values([
            'Vehicle.Speed',
        ])
        print(current_values['Vehicle.Speed'].value)
```

to     

```python 
    myclient=VSSClient('127.0.0.1', 55557)
    await myclient.connect()
    current_values = await myclient.get_current_values([
            'Vehicle.Speed',
    ])
    print(current_values['Vehicle.Speed'].value)
```

Both version work, however subscription is broken, the vanilla example (complete)

```python
import asyncio

from kuksa_client.grpc import Datapoint
from kuksa_client.grpc.aio import VSSClient

async def main():
    async with VSSClient('127.0.0.1', 55557) as client:
        await client.set_target_values({
            'Vehicle.Body.Windshield.Front.Wiping.System.TargetPosition': Datapoint(45),
        })

        async for updates in client.subscribe_current_values([
            'Vehicle.Body.Windshield.Front.Wiping.System.TargetPosition',
        ]):
            current_position = updates['Vehicle.Body.Windshield.Front.Wiping.System.TargetPosition'].value
            print(f"Current wiper position is: {current_position}")


asyncio.run(main())
```

fails with

```
 % python asynciosubscribe.py 
/private/tmp/testkc/asynciosubscribe.py:12: RuntimeWarning: coroutine 'VSSClient.check_connected_async.<locals>.wrapper' was never awaited
  async for updates in client.subscribe_current_values([
RuntimeWarning: Enable tracemalloc to get the object allocation traceback
  File "/private/tmp/testkc/asynciosubscribe.py", line 19, in <module>
    asyncio.run(main())
  File "/Library/Developer/CommandLineTools/Library/Frameworks/Python3.framework/Versions/3.9/lib/python3.9/asyncio/runners.py", line 44, in run
    return loop.run_until_complete(main)
  File "/Library/Developer/CommandLineTools/Library/Frameworks/Python3.framework/Versions/3.9/lib/python3.9/asyncio/base_events.py", line 642, in run_until_complete
    return future.result()
  File "/private/tmp/testkc/asynciosubscribe.py", line 12, in main
    async for updates in client.subscribe_current_values([
TypeError: 'async for' requires an object with __aiter__ method, got coroutine
```

It works well with the released version for kuksa-client (may only complain a None Type is returned, if the value has not been set before)

lukasmittag:
> Started some testing on async side, first example (shortened), works when chaign from
> 
> ```python
> async with VSSClient('127.0.0.1', 55557) as client:
>         current_values = await client.get_current_values([
>             'Vehicle.Speed',
>         ])
>         print(current_values['Vehicle.Speed'].value)
> ```
> 
> to
> 
> ```python
>     myclient=VSSClient('127.0.0.1', 55557)
>     await myclient.connect()
>     current_values = await myclient.get_current_values([
>             'Vehicle.Speed',
>     ])
>     print(current_values['Vehicle.Speed'].value)
> ```
> 
> Both version work, however subscription is broken, the vanilla example (complete)
> 
> ```python
> import asyncio
> 
> from kuksa_client.grpc import Datapoint
> from kuksa_client.grpc.aio import VSSClient
> 
> async def main():
>     async with VSSClient('127.0.0.1', 55557) as client:
>         await client.set_target_values({
>             'Vehicle.Body.Windshield.Front.Wiping.System.TargetPosition': Datapoint(45),
>         })
> 
>         async for updates in client.subscribe_current_values([
>             'Vehicle.Body.Windshield.Front.Wiping.System.TargetPosition',
>         ]):
>             current_position = updates['Vehicle.Body.Windshield.Front.Wiping.System.TargetPosition'].value
>             print(f"Current wiper position is: {current_position}")
> 
> 
> asyncio.run(main())
> ```
> 
> fails with
> 
> ```
>  % python asynciosubscribe.py 
> /private/tmp/testkc/asynciosubscribe.py:12: RuntimeWarning: coroutine 'VSSClient.check_connected_async.<locals>.wrapper' was never awaited
>   async for updates in client.subscribe_current_values([
> RuntimeWarning: Enable tracemalloc to get the object allocation traceback
> Traceback (most recent call last):
>   File "/private/tmp/testkc/asynciosubscribe.py", line 19, in <module>
>     asyncio.run(main())
>   File "/Library/Developer/CommandLineTools/Library/Frameworks/Python3.framework/Versions/3.9/lib/python3.9/asyncio/runners.py", line 44, in run
>     return loop.run_until_complete(main)
>   File "/Library/Developer/CommandLineTools/Library/Frameworks/Python3.framework/Versions/3.9/lib/python3.9/asyncio/base_events.py", line 642, in run_until_complete
>     return future.result()
>   File "/private/tmp/testkc/asynciosubscribe.py", line 12, in main
>     async for updates in client.subscribe_current_values([
> TypeError: 'async for' requires an object with __aiter__ method, got coroutine
> ```
> 
> It works well with the released version for kuksa-client (may only complain a None Type is returned, if the value has not been set before)

I see, will dig deeper into this issue.

kuksa-client/kuksa_client/__init__.py
```diff
@@ -86,17 +87,23 @@ def getValues(self, paths: Iterable[str], attribute="value", timeout=5):
...
 
     # Unsubscribe value changes of to a given path.
     # The subscription id from the response of the corresponding subscription request will be required
     def unsubscribe(self, sub_id, timeout=5):
         return self.backend.unsubscribe(sub_id, timeout)
 
+    def disconnect(self, timeout=5):
+        return self.backend.disconnect(timeout)
+
+    def connect(self, timeout=5):
```
SebastianSchildt, line 104: Wouln't this break all existing users of threaded API, i.e. GPS feeder, S3 example etc? Because they are not calling connect?

lukasmittag: I don't think so because they all call connect from within ```__enter__``` and the threaded API works with the cli_backends which leverage asysnc/sync library parts. I tested with gps feeder and a virtual environment and worked for me.

lukasmittag:
> It works, but I do think it is not ideal that the CLI exposed to the user has (dis)connectThread and (dis)connect functions that do slightly different things.
> 
> I think I want to `connect` (and maybe `disconnect`, although likely I'll just call another connect instead), and I feel those functions should be doing what the "thread" versions are doing now (at least in the connect case)

Agreed, what do you mean by ```another connect instead```

lukasmittag:
> I mean in a workflow I will likely do setServerAdress ... and then connect, I then might do another setServerAdress (to a different location) and call connect again. I would assume it disconnects the previous connection automatically before reconnecting. Therefore, It hink while having a disconnect is consistent and does not hurt, people will likely not use it very often (and should not be "required" to do so, when they decide to connect to a different location)

Tested and works as you expected

SebastianSchildt (approved):
Works for me, tested client and (a)sync examples 

lgtm 🐢

PR #527, opened 2023-04-04
Dash-license support in CI

SebastianSchildt:
Eclipse uses dash-licenses (https://github.com/eclipse/dash-licenses)  to vet license of dependencies.

This PR does two things to better integrate with Eclipse process with less manual work for our RUST components

 - it adds support to createbom to generate the inout file required by dash. I chose to create it via out existing tooling, instead of using the cmdline tricks in the dash-documentaiton example, so we have a "single source of truth" wrt to dependencies (i.e. what goes into the bom in the containers, is the same thing checked against dash).  In both cases this relies on cargo to get hte information
 - I added a composite action, that will take a Dash input file, run it, and archive the full report, as well as the stdout/stderr output form dash
 - I integrated into databroker and databroker-cli builds. I chose to extend an existing stage (lint), because otherwise we would need to set up a complete rust environment a fourth time in the build process
 - Dash check can never fail, only createbom can (and will) if it sees a non-whitelisted license. Rationale is, that sometimes the project may decide to add a component with a "fitting" license between releases, which may not yet have been cleared/scanned yet. So this avoids deadlock (at the price we might need to pull a dependency later. But this is how it has always been.)

Easiest way to check any new dependencies, is checking the build summary, where you get a post like this

image

See here for a complete run https://github.com/boschglobal/kuksa.val/actions/runs/4607432259

This is also a good example, where dash finds some deps that are still in the clearing process. If something new is coming up, the "dash data" aritfact contains the Dash input. This can be used by project leads, e.g. @bs-jokri  directly to trigger the scanning in Eclipse IP ticketing system

### DIFF
--- a/.github/actions/check-dash/action.yml
+++ b/.github/actions/check-dash/action.yml
@@ -1,0 +1,31 @@
+
+name: Checking dependencies with dash-license tool
+description: Check an input file via dash, and archive report. Print output in build step summary. Requires wget and an JRE on the runner
+
+inputs:
+  dashinput:
+    required: true
+    type: string
+    description: "Dash Input file"
+
+
+runs:
+  using: "composite"
+  steps:
+    - name: "Run Dash"
+      shell: bash
+      run: |
+        wget -O dash.jar "https://repo.eclipse.org/service/local/artifact/maven/redirect?r=dash-licenses&g=org.eclipse.dash&a=org.eclipse.dash.licenses&v=LATEST"
+        java -jar dash.jar -summary ${{ inputs.dashinput }}.report  ${{ inputs.dashinput }} > ${{ inputs.dashinput }}.out 2>&1 || true
+        echo -e "Dash output: \n\`\`\` " >> $GITHUB_STEP_SUMMARY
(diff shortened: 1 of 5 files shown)


sophokles73:
In other projects that I am involved in we keep actually keep the output of the Dash tool in the repository itself. This can then also be used as reference for project reviews by the Eclipse EMO team.

bs-jokri:
@SebastianSchildt LGTM. One improvement might be to provide the link to the archived Dash output as comment in the PR. This way, it is easier to access the corresponding artifacts. (Maybe this already happens but I haven't seen it in the code).

bs-jokri (approved):
LGTM

SebastianSchildt:
@bs-jokri suggestion is good, however  that is currently not possible, as the GH action permission system does not allow posting comments, when coming from an external fork (GH standard token does not have the sope then, and any custom secrets would not be accessible in the runs). There are hacks around it running an external  webservice, but we do not currently have that

Will merge for now, and keep it in mind, whether this is something that might be set up, once we are in a new org. Currently any "hacks" would need to be done on a per-repo basis which complicates things further

Issue #493, opened 2023-03-01
Improve documentation on value type in Python API


erikbosch:
As of today most methods in the Python API does not provide any type hints, and does not provide any comments either. Taking the example below it is difficult to deduce what types are allowed for "value".  At least if using websocket/VISS API, the value must be a JSOn object: str, bytes or bytearray. Trying to use a Python int or bool will give an error.

I do not know if we as of now expect it to be called with bytes/bytearray as input, or if we always expect a string? If so we could possibly use str as type hint.

https://github.com/eclipse/kuksa.val/blob/master/kuksa-client/kuksa_client/__init__.py#L70

```
    # Set value to a given path
    def setValue(self, path, value, attribute="value", timeout=5):
        return self.backend.setValue(path, value, attribute, timeout)
```

lukasmittag:
```
# Update Meta Data of a given path
    def updateMetaData(self, path, jsonStr, timeout=5):
        return self.backend.updateMetaData(path, jsonStr, timeout)
```
This could be a good example we named this variable ```jsonStr``` but it accepts Any type in function call.

lukasmittag:
solved by #529

PR #529, opened 2023-04-13
type hints in Python API

lukasmittag:
Added type hints in implementation of kuksa client. If I'm not wrong the type hints already exist in python library. Would solve #493

### DIFF
--- a/kuksa-client/kuksa_client/__init__.py
+++ b/kuksa-client/kuksa_client/__init__.py
@@ -52,23 +52,23 @@ class KuksaClientThread(threading.Thread):
         self.backend.stop()
 
     # Do authorization by passing a jwt token or a token file
-    def authorize(self, token_or_tokenfile=None, timeout=5):
+    def authorize(self, token_or_tokenfile: str=None, timeout=5):
         return self.backend.authorize(token_or_tokenfile, timeout)
 
     # Update VSS Tree Entry
-    def updateVSSTree(self, jsonStr, timeout=5):
+    def updateVSSTree(self, jsonStr: str, timeout=5):
         return self.backend.updateVSSTree(jsonStr, timeout)
 
     # Update Meta Data of a given path
-    def updateMetaData(self, path, jsonStr, timeout=5):
+    def updateMetaData(self, path: str, jsonStr: str, timeout=5):
         return self.backend.updateMetaData(path, jsonStr, timeout)
 
     # Get Meta Data of a given path
-    def getMetaData(self, path, timeout=5):
+    def getMetaData(self, path: str, timeout=5):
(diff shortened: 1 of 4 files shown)


kuksa-client/kuksa_client/__init__.py
```diff
@@ -51,32 +51,32 @@ def checkConnection(self):
     def stop(self):
         self.backend.stop()
 
-    # Do authorization by passing a jwt token or a token file
-    def authorize(self, token=None, timeout=5):
-        return self.backend.authorize(token, timeout)
+    # Do authorization by passing a token file
+    def authorize(self, tokenfile: str = None, timeout=5):
```
erikbosch, line 55: This is maybe a good naming convention to gradually introduce also in other places - use `token` when we refer to the contents of a token, and `tokenfile` when the content is supposed to be a path

kuksa-client/kuksa_client/grpc/__init__.py
```diff
@@ -858,6 +858,11 @@ def subscribe(self, *, entries: Iterable[SubscribeEntry], **rpc_kwargs) -> Itera
 
     @check_connected
     def authorize(self, *, token: str, **rpc_kwargs) -> str:
+        """
+        Parameters:
+            rpc_kwargs
```
erikbosch, line 863: Even if we would have a naming convention - as we not yet are consistent - would it be an idea here to mention in a comment that token shall be the actual token, not a file path.

kuksa-client/kuksa_client/grpc/aio.py
```diff
@@ -318,6 +318,11 @@ async def subscribe(self, *,
 
     @check_connected_async
     async def authorize(self, token: str, **rpc_kwargs) -> str:
```
erikbosch, line 320: Should we possibly when changing here also add a `*, `after `self` to make it identical to the same method in ` __init__.py` and similar to the other methods in this file?

kuksa-client/kuksa_client/grpc/__init__.py
lukasmittag: Yes, good point

kuksa-client/kuksa_client/grpc/aio.py
argerus: My question would be _why_ `*` is included in those other function signatures...

erikbosch: Fair point - as long as it is consistent I am happy.

erikbosch:
@lukasmittag  - apart from fixing the conflicts, is there something more to do? I think it would be good to get this one merged and then create a new kuksa-client pre-release

PR #530, opened 2023-04-13
KUKSA client fails because at startup it is unauthenticated

lukasmittag:
This PR will fix the issue that the KUKSA client fails due to unathenticted state at the beginning. Alternatively we could give a default token which has no rights but the right structure (possible? @argerus). But I think we are good with not ensuring startup connection. Maybe we can even drop the ensure startup connection? it just calls ```get_server_info``` to ensure databroker is there. But for now I think this is a quick little fix. An other alternative could be to give the ability to set a token while strating the client through command line.

### DIFF
--- a/kuksa-client/kuksa_client/__main__.py
+++ b/kuksa-client/kuksa_client/__main__.py
@@ -264,6 +264,7 @@ class TestClient(Cmd):
         self.subscribeIds = set()
         self.commThread = None
         self.token_or_tokenfile = token_or_tokenfile
+        self.insecure = insecure
 
         print("Welcome to Kuksa Client version " + str(_metadata.__version__))
         print()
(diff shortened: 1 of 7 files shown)


kuksa-client/kuksa_client/__main__.py
```diff
@@ -147,10 +147,9 @@ def subscriptionIdCompleter(self, text, line, begidx, endidx):
     tokenfile_completer_method = functools.partial(Cmd.path_complete,
                                                    path_filter=lambda path: (os.path.isdir(path) or path.endswith(".token")))
     ap_authorize.add_argument(
-        'Token',
+        'tokenfile',
         help='JWT(or the file storing the token) for authorizing the client.',
```
erikbosch, line 151: Possibly rephrase - It must be a file path right? I.e. you cannot give the actual token as argument?

SebastianSchildt: On the CLI you can. It will try to be clever figuring out whether it was a token or a file containing a token

lukasmittag: This seems to not work (anymore?) Are you sure it was like this at some time?

SebastianSchildt: yes, see e.g. here https://github.com/eclipse/kuksa.val/blob/38edbfedfac9b0fd8ad8a27903b0a6c9d5562f0f/kuksa_viss_client/__init__.py#L78 (not sure this is the earliest/latest time it worked, just chosen it randomly). basically autorize gets "token", if itis ane xisting file, it will laod it and replace it with the file contetns, if not, it jsut skips taht and sues "token" as is, thus you could either do "authorizes <paste JWT token>" or "authorize <path_to_jwt_token>", that was quite useful, i.e. if life playing with token generation on jwt.io, you could just copy & paste from there

lukasmittag: Ah, this is not in the kuksa-client anymore (My GitLens says Naresh removed it). If we want this I could try adding this. Aligned with Erik and maybe then we should call the argument token_or_tokenfile or something similar. But maybe it would be a good addition.

erikbosch (approved):

SebastianSchildt (approved):

PR #536, opened 2023-04-25
Update requirements.txt

lukasmittag:
Update cryptograhpy to solve security vulnerabilities


SebastianSchildt:
You should not push branches to upstream directly. Rather use a fork

SebastianSchildt (approved):
Token generation still works

PR #537, opened 2023-04-26
Check for SPDX licenses

lukasmittag:
It checks via https://github.com/enarx/spdx action if a SPDX file exists.

### DIFF
--- a/.github/workflows/check_license.yaml
+++ b/.github/workflows/check_license.yaml
@@ -16,6 +16,7 @@ jobs:
 
     - name: Get changed files
       run: |
+        git fetch origin
         echo "files=$(git diff --name-only origin/${{ github.event.pull_request.base.ref }}..origin/${{ github.event.pull_request.head.ref }} | tr '\n' ',')" >> $GITHUB_ENV
         
     - name: Check SPDX Licenses
--- a/.github/workflows/check_license.yaml
+++ b/.github/workflows/check_license.yaml
@@ -8,16 +8,23 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     
-    - name: Checkout code
-      uses: actions/checkout@v3
+    - name: Checkout forked repo
+      uses: actions/checkout@v2
       with:
-        # required to grab the history of the PR
-        fetch-depth: 0
+          ref: pull/${{ github.event.number }}/head:${{ github.event.pull_request.head.ref }}
+          repository: ${{ github.event.pull_request.head.repo.full_name }}
+      
+    - name: Checkout original repo
+      uses: actions/checkout@v2
+      with:
+        ref: pull/${{ github.event.number }}/head:${{ github.event.pull_request.head.ref }}
+        repository: ${{ github.event.pull_request.base.repo.full_name }}
+
 
     - name: Get changed files
       run: |
         git fetch origin
-        echo "files=$(git diff --name-only origin/${{ github.event.pull_request.base.ref }}..origin/${{ github.event.pull_request.head.ref }} | tr '\n' ',')" >> $GITHUB_ENV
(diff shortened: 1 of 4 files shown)


.github/workflows/check_license.yaml
```diff
@@ -0,0 +1,14 @@
...
+
+jobs:
+  check-spdx-headers:
+    runs-on: ubuntu-latest
+    steps:
+    - name: checkout
+      uses: actions/checkout@v2
+    - uses: enarx/spdx@master
+      with:
+        licenses: Apache-2.0 MIT
```
erikbosch, line 14: Missing newline, I assume we want that?

lukasmittag:
I do not know if this is the best solution. First I leveraged enarx/spdx@master but they would look through every file and it would take a long time (after 2 hours of running the action I stopped it). So I went with another approach to check with git diff which files were modified and only check them. Therefore I will match to https://spdx.org/licenses/Apache-2.0.html. But currently not working it seems. I will investigate further.

lukasmittag:
Once you have verified my changes I will remove the two test files.

.github/workflows/check_license.yaml
```diff
@@ -0,0 +1,38 @@
...
+        git remote add eclipse https://github.com/eclipse/kuksa.val
+        git fetch origin
+        git fetch boschglobal
+        git fetch eclipse
+        if git rev-parse --verify origin/${{ github.event.pull_request.head.ref }} >/dev/null 2>&1; then
+          echo "files=$(git diff --name-only eclipse/${{ github.event.pull_request.base.ref }}..origin/${{ github.event.pull_request.head.ref }} | tr '\n' ',')" >> $GITHUB_ENV
+        elif git rev-parse --verify eclipse/${{ github.event.pull_request.head.ref }} >/dev/null 2>&1; then
+          echo "Failed to run git diff command with origin. Trying with eclipse repo..."
+          echo "files=$(git diff --name-only eclipse/${{ github.event.pull_request.base.ref }}..origin/${{ github.event.pull_request.head.ref }} | tr '\n' ',')" >> $GITHUB_ENV
+        elif git rev-parse --verify boschglobal/${{ github.event.pull_request.head.ref }} >/dev/null 2>&1; then
```
erikbosch, line 26: Do we know when/if when we end up here, i.e. need to compare with boschglobal? If we could avoid referencing boschglobal it would be great.

.github/workflows/verify-spdx-headers
```diff
@@ -0,0 +1,140 @@
...
+            if license == None:
+                failed = True
+                print(f"NO SPDX {path}")
+            else:
+                print(f"{license:16} {path}")
+            rv = 1
+    if failed:
+        raise Exception("Check the output, some files have not the right licenses!")
+
+    raise SystemExit(rv)
```
erikbosch, line 140: Do we prefer a newline?

.github/workflows/check_license.yaml
lukasmittag: So, I did not find out why it is not working with origin. Then I tried boschglobal and it worked. That’s why I added eclipse and boschglobal checks. Maybe you know more? What is my understanding is that the checkout part should check out the branch but maybe the beach does not belong to a repository or it could not get resolved. What would be possible is to check via boschglobal and eclipse commits. But then other „own” forks would not work. I will do some investigating again

lukasmittag: Okay, I think I found a solution so resolving this :) hope you are happy now :)

erikbosch:
Concerning the "fork" discussion, you could possibly compare with https://github.com/COVESA/vss-tools/blob/master/.github/workflows/pre-commit.yml where `fetch-depth: 0` seems to do the trick to make sure that everything between "from-version" and "new-version" is present. Have however not tested if that is the solution here as well

lukasmittag:
Will close this PR because of #541 which solves it better

PR #541, opened 2023-05-04
Check SPDX license

erikbosch:
This PR adds checks for SPDX identifiers in files added/modified.
Action could possibly be reused by other KUKSA.val repositories
This PR replaces #537


lukasmittag (approved):
Looks good!

SebastianSchildt:
I see there have been changes after @lukasmittag  approved. What are they and/pr is it still okay @lukasmittag

lukasmittag:
Yes, Erik made the action available from outside to be used e.g. in feeder and services repo. Else no changes so approved

PR #544, opened 2023-05-04
Document python library changes (non context manager, authorization)

lukasmittag:


kuksa-client/docs/examples/async-grpc.md
```diff
@@ -18,6 +18,8 @@ This API makes the assumption you're interested in getting/setting/subscribing t
 
 Hence this is the go-to API if you're seeking minimum boilerplate and the most basic functionality.
 
+#### Unauthorized
```
erikbosch, line 21: Would it possibly be better to have more elaborate headings, like

#### Connecting to a Databroker (without Authorization)


#### Connecting to a Databroker (with Authorization)

kuksa-client/docs/examples/async-grpc.md
```diff
@@ -34,6 +36,21 @@ async def main():
...
 ```
 
+Besides this there is a solution where you are not using the client as context-manager
+```python
+import asyncio
+
+from kuksa_client.grpc.aio import VSSClient
+
+async def main():
+    client = VSSClient('127.0.0.1', 55555, ensure_startup_connection=True)
```
SebastianSchildt, line 46: why do wen need ensure_startup_connection=True?

kuksa-client/docs/examples/async-grpc.md
```diff
@@ -34,6 +36,21 @@ async def main():
...
+    await client.connect()
+    current_values = await client.get_current_values(['Vehicle.Speed'])
+    print(current_values['Vehicle.Speed'].value)
```
SebastianSchildt, line 49: Crashes on freshly created databroker (becasue there is no Datapoint attached, current_values['Vehicle.Speed'] is "none"

For minimal example     
```
print(current_values['Vehicle.Speed'])
```

might be a suitbale fix

kuksa-client/docs/examples/async-grpc.md
```diff
@@ -73,6 +90,44 @@ async def main():
...
+
+async def main():
+    # token from kuksa.val/jwt/provide-all.json
+    _token = "eyJ0eXAiOiJKV1QiLCJhbGciOiJSUzI1NiJ9.eyJzdWIiOiJsb2NhbCBkZXYiLCJpc3MiOiJjcmVhdGVUb2tlbi5weSIsImF1ZCI6WyJrdWtzYS52YWwiXSwiaWF0IjoxNTE2MjM5MDIyLCJleHAiOjE3NjcyMjU1OTksInNjb3BlIjoicHJvdmlkZSJ9.OJWzTvDjcmeWyg3vmBR5TEtqYaHq8HrpFLlTKZAfDBAQBUHpyUEboJ97jfWuWgBnTpnfboyfAbwvLqo6bEVZ6tXzF8n9LtW6HmPbIWoDqXuobM2grUCVaGKuOcnCpMCQYChziqHbYwRJYP9nkYgbQU1kE4dN7880Io4xzq0GEbWksB2CVpOoExQUmCZpCohPs-XEkdmXhcUKnWnOeiSsRGKusx987vpY_WOXh6WE7DfJgzAgpPDo33qI7zQuTzUILORQsiHmsrQO0-zcvokNjaQUzlt5ETZ7MQLCtiUQaN0NMbDMCWkmSfNvZ5hKCNbfr2FaiMzrGBOQdvQiFo-DqZKGNweaGpufYXuaKfn3SXKoDr8u1xDE5oKgWMjxDR9pQYGzIF5bDXITSywCm4kN5DIn7e2_Ga28h3rBl0t0ZT0cwlszftQRueDTFcMns1u9PEDOqf7fRrhjq3zqpxuMAoRANVd2z237eBsS0AvdSIxL52N4xO8P_h93NN8Vaum28fTPxzm8p9WlQh4mgUelggtT415hLcxizx15ARIRG0RiW91Pglzt4WRtXHnsg93Ixd3yXXzZ2i4Y0hqhj_L12SsXunK2VxKup2sFCQz6wM-t_7ADmNYcs80idzsadY8rYKDV8N1WqOOd4ANG_nzWa86Tyu6wAwhDVag5nbFmLZQ"
+    async with VSSClient('127.0.0.1', 55555, token=_token) as client:
+        current_values = await client.get_current_values([
+            'Vehicle.Speed',
+        ])
+        print(current_values['Vehicle.Speed'].value)
```
SebastianSchildt, line 126: ^^

kuksa-client/docs/examples/async-grpc.md
lukasmittag: We do not need it there just wanted to mention it somewhere. The default of this is True.

kuksa-client/docs/examples/async-grpc.md
lukasmittag: Yeah. Might be a fix, but is not really what we want maybe an if to catch it could be an idea. But maybe it’s ok for example code. The examples before this PR were not „better“ but I guess we want it „better“ ;)

SebastianSchildt: Always want it "better", in "real code" I usually do

```python
if values[whatever_path] is not None: #that means there is a datapoint there
    do_stuff_with(values[whatever_path].value)
´´´

was just thinking in terms of minimal example it might be better or also ok, to have less lines, and just print `values[whatever_path]`

I think both is fine, but the current situation (I know it was like this before), that you get an uncaught exception running the examples in typical situation (i.e. a freshly started databroker) is not good

lukasmittag:
Ok will redo

SebastianSchildt (approved):
lgtm 🐒

PR #548, opened 2023-05-05
Consistency in basic sysarch figure. Order and arows

SebastianSchildt:
Changed arrow direction to reflect logical flow of data for consumers

Changed consumer order for easier readability


lukasmittag (commented):
Just some thoughts from my side. Regarding the column with the App n, Provider n: Shouldn't it be a double arrow from Server to Consumer n? because on the southbound the provider has a double arrow too. But then we maybe need to remove the subscribe Vehicle.Speed.

SebastianSchildt:
How about, on Consumer 1 I change text to get/subscribe to indicate both options, and on "consumer n" I remove the text and make a double arrow, das you suggested?


The the first two are like "examples" ordered according to logical data flow direction, and the "n" is the generic template

SebastianSchildt:
Done

PR #549, opened 2023-05-05
Introducing VSS 4.0

erikbosch:


data/vss-core/README.md
```diff
@@ -9,6 +9,7 @@ In addition older versions may be supported. This folder contains copies of all
 
 ## Supported VSS versions
 
+* [VSS 4.0](https://github.com/COVESA/vehicle_signal_specification/releases/tag/v4.0rc0)
```
erikbosch, line 12: Needs to be updated when official release tagged

.github/workflows/kuksa_databroker_build.yml
```diff
@@ -32,6 +32,8 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
+      - name: Query sites
```
erikbosch, line 35: Seems to be needed - otherwise builds fail

SebastianSchildt: "Query sites" is a weird "name"

the "update" might be needed if the base iage for ubuntu:latest at GH is either so old, that referenced packages in the apt database don't exist anymore, of there is not apt db at all

I am currently not sure if the run stuff is executed in a shell, if it is, might be better to to change the run line of "installing JVM" to

```
run: sudo apt update && sudo apt-get install -y default-jre
```

erikbosch: The existing list of package information seemed to be outdated in previous builds, see e.g. https://github.com/eclipse/kuksa.val/actions/runs/5053685693/jobs/9067766790

I will put it within the same run

SebastianSchildt (changes requested):
The databroker container still seems to come up with VSS 3.1.1 as a default?

I think becasue this https://github.com/eclipse/kuksa.val/blob/06cdbeef96d156f39ea814047a8d129cbea52a85/kuksa_databroker/Dockerfile#L64 is not changed

I suggest changing it to 4.0 

(and maybe for convenience also still copying the 3.1.1, that would allow users to go back without needing to mount something. We might kick 3.1.1 out once 5 is in)

wdyt?

erikbosch:
Added 4.0 to Dockercontainer, kept 3.1.1. Reason why I missed that file was that I searched all `*.*` files for 3.1.1 and then I off course missed `Dockerfile` as it does not contain a dot. I have no strong opinion if we should keep old versions in docker containers or not. It could simplify life for PoC-setups.

SebastianSchildt (approved):
Work for me 🛖

PR #554, opened 2023-05-11
Add authorization to GO library

lukasmittag:


kuksa_go_client/README.md
```diff
@@ -64,4 +64,4 @@ if you encounter a problem, you have to give the protoc executable the right to
 
 ### Configuration of the KUKSA.val GO client
 For Configuration you can use the kuksa-client.json file in this directory or provide the parameters through the command line.
-If you want the GO client to work with the KUKSA.val server specify protocol as ws if you want to use the KUKSA.val databroker set protocol to grpc. You can specify the port to look for an instance of KUKSA.val either in the JSON file or through -port on the command line.
\ No newline at end of file
+If you want the GO client to work with the KUKSA.val server specify protocol as ws if you want to use the KUKSA.val databroker set protocol to grpc. On command line it's available thorugh -protocol ws/grpc.
```
erikbosch, line 67: Not "thorugh" I assume? The first sentence could also benefit from rephrasing, or a dot/period after `as ws`

kuksa_go_client/main.go
```diff
@@ -22,105 +22,101 @@ import (
...
+if *protocol == "undefined" {
+	protocol = &configKuksaClient.TransportProtocol
+}
+
+var backend kuksa_client.KuksaBackend
+
+var token string
+if *protocol == "ws" {
+	// example token from kuksa.val/kuksa_certificates/jwt/all-read-write.json.token
+	token = "eyJ0eXAiOiJKV1QiLCJhbGciOiJSUzI1NiJ9.eyJzdWIiOiJrdWtzYS52YWwiLCJpc3MiOiJFY2xpcHNlIEtVS1NBIERldiIsImFkbWluIjp0cnVlLCJpYXQiOjE1MTYyMzkwMjIsImV4cCI6MTc2NzIyNTU5OSwia3Vrc2EtdnNzIjp7IioiOiJydyJ9fQ.QQcVR0RuRJIoasPXYsMGZhdvhLjUalk4GcRaxhh3-0_j3CtVSZ0lTbv_Z3As5BfIYzaMlwUzFGvCVOq2MXVjRK81XOAZ6wIsyKOxva16zjbZryr2V_m3yZ4twI3CPEzJch11_qnhInirHltej-tGg6ySfLaTYeAkw4xYGwENMBBhN5t9odANpScZP_xx5bNfwdW1so6FkV1WhpKlCywoxk_vYZxo187d89bbiu-xOZUa5D-ycFkd1-1rjPXLGE_g5bc4jcQBvNBc-5FDbvt4aJlTQqjpdeppxhxn_gjkPGIAacYDI7szOLC-WYajTStbksUju1iQCyli11kPx0E66me_ZVwOX07f1lRF6D2brWm1LcMAHM3bQUK0LuyVwWPxld64uSAEsvSKsRyJERc7nZUgLf7COnUrrkxgIUNjukbdT2JVN_I-3l3b4YXg6JVD7Y5g0QYBKgXEFpZrDbBVhzo7PXPAhJD6-c3DcUQyRZExbrnFV56RwWuExphw8lYnbMvxPWImiVmB9nRVgFKD0TYaw1sidPSSlZt8Uw34VZzHWIZQAQY0BMjR33fefg42XQ1YzIwPmDx4GYXLl7HNIIVbsRsibKaJnf49mz2qnLC1K272zXSPljO11Ke1MNnsnKyUH7mcwEs9nhTsnMgEOx_TyMLRYo-VEHBDLuEOiBo"
```
erikbosch, line 44: The hard-coded token here means that we practically ignore what has been defined in https://github.com/eclipse/kuksa.val/blob/master/kuksa_go_client/kuksa-client.json right? As we only use it as backup? Or?

lukasmittag: Yes, but since these are some examples how to use the go client, I thought this would be the best way to showcase it.

kuksa_go_client/kuksa_client/grpc.go
```diff
@@ -304,8 +311,30 @@ func (cg *KuksaClientCommGrpc) UnsubscribeFromKuksaVal(id string) error {
 }
 
 // Function to authorize to kuksa.val server
-func (cg *KuksaClientCommGrpc) AuthorizeKuksaValConn() error {
-	log.Println("KUKSA.val databroker does not support authorization mechanism yet. If you want to use authorization please use KUKSA.val server")
+func (cg *KuksaClientCommGrpc) AuthorizeKuksaValConn(TokenOrTokenfile string) error {
+	var tokenString string
+	if TokenOrTokenfile == "" {
+		if cg.Config.TokenOrTokenfile != ""{
```
erikbosch, line 317: Would it be an idea to reverse the logic here

if cg.Config.TokenOrTokenfile != ""
  <use the file from config file>
else
  <use the hardcoded string from TokenOrTokenFile>

Then we have no "dead" code, and one could actually use the config in https://github.com/eclipse/kuksa.val/blob/master/kuksa_go_client/kuksa-client.json (which by the way needs a changed keyword for "token", and also another port if the default is for a kuksa-val server connection (port 8090)

erikbosch:
I have verified:

- Running towards server and databroker with hardcoded tokens
- Running with token file specified in config file (after removing hardcoded token)

erikbosch (approved):
I have reviewed the original changes and have tested towards both databroker and server.
As Lukas is on vacation I have performed the updates I requested.

Tests performed for both Broker and Server:

- Using hardcoded token from code
- Using token file specified in config file
- Checking that you get an error (although not always easy to understand if tokenfile in config does not exist

erikbosch:
@SebastianSchildt - As I did some updates on the PR from Lukas I think it would be good if you could do a sanity check, and if ok merge

SebastianSchildt (approved):
Did some basic checks, runs for me as well

PR #555, opened 2023-05-11
Adapting SPDX checks for Proto files

erikbosch:
Making sure all Proto files are licensed with Apache-2.0 Adapting SPDX pattern to accept multi line comments when searching for SPDX identifiers


SebastianSchildt (approved):
lgtm 🦅

PR #559, opened 2023-05-12
Removing redundant license files and fixing links

erikbosch:
Contains two changes:

- License files in subdirectories removed, they are anyway covered by top LICENSE with same content
- Links checked with `markdown-link-check` and corrected where needed.


SebastianSchildt:
I see the Notice file seems to be outdated in general, and sort of superseeded by the BOMs at least for the RUST parts. I do spot some dependencies for the old appstore in there (Vaadin, Springboot). But for now let's fix the links as done here, and sort it out once we split this repo a bit more.

The NOTICE, if we still want to keep it, might become more or less a snapshot of DASH output

erikbosch:
> I see the Notice file seems to be outdated in general, and sort of superseeded by the BOMs at least for the RUST parts. I do spot some dependencies for the old appstore in there (Vaadin, Springboot). But for now let's fix the links as done here, and sort it out once we split this repo a bit more.
> 
> The NOTICE, if we still want to keep it, might become more or less a snapshot of DASH output

Yes - specifying NOTICE/BOM with exact versions is cumbersome as we often do not pin a specific version, but rather use latest and greatest so exact version will change often, and having a manually maintained file then is a nightmare.

SebastianSchildt (approved):
Better than before. lgtm

Issue #490, opened 2023-02-27
Target values can be set for everything


nayakned:
With kuksa-databroker, target values can be set for all datatpoints including SENSOR, ATTRIBUTE.

![VIN](https://user-images.githubusercontent.com/69036497/221547897-1d4f0b43-a6fe-4414-9583-b283d7d70be0.png)

While this makes sense on ACTUATOR data points, should this be allowed for SENSOR and ATTRIBUTE datapoints.
Currently, the returned JSON has an "actuator_target" as the key. Atleast this should be renamed to sensor_target and attribute_target if this is indeed the requirement.

![accl](https://user-images.githubusercontent.com/69036497/221548427-b6533a38-2718-4f9e-81ba-1068149e4c20.png)

argerus:
That's definitely a bug. Currently it's enforced in the `sdv.databroker.v1` API (i.e. it's only possible to set actuators), but it should really be enforced in the `core` instead so that it's never possible to set non-actuators.

SebastianSchildt:
Assigned @argerus , but maybe @lukasmittag it is also easy/doable for you. Wasn't for me just now. This bug just annoyed me. It should be gone. (I would need to look deeper, and while writing this in train, my cargo does not even manage to download the missing crates for compiling :-/ )

SebastianSchildt:
Solved

PR #565, opened 2023-05-24
Prevent cli from modifying attributes and sensors in databroker

rafaeling:
Related issue https://github.com/eclipse/kuksa.val/issues/490


argerus (changes requested):
Setting the values of attributes and sensors (in addition to actuators) should still be possible.

Setting _actuator target_ though, should only be possible on actuators.

PR #566, opened 2023-05-24
Fix/databroker cli array support

lukasmittag:
Fixes handling of comma in values and usage of double quotes and escaped double quotes


SebastianSchildt (changes requested):
I think something is wrong with the rebase/commit. there is far to much stuff in the commit (see "files changed"

SebastianSchildt (approved):
Works 🏖️

PR #567, opened 2023-05-24
Only actuactor target value can be set

rafaeling:
Related issue: https://github.com/eclipse/kuksa.val/issues/490

### DIFF
--- a/kuksa_databroker/databroker/src/grpc/kuksa_val_v1/val.rs
+++ b/kuksa_databroker/databroker/src/grpc/kuksa_val_v1/val.rs
@@ -148,6 +148,16 @@ impl proto::val_server::Val for broker::DataBroker {
                                 |id| proto::Field::from_i32(*id), // Ignore unknown fields for now
                             ));
 
+                        if entry.actuator_target.is_some() {
+                            if let Some(metadata) = broker.get_metadata(id).await {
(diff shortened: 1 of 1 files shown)


kuksa_databroker/databroker/src/grpc/kuksa_val_v1/val.rs
```diff
@@ -148,6 +148,16 @@ impl proto::val_server::Val for broker::DataBroker {
                                 |id| proto::Field::from_i32(*id), // Ignore unknown fields for now
                             ));
 
+                        if entry.actuator_target.is_some(){
+                            if let Some(metadata) = broker.get_metadata(id).await {
+                                if metadata.entry_type != broker::EntryType::Actuator {
+                                    return Err(tonic::Status::invalid_argument(
+                                        "Target value is not an actuator. Only Actuactor target EntryType can be set".to_string(),
```
lukasmittag, line 155: I think here would be better:
```suggestion
                                        "You tried setting a target value for a non-actuator. Non-actuators have no target value".to_string(),
```

lukasmittag (approved):
I tried setting target values for an actuator and a non-actuator. Worked. Just a small comment to the message displayed if you want to set the target value of a non-actuator.

PR #571, opened 2023-05-26
Upgrade to Vergen 8 and update dependencies

SebastianSchildt:
To fix a security issue in libgit, vergen needed to be updated. I upgraded to 8, and updated dependencies


erikbosch (approved):
I did a very basic sanity-check (building databroker, running it and testing a few commands with kuksa-client) and it worked without problems.

PR #572, opened 2023-05-30
Removing devcontainer support

### LINKED ISSUE #569: Update or Delete DevContainer
We have some files related to devcontainer in https://github.com/eclipse/kuksa.val/tree/master/.devcontainer
There is a little documentation in https://github.com/eclipse/kuksa.val/blob/9a4fe306036443012273f8bd622a8b05baa3706e/kuksa-val-server/README.md?plain=1#L61

We should better decide

- If we want to keep the DevContainer support
- If not, remove it
- If yes, add better documentation on how to actually use it and how it can be tested

Related to #355

erikbosch:
Devcontainer support is outdated and we are currently not testing/maintaining it. We are not aware of any users.
This PR suggests removing support.

Fixes #569


lukasmittag (approved):
Once CI is green will merge :) But quick question will .devcontainer be removed too? Git diff shows not exactly if the folder will be deleted too.

PR #578, opened 2023-06-15
Update threaded API doc

### LINKED ISSUE #524: Document expected return values for kuksa-client methods, and possibly align them
**_Note - this issue concerns the "old" threaded kuksa-client API. That API may be considered deprecated and then improving that API may have low priority!_**

As of today it is not documented what return values to expected when using kuksa-client APIs, like in the [example](https://github.com/eclipse/kuksa.val/blob/master/kuksa-client/kuksa_client/__init__.py#L70) below.

```
    # Set value to a given path
    def setValue(self, path, value, attribute="value", timeout=5):
        return self.backend.setValue(path, value, attribute, timeout)
```

The actual response varies a lot depending on which "backend" you use, making it difficult to implement correct handling in the feeder/provider using the client, some examples of return values

Successful (ws):
```

{
  "action": "set",
  "requestId": "528137c6-92dd-4dba-b5c7-0fc403e5b683",
  "ts": "2023-03-30T12:01:33.1680174093Z"
}
```

Server down (ws):

```
{
  "action": "set",
  "path": "Vehicle.Powertrain.Transmission.CurrentGear",
  "attribute": "value",
  "value": "0",
  "requestId": "85d5d056-6b75-4ed6-aa0a-ea313f4f8896",
  "error": "timeout"
}
```

Successful(grpc)

`OK`

 Server down (grpc):

```
{
    "error": {
        "code": 14,
        "reason": "unavailable",
        "message": "failed to connect to all addresses; last error: UNKNOWN: ipv4:127.0.0.1:55555: Failed to connect to remote host: Connection refused"
    },
    "errors": []
}

```

Here parsing of the result would be simplified if we always used the same return format, like always returning JSON, and always have a field error in the json on failure. I.e. instead of `OK` return `{"result":"ok"}`or similar.

erikbosch:
Fixes #524

### DIFF
--- a/kuksa-client/README.md
+++ b/kuksa-client/README.md
@@ -72,17 +72,29 @@ or you can also use following command inside `kuksa-client` to find the via `pip
 ```console
 Test Client> printTokenDir
 ```
-Select one of the tokens and use the `authorize` command like below.
+Select one of the tokens and use the `authorize` command like below:
 
 ```console
 Test Client> authorize /some/path/kuksa_certificates/jwt/super-admin.json.token
 ```
 
+### Authorizing against KUKSA.val Databroker
+
+If connecting to Databroker the command `printTokenDir` is not much help as it shows the default token directories
+for KUKSA.val Server example tokens. If the KUKSA.val Databroker use default example tokens then one of the
+tokens in [../jwt](..//jwt) can be used, like in the example below:
+
+```console
+Test Client> authorize /some/path//jwt/provide-all.token
+```
+
+
 ## Usage Instructions
 
 Refer help for further information
+
 ```console
-VSS Client> help -v
+Test Client> help -v
 
 Documented commands (use 'help -v' for verbose/'help <topic>' for details):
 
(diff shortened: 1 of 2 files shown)


kuksa-client/README.md
```diff
@@ -33,7 +33,7 @@ If you wish to connect to a gRPC server e.g. `kuksa-databroker`, you should inst
 ```console
 $ kuksa-client --ip 127.0.0.1 --port 55555 --protocol grpc --insecure
 ```
-Note: `--insecure` is required because `kuksa-databroker` does not yet support encryption or authentication.
+Note: `--insecure` is required because `kuksa-databroker` does not yet support TLS encryption or authentication.
```
lukasmittag, line 36: this isn't true anymore or am I wrong. Databroker supports TLS but the client not at this time?

erikbosch: Correct, or at least not with the current default certificates. I am touching this file in https://github.com/eclipse/kuksa.val/pull/580/files#diff-9e56616746625998b6fde6fba9bd42bf7385758ccb26a51ef83462310f06123c as well, but it could better be rephrased here

lukasmittag (approved):
everything looks good here

PR #579, opened 2023-06-16
Update tonic

argerus:


erikbosch:
The failure on `test` is interesting. I have seen it before, now and then. I do not really know how we specify cargo/protoc-version to use. Is it possibly so that we use "latest and greatest" which often but not always is a nightly build. I retriggered the build.

`error: the option `Z` is only accepted on the nightly compiler`

erikbosch:
Works well locally for me, but I would prefer if we could solve the build regression on master so that build can pass here before merging

argerus:
Will be included in #580 instead

PR #580, opened 2023-06-16
TLS support in KUKSA.val Client

erikbosch:
This intends to provide a working solution for TLS usage with KUKSA.val Client.

Functional changes:

- Adding arguments so you can specify what certificates to use
- Trying to solve name verification (that host name and name in certificate must match). Sometimes is a problem when using  numeric IP, by that reason there is a work-around parameter to specify name to expect

Tests have so far been performed towards both Server and Databroker, and also with DBC Feeder.

There should hopefully be no changes introducing incompatibility, but there are things we need to agree on later:

- Shall Databroker start to use TLS by default?
- In which applications shall we use "example certs" by default? Could be OK for clients, but for server/databroker I would prefer that you need to make an active decision (command line argument, change config, env-var) to use the default certs. I.e. default behavior for Broker/Server shall be that you get an error "No CA file provided"


lukasmittag (changes requested):
What I did:
* ran against databroker (cargo run .. and docker confirmed)
* running against kuksa-val-server (docker run ..) throws error provided in comments

kuksa_databroker/databroker-cli/src/main.rs
```diff
@@ -116,8 +116,10 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
         let ca_cert = tonic::transport::Certificate::from_pem(pem);
 
         let tls_config = tonic::transport::ClientTlsConfig::new()
-            .ca_certificate(ca_cert)
-            .domain_name("Server");
+            .ca_certificate(ca_cert);
+            // With new tonic we do not need the hack of setting Server as domain name,
+            // it can resolve both localhost and 127.0.0.1 as subjectAltName
+            //.domain_name("Server");
```
argerus, line 122: While informative for this PR, the comment should probably not be kept in the code going forward (i.e. should not be here).

doc/tls.md
```diff
@@ -0,0 +1,88 @@
...
+## KUKSA.val databroker
+
+KUKSA.val Databroker supports TLS, but not mutual authentication.
+By default (for historical reasons) an insecure connection is used.
+In the future this might be change so that `--insecure` is required to start an insecure connection.
+
+To use a secure connection specify `--tls-cert`and `--tls-private-key`
+
+```
+~/kuksa.val/kuksa_databroker$ cargo run --bin databroker -- --metadata ../data/vss-core/vss_release_4.0.json --tls-cert ../kuksa_certificates/Server.pem --tls-private-key ../kuksa_certificates/Server.key
```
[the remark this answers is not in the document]
lukasmittag: I'd say it is the preferred one because it's easier for testing. And we mention the docker things in our quickstart. Butmentioning both wouldn't hurt.

kuksa_databroker/databroker-cli/src/main.rs
argerus: I can just close that PR as it would be just as good to include that it in this one (as it's TLS related anyway).

argerus: Closed the PR. Also, when you remove this comment, run `cargo fmt` as the build is failing because of faulty formatting (at these lines).

kuksa-client/README.md
```diff
@@ -73,6 +72,53 @@ Select one of the tokens and use the `authorize` command like below.
...
+
+```
+~/kuksa.val/kuksa-client$ kuksa-client
+```
+
+This corresponds to this call:
+
+```
+kuksa-client --ip 127.0.0.1 --port 8090 --protocol ws --cacertificate ./kuksa_certificates/CA.pem
+```
```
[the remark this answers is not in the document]
erikbosch: I tested by building the kuksa-val-server container locally and running it as

`docker run -it --rm -p 127.0.0.1:8090:8090 -e LOG_LEVEL=ALL kuksa-val:latest`

and connecting it to it with both 127.0.0.1 and localhost with kuksa-client and both connections worked.


```
(kuksa-client) erik@debian3:~/kuksa.val/kuksa-client$ kuksa-client --ip localhost --port 8090 --protocol ws --cacertificate ./kuksa_certificates/CA.pem
Welcome to Kuksa Client version 0.4.0a2.post36+git.556a8bbc.dirty
```

```
(kuksa-client) erik@debian3:~/kuksa.val/kuksa-client$ kuksa-client --ip 127.0.0.1 --port 8090 --protocol ws --cacertificate ./kuksa_certificates/CA.pem
Welcome to Kuksa Client version 0.4.0a2.post36+git.556a8bbc.dirty
```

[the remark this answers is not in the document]
lukasmittag: Tested it too with ```--tls-server-name``` Server it works fine.

doc/tls.md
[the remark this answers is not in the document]
SebastianSchildt: I sort of agree that "cargo run" is more a development kind of view (also important), but there should be an answer for people "just downloaded/pulled a release", which also would act as guidance how e.g. projects like Velocitas would do it, when they provision everthing as containers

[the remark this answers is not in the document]
erikbosch: I think this is part of a bigger discussion on what different deployment alternatives we support out of the box. In the future we might for instance release Databroker as a Debian/Apertis package as well and then we might need to have yet another alternative documented. I added a Docker example below.

erikbosch:
> > * Shall Databroker start to use TLS by default?
> >   My feeling is yes, but maybe "not yet", but also depends a little bit: At this point, how much more complicated would quickstart get?
> 
> > * In which applications shall we use "example certs" by default? Could be OK for clients, but for server/databroker I would prefer that you need to make an active decision (command line argument, change config, env-var) to use the default certs. I.e. default behavior for Broker/Server shall be that you get an error "No CA file provided"
> 
> That sounds like a good idea, so you would not just "enable TLS" and run into danger of (unintendedly) running with "factory certs". In the long run/making TLS mandatory question would be, how easy it can be get up and running (thinking quickstart again)

Taking Databroker as an example, in the future I would like to support only two alternatives to start it:

```
databroker --tls-cert /my/path/Server.pem --tls-private-key /my/path/Server.key
databroker --insecure
```

I.e. not having TLS mandatory, but if you want to run without TLS you must state it explicitly. The [quickstart](https://github.com/eclipse/kuksa.val/blob/master/doc/quickstart.md) does not need to be that much more complicated, just add `--insecure`, which I better should do as part of this PR anyway.

erikbosch:
@lukasmittag - please review when you are back from vacation

doc/tls.md
```diff
@@ -0,0 +1,111 @@
...
+as the [Dockerfile](../kuksa-val-server/docker/Dockerfile) specifies that the default shall be used.
+
+## KUKSA.val Client (command line)
+
+See [KUKSA.val Client Documentation](../kuksa-client/README.md).
+
+## KUKSA.val Client (library)
+
+Clients like [KUKSA.val CAN Feeder](https://github.com/eclipse/kuksa.val.feeders/tree/main/dbc2val)
+tht use KUKSA.val Client library must typically set the path to the root CA certificate.
```
lukasmittag, line 96: typo ```tht```

kuksa-client/README.md
```diff
@@ -82,12 +81,59 @@ Test Client> authorize /some/path/kuksa_certificates/jwt/super-admin.json.token
...
 for KUKSA.val Server example tokens. If the KUKSA.val Databroker use default example tokens then one of the
-tokens in [../jwt](..//jwt) can be used, like in the example below:
+tokens in [../jwt](../jwt) can be used, like in the example below:
 
 ```console
-Test Client> authorize /some/path//jwt/provide-all.token
+Test Client> authorize /some/path/jwt/provide-all.token
 ```
 
+## Using TLS
```
lukasmittag, line 90: structure wise comment: ```Using TLS``` and then ```Connecting to KUKSA.val Server``` as sub headline without having a Connecting to KUKSA.val Databroker is not the best way. Only my opinion and nothing 'big'

lukasmittag (approved):
What I've done again:
* Started kuksa-client and Databroker with tls set some datapoints and it worked
* Read through the documentation

Comments can be ignored just some typos

kuksa-client/README.md
erikbosch: Good point, will fix it as soon as I see that all builds pass

PR #582, opened 2023-06-19
Fix nightly

erikbosch:
Testing how we can get rid of the problem with -Z errors.

### DIFF
--- a/.github/workflows/kuksa_databroker_build.yml
+++ b/.github/workflows/kuksa_databroker_build.yml
@@ -78,31 +78,13 @@ jobs:
           crate: cargo-strip
           version: latest
 
-      - uses: actions-rs/install@v0.1
-        with:
-          crate: cargo2junit
-          version: latest
-
-      - name: Run Tests
+      - name: Show toolchain information
         working-directory: ${{github.workspace}}
         run: |
-          cargo test --all-targets -- -Z unstable-options --report-time --format json | cargo2junit > results.xml;
+          rustup toolchain list
+          cargo --version
 
-      # - name: Publish test results as PR comment
-      #   uses: EnricoMi/publish-unit-test-result-action@v1
-      #   if: github.event_name == 'pull_request'
-      #   with:
-      #     check_name: Test Results
-      #     github_token: ${{ secrets.GITHUB_TOKEN }}
-      #     files: results.xml
-
-      - name: Upload testing report
-        uses: actions/upload-artifact@v3
-        with:
-          name: Unit test results
-          path: results.xml
-
-      - name: Run code coverage
+      - name: Run tests and report code coverage
         run: |
           cargo tarpaulin -o Xml
 
--- a/kuksa_databroker/databroker/src/grpc/server.rs
+++ b/kuksa_databroker/databroker/src/grpc/server.rs
@@ -152,6 +152,7 @@ where
 pub async fn serve_with_incoming_shutdown<F>(
     stream: TcpListenerStream,
     broker: broker::DataBroker,
+    authorization: Authorization,
     signal: F,
 ) -> Result<(), Box<dyn std::error::Error>>
 where
@@ -162,13 +163,22 @@ where
     Server::builder()
         .http2_keepalive_interval(Some(Duration::from_secs(10)))
         .http2_keepalive_timeout(Some(Duration::from_secs(20)))
-        .add_service(sdv::databroker::v1::broker_server::BrokerServer::new(
+        .add_service(
+            sdv::databroker::v1::broker_server::BrokerServer::with_interceptor(
+                broker.clone(),
+                authorization.clone(),
+            ),
+        )
+        .add_service(
+            sdv::databroker::v1::collector_server::CollectorServer::with_interceptor(
+                broker.clone(),
+                authorization.clone(),
+            ),
+        )
+        .add_service(kuksa::val::v1::val_server::ValServer::with_interceptor(
             broker.clone(),
+            authorization,
         ))
-        .add_service(sdv::databroker::v1::collector_server::CollectorServer::new(
-            broker.clone(),
-        ))
-        .add_service(kuksa::val::v1::val_server::ValServer::new(broker.clone()))
         .serve_with_incoming_shutdown(stream, shutdown(broker, signal))
         .await?;
 
--- a/kuksa_databroker/databroker/tests/current_values.rs
+++ b/kuksa_databroker/databroker/tests/current_values.rs
@@ -1,10 +1,23 @@
+/********************************************************************************
+ * Copyright (c) 2023 Contributors to the Eclipse Foundation
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Apache License 2.0 which is available at
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ ********************************************************************************/
+
 use core::panic;
 use std::{future, time::SystemTime, vec};
 
 use cucumber::{cli, given, then, when, writer, World as _};
 use databroker_proto::kuksa::val::v1::{
-    datapoint::Value, DataEntry, DataType, Datapoint, EntryRequest, EntryType, EntryUpdate, Field,
-    GetRequest, Metadata, SetRequest, View,
+    datapoint::Value, DataEntry, DataType, Datapoint, EntryRequest, EntryUpdate, Field, GetRequest,
+    SetRequest, View,
 };
 use tracing::debug;
 use world::DataBrokerWorld;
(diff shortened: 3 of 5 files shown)


.github/workflows/kuksa_databroker_build.yml
```diff
@@ -63,6 +63,11 @@ jobs:
       - name: Install Protoc
         run: sudo apt-get install -y protobuf-compiler
       - uses: actions/checkout@v3
+      - name: Install minimal nightly
```
SebastianSchildt, line 66: Is it a really a good idea to REQUIRE _nightly_ toolchains to buold databroker (if  I understand correctly)

If so, what is our really, really good reason for it?

erikbosch: If I understand things correctly.

- This installs the nightly toolchain, but the nightly toolchain is only used if requested
- We need the nightly toolchain for our test, at least as long as we want to use `--report-time` which is an unstable option and thus (most of the time?) only available in nightly tool-chain. I actually do not know why this test has worked now and then in the past. If "default" sometimes have been nighlty anyway? I actually do not know what controls what cargo version we use for building, i.e. how that is controlled
- The nightly will only be used when we use `+nightly`

I will add some printouts to check if my assumptions above seems to be true

erikbosch: Did some tests removing `--report-time` and the json report format (which also is experimental). The problem then is that it seems difficult to capture the test result. On failure exit code is still 0, and it is not as easy as searching for "failure" or "error" in stdout it seems. So in short, verifying test results seems easier if using nightly. The main build of databroker anyway happen inside the docker container, so whatever we set in the *.yml file should not matter.

kuksa_databroker/databroker/src/grpc/server.rs
```diff
@@ -162,13 +163,22 @@ where
...
+        )
+        .add_service(
+            sdv::databroker::v1::collector_server::CollectorServer::with_interceptor(
+                broker.clone(),
+                authorization.clone(),
+            ),
+        )
+        .add_service(kuksa::val::v1::val_server::ValServer::with_interceptor(
             broker.clone(),
+            authorization,
```
erikbosch, line 180: @argerus - don't we need clone here as well?

kuksa_databroker/databroker/tests/current_values.rs
```diff
@@ -158,7 +167,7 @@ fn assert_set_request_failure(w: &mut DataBrokerWorld, path: String, expected_er
 
 #[tokio::main]
 async fn main() {
-    databroker::init_logging();
+    // databroker::init_logging();
```
erikbosch, line 170: @argerus - shall we better keep or remove it?

kuksa_databroker/databroker/src/grpc/server.rs
argerus: No, we don't use it (further) after this point.

(and if `clone()` was actually needed, it wouldn't compile)

.github/workflows/kuksa_databroker_build.yml
argerus: Yes, nightly is only used for running the tests (not building databroker), and as Erik mentions it's needed if we want these output files.

Originally, this output was used to create a Github comment in the corresponding PR with a fancy test summary.

argerus:
> If we want to refactor test cases to not use experimental features

I don't think the test cases use any experimental features. Only the test _reporting_.

.github/workflows/kuksa_databroker_build.yml
SebastianSchildt: I see that this does not influence the "container builds", but doing this, are we not essentially also _building_ databroker used for the tests with a nightly toolchain? I.e. what we test is build with a different toolchain than what we release (I do understand, this was not perfectly in sync before, except for the integration tests, but now it is nightly vs. stable).

If this is the only (simple) way, maybe we need to live with it for now. Just want to confirm whether my understanding is correct. (and why did this even break now?)

erikbosch: My theory so far:

- In https://github.com/eclipse/kuksa.val/blob/master/.github/workflows/kuksa_databroker_build.yml we have a lot of references to `version: latest`. This I suspect gets "latest master", which often is a "nightly" version, but sometimes a released version. Which could explain why we only sometimes gets an error for `-Z`

If you look at old "successful" builds like https://github.com/eclipse/kuksa.val/actions/runs/5120892422/jobs/9207950580  you can see that they anyway was not successful, there are errors like:

```
 error: the following required arguments were not provided:
  --test-threads <TEST_THREADS>
```

(but they did not cause the build to fail which is a bit strange IMHO)

As I said before, we could remove the dependency and instead try to do some greps on the output of the tests and that way try to figure out if there are any errors. Just let me know if you think that is the preferred approach.

argerus: `cargo test` does return non-successful return codes when a test fails. At least for the "normal" tests.

So, I think the issue is specific to the cucumber tests and it's runner.
```rust
#[tokio::main]
async fn main() {
    // databroker::init_logging();

    let opts = cli::Opts::<_, _, _, world::UnsupportedLibtestArgs>::parsed();

    DataBrokerWorld::cucumber()
        .with_writer(writer::Libtest::or_basic())
        .with_cli(opts)
        .after(|_feature, _rule, _scenario, _ev, world| {
            if let Some(w) = world {
                w.stop_databroker();
            }
            Box::pin(future::ready(()))
        })
        .run("tests/features/current_values.feature")
        .await;
}
```

So I think the preferred solution is to modify this runner to return non-success on test failures.

Separate from that, we might at some point want to report the results of our unit tests  in some fancy way (again), which might require unstable features. But for simple success / failure, I don't think it's needed.

SebastianSchildt: Do we know how to achieve this in the cucumber runner?  Maybe @sophokles73  knows or @rafaeling has looked into the cucumber magic already?

I agree with @erikbosch  we should repair CI fast, but if there is a "similarly" easy way to achieve this without adding deps on "nightly" toolchain (that just "feels" bad to me) I would prefer that

In that case we could revisit "fancy" reporting later, but since we as of today are not using it anyway we wouldn't loose anything...

argerus: Replacing `run` with `run_and_exit` seems to do the trick..
```diff
diff --git a/kuksa_databroker/databroker/tests/current_values.rs b/kuksa_databroker/databroker/tests/current_values.rs
index a174f37..288e5d8 100644
--- a/kuksa_databroker/databroker/tests/current_values.rs
+++ b/kuksa_databroker/databroker/tests/current_values.rs
@@ -14,7 +14,7 @@
 use core::panic;
 use std::{future, time::SystemTime, vec};
 
-use cucumber::{cli, given, then, when, writer, World as _};
+use cucumber::{given, then, when, writer, World as _};
 use databroker_proto::kuksa::val::v1::{
     datapoint::Value, DataEntry, DataType, Datapoint, EntryRequest, EntryUpdate, Field, GetRequest,
     SetRequest, View,
@@ -169,17 +169,14 @@ fn assert_set_request_failure(w: &mut DataBrokerWorld, path: String, expected_er
 async fn main() {
     // databroker::init_logging();
 
-    let opts = cli::Opts::<_, _, _, world::UnsupportedLibtestArgs>::parsed();
-
     DataBrokerWorld::cucumber()
         .with_writer(writer::Libtest::or_basic())
-        .with_cli(opts)
         .after(|_feature, _rule, _scenario, _ev, world| {
             if let Some(w) = world {
                 w.stop_databroker();
             }
             Box::pin(future::ready(()))
         })
-        .run("tests/features/current_values.feature")
+        .run_and_exit("tests/features/current_values.feature")
         .await;
 }
```

kuksa_databroker/databroker/tests/features/current_values.feature
```diff
@@ -35,7 +35,7 @@ Feature: Writing and reading the current value of a VSS Data Entry
         | Vehicle.TraveledDistanceHighRes        | uint64 | 23425462462563924 |
         | Vehicle.CurrentLocation.Longitude      | double |        145.023544 |
         | Vehicle.Speed                          | float  |              45.5 |
-        | Vehicle.IsMoving                       | bool   |              true |
+        | Vehicle.IsMogving                       | bool   |              true |
```
erikbosch, line 38: Deliberate error to check if it makes build fail, if working I will remove it in a later update

erikbosch:
@SebastianSchildt @argerus 

Now the dependency to experimental features are gone.
I actually removed the whole "cargo test" section as the tests are included as well when you do the tarpaulin part to get coverage. I verified that the build fails if there is an error in the cucumber tests.

.github/workflows/kuksa_databroker_build.yml
```diff
@@ -78,31 +78,13 @@ jobs:
...
-      - uses: actions-rs/install@v0.1
-        with:
-          crate: cargo2junit
-          version: latest
-
-      - name: Run Tests
+      - name: Show toolchain information
         working-directory: ${{github.workspace}}
         run: |
-          cargo test --all-targets -- -Z unstable-options --report-time --format json | cargo2junit > results.xml;
```
erikbosch, line 89: If we remove the "unstable part" we cannot do the result upload part. less important as I see it as the log anyway is visible in stdout.
But if we do not have the report part there is no need to have it as tests anyway are run by "cargo tarpaulin" below

SebastianSchildt (approved):
looking good to me

Issue #557, opened 2023-05-12
Protoc compiler has to be installed manually before building the repository


rafaeling:
Documentation does not include Protoc compiler as prerequisite to build the repository:

https://github.com/eclipse/kuksa.val/blob/master/kuksa_databroker/README.md

SebastianSchildt:
Question: Do we want to "fix" it in documentation, or do we want to do it during build again?

argerus:
There is a rather simple fix available. This will build protoc from source (as part of the `protobuf-src` crate). This will be cached locally and wouldn't affect build times with the exception of the first time it's built on a machine.

The only downside is that this (currently) wouldn't be cached by the CI, making it slower. Could still be worth the convenience.


```diff
diff --git a/kuksa_databroker/databroker-proto/Cargo.toml b/kuksa_databroker/databroker-proto/Cargo.toml
index 88d4000..3289efb 100644
--- a/kuksa_databroker/databroker-proto/Cargo.toml
+++ b/kuksa_databroker/databroker-proto/Cargo.toml
@@ -24,4 +24,5 @@ prost = { workspace = true }
 prost-types = { workspace = true }
 
 [build-dependencies]
+protobuf-src = "1.1.0"
 tonic-build = { workspace = true, features = ["transport", "prost"] }
diff --git a/kuksa_databroker/databroker-proto/build.rs b/kuksa_databroker/databroker-proto/build.rs
index d3a70ac..d02a006 100644
--- a/kuksa_databroker/databroker-proto/build.rs
+++ b/kuksa_databroker/databroker-proto/build.rs
@@ -12,6 +12,7 @@
 ********************************************************************************/
 
 fn main() -> Result<(), Box<dyn std::error::Error>> {
+    std::env::set_var("PROTOC", protobuf_src::protoc());
     tonic_build::configure()
         .compile_well_known_types(false)
         .protoc_arg("--experimental_allow_proto3_optional")

```

PR #585, opened 2023-06-27
Pre-build protoc from source

rafaeling:
It closes this issue: https://github.com/eclipse/kuksa.val/issues/557


argerus (changes requested):
This should mean we don't have any dependency on protoc.

We should probably remove the "install protoc" steps from the github actions, as they take time.

argerus (approved):

Issue #553, opened 2023-05-10
subscribe on non-existing path gives wrong error code


SebastianSchildt:
I only observed this python side, but I do think it is caused by databroker.

When doing a "get" on `Vehicle.Notexist` the returned error is

```
{"error": {"number": 404, "reason": "not_found", "message": "Path not found"}}
```

When doing subscribe on `Vehicle.Notexist` the returned error is

```
{"error": {"number": 3, "reason": "invalid argument", "message": "NotFound"}}
```

In line with viss-style error codes I would expect the same error here (404 with reason not_found

PR #586, opened 2023-06-28
Fix wrong error code when subscribing to a non-existing path

rafaeling:
This closes https://github.com/eclipse/kuksa.val/issues/553

Since SubscribeResponse does not return a protobuf error, the current approach is to use a tonic error as the return ErrorCode. This means that the error code returned will not be 404, but rather the tonic error code. However, it might be necessary to modify the SubscribeResponse to include DataEntryError as well.

### DIFF
--- a/kuksa_databroker/databroker/src/grpc/kuksa_val_v1/val.rs
+++ b/kuksa_databroker/databroker/src/grpc/kuksa_val_v1/val.rs
@@ -24,6 +24,7 @@ use tracing::debug;
 
 use crate::broker;
 use crate::broker::ReadError;
+use crate::broker::SubscriptionError;
 use crate::permissions::Permissions;
 
 #[tonic::async_trait]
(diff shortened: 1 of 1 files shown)


kuksa_databroker/databroker/src/grpc/kuksa_val_v1/val.rs
```diff
@@ -326,10 +327,16 @@ impl proto::val_server::Val for broker::DataBroker {
                 let stream = convert_to_proto_stream(stream);
                 Ok(tonic::Response::new(Box::pin(stream)))
             }
-            Err(e) => Err(tonic::Status::new(
+            Err(SubscriptionError::NotFound) => {
+                Err(tonic::Status::new(tonic::Code::NotFound, "Path not found"))
```
lukasmittag, line 331: Couldn't you use 
```
Err(ReadError::NotFound) => {
                        errors.push(proto::DataEntryError {
                            path: request.path,
                            error: Some(proto::Error {
                                code: 404,
                                reason: "not_found".to_owned(),
                                message: "Path not found".to_owned(),
                            }),
                        });
                    } 
```
and exchange ``` ReadError::NotFound ``` with ``` SubscriptionError::NotFound ``` Not an expert but what I think could be done. But as erik mentioned we need to agree which error codes we want. Personally I like the 404 and so on better. But I can see that grpc codes could be good for debugging and more understanding the error without having a look to deep in the code.

rafaeling: For other methods like 'get,' it would be feasible as you suggest because the proto response contains a 'proto::DataEntryError' vector. However, in the case of the subscribe method, it is not possible because the response does not contain such a field and any similar proto error.
I also agree with you and Erik; I prefer using 404 and making improvements in the error information. However, with the current proto schema, it is not possible to fix it in any other way.

erikbosch: It is part of the bigger decision - do we prefer to use gRPC error codes or our own. Even if we would define our own error solution in proto files (like we partially have today) clients would anyway need to check the gRPC error code first, as there are a lot of error codes that can be generated by the framework (see https://grpc.github.io/grpc/core/md_doc_statuscodes.html) which we cannot control

erikbosch:
@argerus - do you want to take a look at this one as well?

PR #590, opened 2023-06-30
Fixing some linter issues

erikbosch:
I tried running mypy, pylint and flake8.
Nothing alarming found but many false positives,
so we are not mature to enable them in CI.
This addresses some low hanging fruits.

### DIFF
--- a/kuksa-client/kuksa_client/cli_backend/grpc.py
+++ b/kuksa-client/kuksa_client/cli_backend/grpc.py
@@ -25,6 +25,7 @@ from typing import Any
 from typing import Callable
 from typing import Dict
 from typing import Iterable
+from typing import Optional
 import uuid
 import os
 import re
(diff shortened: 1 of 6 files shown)


kuksa-client/setup.py
```diff
@@ -1,3 +1,15 @@
+# /********************************************************************************
```
lukasmittag, line 1: I thought in this files we do not need a header @SebastianSchildt ?

kuksa-client/kuksa_client/cli_backend/grpc.py
```diff
@@ -144,12 +145,12 @@ def setValues(self, updates: Dict[str, Any], attribute="value", timeout=5):
...
 
     # Function for authorization
-    def authorize(self, token_or_tokenfile:str =None, timeout=5):
+    def authorize(self, token_or_tokenfile:Optional[str] =None, timeout=5):
         if token_or_tokenfile is None:
             token_or_tokenfile = self.token_or_tokenfile
         if os.path.isfile(token_or_tokenfile):
-            token_or_tokenfile = pathlib.Path(token_or_tokenfile)
-            token = token_or_tokenfile.expanduser().read_text(encoding='utf-8').rstrip('\n')
+            token_or_tokenfile_path = pathlib.Path(token_or_tokenfile)
```
lukasmittag, line 152: why do we need this renaming here?

lukasmittag (approved):
What I did is:
* Ran against databroker without authorization enabled (set, get, subscribe)
* Ran against databroker with authorization enabled (set, get, subscribe, authorize)
Worked

kuksa-client/kuksa_client/cli_backend/grpc.py
erikbosch: It is mypy that does not like when you change type of a variable. `token_or_tokenfile` was first a `str` and then we did an assignment changing it to `Path`. Now instead we use a different variable name.

_error: Incompatible types in assignment (expression has type "Path", variable has type "str")  [assignment]_

lukasmittag: Ah, now I see :)

kuksa-client/setup.py
erikbosch: I added the license as otherwise our own spdx license checker complains. As of today we require a license in all `*.py` files changed to make the checker happy. This could off course be changed, for example by explicitly excluding files called `setup.py`. But is it worth it?

lukasmittag: Yes, indeed. Yeah dont think it is worth it then.

PR #592, opened 2023-07-06
Update Go dependencies.

erikbosch:
Triggering reason is to fix a vulnerability in gRPC <1.53

I am by no means a go expert and I do not really know why the `go.sum` file becomes so much larger than before. but I have documented in the README how I updated the versions, and I have tested against both Databroker and Server that things seems to work like before.


lukasmittag (approved):
What I did:
* tested if it runs
I see that there are many changes in .sum. I think it is okay. It could be avoided by ```go get -u google.golang.org/grpc```. That would only update grpc dependencies. But I thinks since it does not break anything it is okay to update everything.

PR #596, opened 2023-07-10
Kuksa Python & Go client: Refactor array handling and add tests

erikbosch:
This is work in progress. Background is that code scanning showed that the current regexp may be inefficient for certain strings. To see if it could be simplified without causing side effects a number of test cases has been added, which also partially proves as documentation on how it is intended to work. A similar change is one for the go client.

### DIFF
--- a/kuksa-client/kuksa_client/grpc/__init__.py
+++ b/kuksa-client/kuksa_client/grpc/__init__.py
@@ -315,6 +315,57 @@ class Datapoint:
             ) if message.HasField('timestamp') else None,
         )
 
+
+    def cast_array_values(cast, array):
+        """
+        Parses array input and cast individual values to wanted type.
+        Note that input value to this function is not the same as given if you use kuksa-client command line
+        as parts (e.g. surrounding quotes) are removed by shell, and then do_setValue also do some magic.
+        """
+        array = array.strip('[]')
+
+        # Split the input string into separate values
+        # First alternative, not quotes including escaped single or double quote, ends at comma, whitespace or EOL
+        # Second group is double quoted string, may contain quoted strings and single quotes inside, ends at non-escaped
+        # double quote
+        # Third is similar but for single quote
+        # Using raw strings with surrounding single/double quotes to minimize need for escapes
+        pattern = r'(?:\\"|\\' + \
+                  r"'|[^'" + r'",])+|"(?:\\"|[^"])*"|' + \
+                  r"'(?:\\'|[^'])*'"
+        values = re.findall(pattern, array)
+        for item in values:
+            # We may in some cases match blanks, that is intended as we want to be able to write arguments like
+            # My Way
+            # ... without quotes
+            if item.strip() == '':
+                #skip
+                pass
+            else:
+                yield cast(item)
+
+    def cast_bool(value) -> bool:
+        if value in ('False', 'false', 'F', 'f'):
+            value = 0
+        return bool(value)
+
+    def cast_str(value) -> str:
+        """
+        Strip based on the following rules.
+        - Leading/Trailing blanks are removed
+        - If there are single or double quote at both start and end they are removed
+        - Finally any quote escapes are removed
+        """
+        new_val = value.strip()
+        if new_val.startswith('"') and new_val.endswith('"'):
+            new_val = new_val[1:-1]
+        if new_val.startswith("'") and new_val.endswith("'"):
+            new_val = new_val[1:-1]
+        # Replace escaped quotes with normal quotes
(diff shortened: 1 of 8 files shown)


kuksa-client/kuksa_client/grpc/__init__.py
```diff
@@ -315,6 +315,27 @@ def from_message(cls, message: types_pb2.Datapoint):
             ) if message.HasField('timestamp') else None,
         )
 
+
+    def cast_array_values(cast, array):
+    
+        array = array.strip('[]')
+        pattern = r'(?:\\"|[^",])+|"(?:\\"|[^"])*"'
```
lukasmittag, line 322: This will lead from 
```
setValue Vehicle.OBD.DTCList ["dtc1, dtc2", dtc3, \" dtc4, dtc4\"]
```
to 
```
{
    "path": "Vehicle.OBD.DTCList",
    "value": {
        "value": {
            "values": [
                "dtc1, dtc2",
                "dtc3",
                "\" dtc4",
                "dtc4\""
            ]
        },
        "timestamp": "2023-07-11T06:37:13.441133+00:00"
    }
}
```
the comma in \" will not be treated correct.

kuksa_go_client/kuksa_client/grpc.go
```diff
@@ -83,7 +83,7 @@ func getArrayFromInput[T any](input string) ([]T, error) {
 	input = strings.TrimSuffix(strings.TrimPrefix(input, "["), "]")
 	
 	// Split the input string into separate values
-	pattern := `(?:\\.|[^",])*"(?:\\.|[^"])*"|'(?:\\.|[^']|\\')*'|[^",]+`
+	pattern := `(?:\\"|[^",])+|"(?:\\"|[^"])*"|'(?:\\"|[^']|\\')*'`
```
lukasmittag, line 86: In GOit is only possible to set a certain value through for example
```
"['dtc1, dtc2', dtc2, \"dtc3, dtc3\"]"
```
and this will result in 
```
{
    "path": "Vehicle.OBD.DTCList",
    "value": {
        "value": {
            "values": [
                "dtc1",
                "dtc2",
                "dtc2",
                "",
                "\"dtc3, dtc3\""
            ]
        },
        "timestamp": "2023-07-11T06:41:14.537216+00:00"
    }
}
```

erikbosch: This seems odd, I will need to look into it a bit more. But then it is also so that it is a difference between how a string is set in code on what it actually will contain.

erikbosch: With latest PR version the string /from main.go)

`var valstr = "['dtc1, dtc2', dtc2, \"dtc3, dtc3\", dtc4]"`

Results in:

```
Test Client> getValue Vehicle.OBD.DTCList
{
    "path": "Vehicle.OBD.DTCList",
    "value": {
        "value": {
            "values": [
                "dtc1, dtc2",
                "dtc2",
                "dtc3, dtc3",
                "dtc4"
            ]
        },
        "timestamp": "2023-07-12T06:11:15.227045+00:00"
    }
}

```

kuksa-client/kuksa_client/grpc/__init__.py
erikbosch: I am adding some recommendations to the README. Main problem is that not surrounding the whole array with quotes leaves you to the special handling of the shell/argparse. Like in this case (blank before first double quote)

`setValue Vehicle.OBD.DTCList [ "dtc1, dtc2", dtc3, \" dtc4, dtc4\"]`

... which neither now or before give you what you want as the shell "eats" the quotes. We could possibly do something about it, like add quotes back but that gives us a lot of corner cases where we want to escape any quotes inside, but not quotes that already are escaped. So for now we need to live with that limitation. Better then to use surround everything with single quotes:

`setValue Vehicle.OBD.DTCList '[ "dtc1, dtc2", dtc3,   \" dtc4, dtc4\"]'`

Which gives you what i think is correct
```
Test Client> getValue Vehicle.OBD.DTCList                                           
{
    "path": "Vehicle.OBD.DTCList",
    "value": {
        "value": {
            "values": [
                "dtc1, dtc2",
                "dtc3",
                "\" dtc4",
                "dtc4\""
            ]
        },
        "timestamp": "2023-07-12T06:19:14.369683+00:00"
    }
}

```

lukasmittag (approved):
Didn't manage to run the tests. Maybe related to python 3.8 or something else. Also error is in conftest.py not newly added one. Other than that I did run go tests and go client looked good.
kuksa-client and README tested and tested a uint8Array works as well as expected.

PR #597, opened 2023-07-10
Add pre-commit checks

erikbosch:
This will introduce pre-commit checks for kuksa-val. We already have it for kuksa.val.feeders

What it will do:

- For most files check that there is a line-break at end-of-file
- For most files verify that you do not have trailing blanks
- For python verify that flake8 pass, checking for instance line lengths

Suggest approach is to have a pre-commit hook installed on your local machine, then the check will be performed before you commit and this check shall be guaranteed to pass. But you may work reactively as well and fix issues if build fails and upload a new version. It only checks files affected by the commit/PR, but it checks the whole file, so initially you may need to fix some old sins.

Caveat: There might be cases where we do not want files to be changed, I took some settings "inherited" from kuksa.val.feeders. If you would notice that the checker wants to change files that should not be changed, then we may need to tweak the setting.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,18 @@
+# See https://pre-commit.com for more information
+# See https://pre-commit.com/hooks.html for more hooks
+repos:
+-   repo: https://github.com/pre-commit/pre-commit-hooks
+    rev: v3.2.0
+    hooks:
+    -   id: trailing-whitespace
+        exclude_types: ["dbc"]
+    -   id: end-of-file-fixer
+        exclude_types: ["dbc", "json"]
+        exclude: \.token$
+    -   id: check-yaml
+    -   id: check-added-large-files
+
+-   repo: https://github.com/pycqa/flake8
+    rev: '6.0.0'
+    hooks:
+    -   id: flake8
--- a/.flake8
+++ b/.flake8
@@ -1,0 +1,2 @@
+[flake8]
+max_line_length = 120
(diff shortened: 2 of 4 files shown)


.pre-commit-config.yaml
```diff
@@ -0,0 +1,18 @@
+# See https://pre-commit.com for more information
+# See https://pre-commit.com/hooks.html for more hooks
+repos:
+-   repo: https://github.com/pre-commit/pre-commit-hooks
+    rev: v3.2.0
+    hooks:
+    -   id: trailing-whitespace
```
lukasmittag, line 7: should we add another hook for python linter? Then you do not need to wait until CI runs through and clippy for rust files maybe. Just a thought

lukasmittag (approved):
Minor comment

.pre-commit-config.yaml
erikbosch: flake8 and pylint has similar checks, but with some differences. At the moment we do not do any linter checks with pylint/mypy in CI, but in the long run we better should. In feeder we use pylint/mypy, but only fail on errors. I did by the way try mypy some time ago on kuksa-client but it gave some false positives as it could not really understand the gRPC integration correctly, so there are some tweaks needed if we are to use it.

It is possiblw to run pylint as part of pre-commit, see https://pylint.pycqa.org/en/latest/user_guide/installation/pre-commit-integration.html, but it requires that you do a local install. That is no problem for us working with Python but I am not sure if our rust/c++-guys want to do that, as it might be needed even if you have not changed any Python files.

I suggest we take extending/improving checks as a later step.

PR #600, opened 2023-07-13
Use default time (24h) for ttl images

### LINKED ISSUE #598: Prefferred ttl.sh image lifetime
As of today, in KUKSA.val and KUKSA.val.feeders we use very different settings for lifetime of the images. In some cases we do not specify and get default (24 hours), in other cases we specify 8 hours or 1 hour. Does anyone have any objection to that we use default 24 hours everywhere, which is also maximum allowed lifetime?

I sometimes feel that 8h/1h is too short, if you for instance build at end of work day and want to test next day. If no-one objects I may create PRs to do the change.

erikbosch:
Use default lifetime (24 hours) instead of 1h/8h, to be able to use e.g. a image build last night the following morning. 

Fixes #598

### DIFF
--- a/.github/workflows/kuksa-client.yml
+++ b/.github/workflows/kuksa-client.yml
@@ -92,7 +92,7 @@ jobs:
         context: .
         file: kuksa-client/Dockerfile
         push: false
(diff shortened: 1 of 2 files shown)


int0x27 (approved):
looks good, ttl.sh default is 24h

PR #602, opened 2023-07-17
Update status and author for kuksa-client

### LINKED ISSUE #599: What is our pypi author policy?
As of today we list Sebastian, Naresh and Wenwen in https://github.com/eclipse/kuksa.val/blob/master/kuksa-client/setup.cfg

This also gets visible in https://pypi.org/project/kuksa-client/.

Naresh and Wenwen however are not that active in the project. From a communication perspective a public address like `kuksa-dev@eclipse.org` seems more appropriate, and that is also what we use in https://github.com/eclipse/kuksa.val/blob/master/kuksa_certificates/genCerts.sh, but then author better should be "Eclipse KUKSA project" or similar.

I can see three possible policies:

- Historical persons that never should change. That is what we have today
- Listing persons actively working with the component. if so we should possible replace Naresh and Wenwen with Erik and Lukas
- Use a "community name/e-mail", like "Eclipse KUKSA project"

For practical perspectives the second or third option seems reasonable, and the third is easiest to maintain, but require that we have a working common mail address

erikbosch:
Fixes #599

An example on the result can be seen in pypi, see https://test.pypi.org/project/kuksa-client/0.4.0a6/

### DIFF
--- a/kuksa-client/setup.cfg
+++ b/kuksa-client/setup.cfg
@@ -1,8 +1,8 @@
 [metadata]
 name = kuksa_client
-author = Sebastian Schildt, Naresh Nayak, Wenwen Chen
-author_email = sebastian.schildt@de.bosch.com, naresh.nayak@de.bosch.com, wenwen.chen@de.bosch.com
-description = kuksa.val python clients and SDK
+author = Eclipse KUKSA Project
+author_email = kuksa-dev@eclipse.org
+description = KUKSA Python Client and SDK
 long_description = file:./README.md
 long_description_content_type = text/markdown
 url=https://github.com/eclipse/kuksa.val
@@ -11,7 +11,7 @@ project_urls=
   Bug Tracker=https://github.com/eclipse/kuksa.val/issues
 classifiers =
     Intended Audience :: Developers
-    Development Status :: 3 - Alpha
+    Development Status :: 5 - Production/Stable
     Environment :: Console
     Programming Language :: Python :: 3
     License :: OSI Approved :: Apache Software License
@@ -27,7 +27,6 @@ install_requires =
     websockets >= 10.1
     cmd2 >= 1.4, <2.0
     pygments >= 2.5
-    importlib_metadata >= 3.6 ; python_version < "3.8"
     grpcio-tools >= 1.54.2
     jsonpath-ng >= 1.5.3
 packages = find:
--- a/kuksa-client/README.md
+++ b/kuksa-client/README.md
@@ -1,19 +1,19 @@
-# KUKSA.val Client
+# KUKSA Python Client and SDK
 ![kuksa.val Logo](https://raw.githubusercontent.com/eclipse/kuksa.val/0.2.5/doc/pictures/logo.png)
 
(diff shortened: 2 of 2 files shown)


kuksa-client/setup.cfg
```diff
@@ -27,7 +27,6 @@ install_requires =
     websockets >= 10.1
     cmd2 >= 1.4, <2.0
     pygments >= 2.5
-    importlib_metadata >= 3.6 ; python_version < "3.8"
```
erikbosch, line 30: This line was useless as it only should be installed for Python versions below 3.8, but we require at least 3.8 so it will never trigger

kuksa-client/setup.cfg
```diff
@@ -11,7 +11,7 @@ project_urls=
   Bug Tracker=https://github.com/eclipse/kuksa.val/issues
 classifiers =
     Intended Audience :: Developers
-    Development Status :: 3 - Alpha
+    Development Status :: 5 - Production/Stable
```
erikbosch, line 14: This can be discussed, but I would say kuksa-client has passed alpha stage

https://pypi.org/classifiers/
https://martin-thoma.com/software-development-stages/

lukasmittag: would agree with production/stable

SebastianSchildt: Works for me. We might have a long argument about whether it is 4 or 5, but anyway, this is one of our few main/core components and it has been a long time around and is reasonably stable.

kuksa-client/setup.cfg
```diff
@@ -1,8 +1,8 @@
 [metadata]
 name = kuksa_client
-author = Sebastian Schildt, Naresh Nayak, Wenwen Chen
-author_email = sebastian.schildt@de.bosch.com, naresh.nayak@de.bosch.com, wenwen.chen@de.bosch.com
-description = kuksa.val python clients and SDK
+author = Eclipse KUKSA Project
+author_email = kuksa-dev@eclipse.org
+description = KUKSA.val Python Client and SDK
```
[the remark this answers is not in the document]
erikbosch: I have no problem dropping `.val` whenever used. Should I possibly update https://github.com/eclipse/kuksa.val/wiki/KUKSA.val-Naming-Guidelines to reflect that?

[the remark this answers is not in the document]
lukasmittag: Yes, if we drop it we should change it there too. https://github.com/eclipse/kuksa.val/wiki/KUKSA.val-Naming-Guidelines but htis link does not work for me :/

lukasmittag (approved):

kuksa-client/setup.cfg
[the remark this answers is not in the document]
erikbosch: > Yes, if we drop it we should change it there too. https://github.com/eclipse/kuksa.val/wiki/KUKSA.val-Naming-Guidelines but htis link does not work for me :/

That is because I changed the name to https://github.com/eclipse/kuksa.val/wiki/KUKSA-Naming-Guidelines

SebastianSchildt (approved):

PR #604, opened 2023-07-19
Replace actions-rs/install build action

erikbosch:
Old action is not maintained and uses Node.js 12, so will give us problems sooner or later

See https://github.com/actions-rs/toolchain/issues/216


SebastianSchildt (approved):
Looking good

Issue #411, opened 2022-11-22
Reintroduce autocompletion in Python command line client for GRPC


SebastianSchildt:
kuksa_(viss)_client used to support tab-automcompletion for VSS Paths in the VISS (websocket) side. This does not work in GRPC side.

For usability reasons (as the cmdline client really is about exploring/debugging) we should add this feature for GRPC connections.

It can probably not be a direct port of the VISS side, because that makes use of the KUKSA VISS getMetadata call (which basically gives you verbatim excerpts from the VSS json tree)

romainletendart:
I don't know how this should work with the unified API.
Since the switch to that new API, here is what the client gets back from databroker when querying a wildcard:
```console
Test Client> getMetaData *
{
    "error": {
        "code": 404,
        "reason": "not_found",
        "message": "Path '*' not found"
    },
    "errors": [
        {
            "path": "*",
            "error": {
                "code": 404,
                "reason": "not_found",
                "message": "Path '*' not found"
            }
        }
    ]
}
```
I think wildcards are only supported with the query syntax of the former databroker API.
@argerus any thoughts?

SebastianSchildt:
Do we have a specific way to query the childs of a node? Like a Metadata Field we can query?

lukasmittag:
One option could be to use the newest vss release json file. Would then not be backwards compatible though.

SebastianSchildt:
I think the only "good way" is, when this is done via API, e.g. requesting the children of any Node.

I do however not think that also adding the (functionlity-wise overlapping) sdv API is the solutions, and I think kuksa.val.v1 can not yet do it, it would at least need something liuke StringArray children in Metadata. Which is only one but  may or may not be the best way to do it

lukasmittag:
Is there a reason why not handling it like in sdv API?

argerus:
I think it would make sense to add a `List( "wildcard pattern", [fields] )` to the `kuksa.val.v1` API that would return all matching VSS entries. The syntax of "wildcard pattern" should be the same as what the authorization scope uses.

SebastianSchildt:
That makes sense to me, but thinking of authorization, would it returns everything where you possess a read scope? Would you "see" the existence of a Child if you have no access?

And what would we expect this to return. An array of all metadata?

rafaeling:
Here is the draft PR for autocompletion: https://github.com/eclipse/kuksa.val/pull/605

SebastianSchildt:
Done in #653

PR #605, opened 2023-07-20
Discoverabiltiy in kuksa.val.v1 API to enable e.g. Autocompletion

rafaeling:
Here is some draft implementation for discoverability in kuksa.val.v1 to address the autocompletion feature.

Since there are varying perspectives on how to implement the autocompletion feature for the client, this pull request serves as a preliminary draft for collaborative feedback from different users and perspectives. The initial approach was to utilize the existing methods getValue and getMetaData without altering the original API and data structure. Here's how it works:

To retrieve sensor, actuator, etc., values under a VSS branch level using one asterisk.
```
Client -> getValue Vehicle.ADAS.*
Response -> 
[
    {
        "path": "Vehicle.ADAS.ActiveAutonomyLevel"
    },
    {
        "path": "Vehicle.ADAS.SupportedAutonomyLevel"
    },
    {
        "path": "Vehicle.ADAS.PowerOptimizeLevel"
    }
]
```

To obtain possible branches under a branch level using two asterisks
```
Client -> getValue Vehicle.ADAS.**
Response -> 
[
    {
        "path": "Vehicle.ADAS.EBA"
    },
    {
        "path": "Vehicle.ADAS.LaneDepartureDetection"
    },
    {
        "path": "Vehicle.ADAS.ABS"
    },
    {
        "path": "Vehicle.ADAS.ObstacleDetection"
    },
    {
        "path": "Vehicle.ADAS.EBD"
    },
    {
        "path": "Vehicle.ADAS.ESC"
    },
    {
        "path": "Vehicle.ADAS.CruiseControl"
    },
    {
        "path": "Vehicle.ADAS.TCS"
    }
]
```
The getMetaData method will return the same response but with additional details for sensors, actuators, etc.

This pull request is a starting point to collectively refine and finalize the autocompletion feature, feedback is appreciated


erikbosch:
Seems the new pre-commit check complains - there seems to be a trailing blank in on of the files you have modified (https://github.com/eclipse/kuksa.val/actions/runs/5608284568/jobs/10260450011?pr=605)

rafaeling:
Thanks @erikbosch, forgot to apply it before creating PR.

lukasmittag:
I find the use of getValue a bit missleading. But if it works with getMetaData as well i think its okay. But in the documentation we should definitely use getMetaData getValue is a false command.

lukasmittag:
Another comment. I think maybe of the structure it is not possible but in databroker-cli if I do a ``` metadata Vehicle.* ``` I get every possible sensor/actuator/attribute not only the ones directly under Vehicle. At least we should have the same behaviour in both components in my pov

lukasmittag:
```
metadata Vehicle.OBD.*
[metadata]  OK  
Path                                      Entry type Data type
Vehicle.OBD.Catalyst.Bank1.Temperature1   Sensor     Float    
Vehicle.OBD.Catalyst.Bank1.Temperature2   Sensor     Float    
Vehicle.OBD.Catalyst.Bank2.Temperature1   Sensor     Float    
Vehicle.OBD.Catalyst.Bank2.Temperature2   Sensor     Float    
Vehicle.OBD.DriveCycleStatus.DTCCount     Sensor     Uint8    
Vehicle.OBD.DriveCycleStatus.IgnitionType Sensor     String   
Vehicle.OBD.DriveCycleStatus.IsMILOn      Sensor     Bool     
Vehicle.OBD.O2.Sensor1.ShortTermFuelTrim  Sensor     Float    
Vehicle.OBD.O2.Sensor1.Voltage            Sensor     Float    
Vehicle.OBD.O2.Sensor2.ShortTermFuelTrim  Sensor     Float    
Vehicle.OBD.O2.Sensor2.Voltage            Sensor     Float    
Vehicle.OBD.O2.Sensor3.ShortTermFuelTrim  Sensor     Float    
Vehicle.OBD.O2.Sensor3.Voltage            Sensor     Float    
Vehicle.OBD.O2.Sensor4.ShortTermFuelTrim  Sensor     Float    
Vehicle.OBD.O2.Sensor4.Voltage            Sensor     Float    
Vehicle.OBD.O2.Sensor5.ShortTermFuelTrim  Sensor     Float    
Vehicle.OBD.O2.Sensor5.Voltage            Sensor     Float    
Vehicle.OBD.O2.Sensor6.ShortTermFuelTrim  Sensor     Float    
Vehicle.OBD.O2.Sensor6.Voltage            Sensor     Float    
Vehicle.OBD.O2.Sensor7.ShortTermFuelTrim  Sensor     Float    
Vehicle.OBD.O2.Sensor7.Voltage            Sensor     Float    
Vehicle.OBD.O2.Sensor8.ShortTermFuelTrim  Sensor     Float    
Vehicle.OBD.O2.Sensor8.Voltage            Sensor     Float    
Vehicle.OBD.O2WR.Sensor1.Current          Sensor     Float    
Vehicle.OBD.O2WR.Sensor1.Lambda           Sensor     Float    
Vehicle.OBD.O2WR.Sensor1.Voltage          Sensor     Float    
Vehicle.OBD.O2WR.Sensor2.Current          Sensor     Float    
Vehicle.OBD.O2WR.Sensor2.Lambda           Sensor     Float    
Vehicle.OBD.O2WR.Sensor2.Voltage          Sensor     Float    
Vehicle.OBD.O2WR.Sensor3.Current          Sensor     Float    
Vehicle.OBD.O2WR.Sensor3.Lambda           Sensor     Float    
Vehicle.OBD.O2WR.Sensor3.Voltage          Sensor     Float    
Vehicle.OBD.O2WR.Sensor4.Current          Sensor     Float    
Vehicle.OBD.O2WR.Sensor4.Lambda           Sensor     Float    
Vehicle.OBD.O2WR.Sensor4.Voltage          Sensor     Float    
Vehicle.OBD.O2WR.Sensor5.Current          Sensor     Float    
Vehicle.OBD.O2WR.Sensor5.Lambda           Sensor     Float    
Vehicle.OBD.O2WR.Sensor5.Voltage          Sensor     Float    
Vehicle.OBD.O2WR.Sensor6.Current          Sensor     Float    
Vehicle.OBD.O2WR.Sensor6.Lambda           Sensor     Float    
Vehicle.OBD.O2WR.Sensor6.Voltage          Sensor     Float    
Vehicle.OBD.O2WR.Sensor7.Current          Sensor     Float    
Vehicle.OBD.O2WR.Sensor7.Lambda           Sensor     Float    
Vehicle.OBD.O2WR.Sensor7.Voltage          Sensor     Float    
Vehicle.OBD.O2WR.Sensor8.Current          Sensor     Float    
Vehicle.OBD.O2WR.Sensor8.Lambda           Sensor     Float    
Vehicle.OBD.O2WR.Sensor8.Voltage          Sensor     Float    
Vehicle.OBD.Status.DTCCount               Sensor     Uint8    
Vehicle.OBD.Status.IgnitionType           Attribute  String   
Vehicle.OBD.Status.IsMILOn                Sensor     Bool 
```
```
sdv.databroker.v1 > metadata Vehicle.OBD.*.ShortTermFuelTrim
[metadata]  OK  
Path                                     Entry type Data type
Vehicle.OBD.O2.Sensor1.ShortTermFuelTrim Sensor     Float    
Vehicle.OBD.O2.Sensor2.ShortTermFuelTrim Sensor     Float    
Vehicle.OBD.O2.Sensor3.ShortTermFuelTrim Sensor     Float    
Vehicle.OBD.O2.Sensor4.ShortTermFuelTrim Sensor     Float    
Vehicle.OBD.O2.Sensor5.ShortTermFuelTrim Sensor     Float    
Vehicle.OBD.O2.Sensor6.ShortTermFuelTrim Sensor     Float    
Vehicle.OBD.O2.Sensor7.ShortTermFuelTrim Sensor     Float    
Vehicle.OBD.O2.Sensor8.ShortTermFuelTrim Sensor     Float    
```
```
sdv.databroker.v1 > metadata Vehicle.*.ShortTermFuelTrim
[metadata]  OK  
Path                                     Entry type Data type
Vehicle.OBD.O2.Sensor1.ShortTermFuelTrim Sensor     Float    
Vehicle.OBD.O2.Sensor2.ShortTermFuelTrim Sensor     Float    
Vehicle.OBD.O2.Sensor3.ShortTermFuelTrim Sensor     Float    
Vehicle.OBD.O2.Sensor4.ShortTermFuelTrim Sensor     Float    
Vehicle.OBD.O2.Sensor5.ShortTermFuelTrim Sensor     Float    
Vehicle.OBD.O2.Sensor6.ShortTermFuelTrim Sensor     Float    
Vehicle.OBD.O2.Sensor7.ShortTermFuelTrim Sensor     Float    
Vehicle.OBD.O2.Sensor8.ShortTermFuelTrim Sensor     Float
```
also this is possible in databroker-cli/sdv.databroker.v1 but not now. But yeah maybe too because of the structure. Just dropping some things that I find nice in the databroker-cli/sdv.databroker.v1. Not necessary more a notice what I am missing. Do not know if its possible without changes to the API.

erikbosch:
Without considering limitations in the current *.proto files - would it not be interesting to have a syntax where `*` indicates just next level while `**` indicates all levels or arbitrary number of levels.

So that if we have a tree like:

```
A # Branch
A.B # Signal
A.C # Branch
A.C.D # Signal
A.C.E # Signal
A.F # Signal
```

and do `getValue A.*` then we would get `A.B` and `A.F` only, but if we do  `getValue A.**` we would get  `A.B`,`A.F`, `A.C.D` and `A.C.E`. 

To get a branch without showing signals within looks a bit odd for me if we do a `getValue` without extra parameters.

SebastianSchildt:
I am not super convinced about the "special meaning" of `*`and `**`. I think it would be better databroker/the API "understood" the concpet of branches (as they are a core element of VSS), and I think it is minimally-invasive

Let me try to look through this from Client/API and Databroker perspective


So from a client point of view when i the example you do

```
Client -> getValue Vehicle.Speed
```

This will create a getRequest with an EntryRequests that actually sets the Fields/Views in a way that you are interested in current value.

Now in the example

```
Client -> getValue Vehicle.ADAS.* 
Response -> 
[
    {
        "path": "Vehicle.ADAS.ActiveAutonomyLevel"
    },
    {
        "path": "Vehicle.ADAS.SupportedAutonomyLevel"
    },
    {
        "path": "Vehicle.ADAS.PowerOptimizeLevel"
    }
]
```

databroker does the expansion and basically gives you the response, you would get if you created call yourself, listing EntryRequests for all paths. We don't see values here, as they have not been set (i.e. Datapoint value in DataEntry strcuture in proto not set).

I would _expect_ that the expansion (sor for string poath in Entry Request) would also expand branches, so the expansion of 
`Vehicle.ADAS.*` would yield

```
Vehicle.ADAS.ActiveAutonomyLevel
Vehicle.ADAS.SupportedAutonomyLevel
Vehicle.ADAS.PowerOptimizeLevel
Vehicle.ADAS.EBA
Vehicle.ADAS.LaneDepartureDetection
Vehicle.ADAS.ABS
Vehicle.ADAS.ObstacleDetection
"Vehicle.ADAS.EBD"
"Vehicle.ADAS.ESC"
"Vehicle.ADAS.CruiseControl"
"Vehicle.ADAS.TCS"
```

now I do write "expansion" because for the returns on getValue I could all coming back with empty value (as branches have none), but that is not quite consistent with waht I would expect, becasue when doing

```
GetValue Vehicle.ADAS.CruiseControl
```
 I would expect an answer like "Can't read branches", what I am saying is, the functions might try to be clever how to expand, especially when using VIEWs.

On the other hand when doing a potential

```
Client -> getMetadata Vehicle.ADAS.* 
```

Whoch would generate a GetRequest->EntryRequest with METADATA_FIELDS or VIEW_METADATA
I want the whole expansion taken into account. Of the response should correctly identify stuff like `Vehicle.ADAS.TCS`as branch.

THis only needs `*`, if we want we can use `**`to mean recursively getting all paths.


-----------

Nor for the changes in .proto and maybe databroker. I think they can be pretty *minimal*.

## proto
In the proto `enum EntryType` needs to be extended with `BRANCH`. Done!

## databroker
Databroker has a Hashmap of VSSPaths (only sensor/actuators/attributes). We can _keep_ it just also store Paths that aare Branches in the same Hashmap. Similar to the proto, we need to make sure databroker understands this. I am not that deep in the RUST code but I think extending the EntryType struct https://github.com/eclipse/kuksa.val/blob/2b2a9082ca814f00b91b35ac01c36d885a006876/kuksa_databroker/databroker/src/types.rs#L47 with Branch might do it as well.
There is no need for making any complex tree structure just now. Of course this implies some conditionals in varous places in the code to check "is what I have here really  sensor/not a branch", but we have this logic somewhere anyway, e.g. when trying to set a target value for sensor or attribute, which is not allowed.
Of ocurse the JSON parsing needs to also add branches, but that seems very doable

Also, -but I admit this is conjecture/speculation at this point - you already get the VSS defined branch meta "for free", and should somebody decide to annotate more useful stuff on branch level via cusrtom VSS attributes, it would be much easier to support that.

My main argument however is, I feel doing it as lined out above feels much cleaner/consistent with how the API and datastructure already is, isntead of a custom hack for one use case only

(_if_ going for a custom hack, and I am not saying. we should - just being "Honest" about it, and adding a fucntion "get Children(Path) returning an array of strings might be a cleaner way to go)

rafaeling:
Here are the new outputs after some changes on API level, so far only ```*``` case has been fixed, ```**``` recursive case will be fixed later in case we want to continue with this approach.

```console
getMetaData Vehicle.ADAS.*
[
    {
        "path": "Vehicle.ADAS.ABS",
        "metadata": {
            "data_type": "UNSPECIFIED",
            "entry_type": "BRANCH"
        }
    },
    {
        "path": "Vehicle.ADAS.EBD",
        "metadata": {
            "data_type": "UNSPECIFIED",
            "entry_type": "BRANCH"
        }
    },
    {
        "path": "Vehicle.ADAS.EBA",
        "metadata": {
            "data_type": "UNSPECIFIED",
            "entry_type": "BRANCH"
        }
    },
    {
        "path": "Vehicle.ADAS.LaneDepartureDetection",
        "metadata": {
            "data_type": "UNSPECIFIED",
            "entry_type": "BRANCH"
        }
    },
    {
        "path": "Vehicle.ADAS.ObstacleDetection",
        "metadata": {
            "data_type": "UNSPECIFIED",
            "entry_type": "BRANCH"
        }
    },
    {
        "path": "Vehicle.ADAS.CruiseControl",
        "metadata": {
            "data_type": "UNSPECIFIED",
            "entry_type": "BRANCH"
        }
    },
    {
        "path": "Vehicle.ADAS.DMS",
        "metadata": {
            "data_type": "UNSPECIFIED",
            "entry_type": "BRANCH"
        }
    },
    {
        "path": "Vehicle.ADAS.SupportedAutonomyLevel",
        "metadata": {
            "data_type": "STRING",
            "entry_type": "ATTRIBUTE",
            "description": "Indicates the highest level of autonomy according to SAE J3016 taxonomy the vehicle is capable of."
        }
    },
    {
        "path": "Vehicle.ADAS.TCS",
        "metadata": {
            "data_type": "UNSPECIFIED",
            "entry_type": "BRANCH"
        }
    },
    {
        "path": "Vehicle.ADAS.ESC",
        "metadata": {
            "data_type": "UNSPECIFIED",
            "entry_type": "BRANCH"
        }
    },
    {
        "path": "Vehicle.ADAS.ActiveAutonomyLevel",
        "metadata": {
            "data_type": "STRING",
            "entry_type": "SENSOR",
            "description": "Indicates the currently active level of autonomy according to SAE J3016 taxonomy."
        }
    },
    {
        "path": "Vehicle.ADAS.PowerOptimizeLevel",
        "metadata": {
            "data_type": "UINT8",
            "entry_type": "ACTUATOR",
            "description": "Power optimization level for this branch/subsystem. A higher number indicates more aggressive power optimization. Level 0 indicates that all functionality is enabled, no power optimization enabled. Level 10 indicates most aggressive power optimization mode, only essential functionality enabled."
        }
    }
]
```

```console
getValue Vehicle.ADAS.*
[
    {
        "path": "Vehicle.ADAS.ABS"
    },
    {
        "path": "Vehicle.ADAS.EBD"
    },
    {
        "path": "Vehicle.ADAS.EBA"
    },
    {
        "path": "Vehicle.ADAS.LaneDepartureDetection"
    },
    {
        "path": "Vehicle.ADAS.ObstacleDetection"
    },
    {
        "path": "Vehicle.ADAS.CruiseControl"
    },
    {
        "path": "Vehicle.ADAS.DMS"
    },
    {
        "path": "Vehicle.ADAS.SupportedAutonomyLevel"
    },
    {
        "path": "Vehicle.ADAS.TCS"
    },
    {
        "path": "Vehicle.ADAS.ESC"
    },
    {
        "path": "Vehicle.ADAS.ActiveAutonomyLevel"
    },
    {
        "path": "Vehicle.ADAS.PowerOptimizeLevel",
        "value": {
            "value": 15,
            "timestamp": "2023-07-26T07:43:35.406717+00:00"
        }
    }
```
```console
setValue Vehicle.ADAS.LaneDepartureDetection 40
{
    "error": {
        "code": 404,
        "reason": "not_found",
        "message": "Can't read branches"
    },
    "errors": [
        {
            "path": "Vehicle.ADAS.LaneDepartureDetection",
            "error": {
                "code": 404,
                "reason": "not_found",
                "message": "Can't read branches"
            }
        }
    ]
}
```

rafaeling:
Here are the new outputs after some changes on API level, so far only ```*``` case has been fixed, ```**``` recursive case will be fixed later in case we want to continue with this approach.

```console
getMetaData Vehicle.ADAS.*
[
    {
        "path": "Vehicle.ADAS.ABS",
        "metadata": {
            "data_type": "UNSPECIFIED",
            "entry_type": "BRANCH"
        }
    },
    {
        "path": "Vehicle.ADAS.EBD",
        "metadata": {
            "data_type": "UNSPECIFIED",
            "entry_type": "BRANCH"
        }
    },
    {
        "path": "Vehicle.ADAS.EBA",
        "metadata": {
            "data_type": "UNSPECIFIED",
            "entry_type": "BRANCH"
        }
    },
    {
        "path": "Vehicle.ADAS.LaneDepartureDetection",
        "metadata": {
            "data_type": "UNSPECIFIED",
            "entry_type": "BRANCH"
        }
    },
    {
        "path": "Vehicle.ADAS.ObstacleDetection",
        "metadata": {
            "data_type": "UNSPECIFIED",
            "entry_type": "BRANCH"
        }
    },
    {
        "path": "Vehicle.ADAS.CruiseControl",
        "metadata": {
            "data_type": "UNSPECIFIED",
            "entry_type": "BRANCH"
        }
    },
    {
        "path": "Vehicle.ADAS.DMS",
        "metadata": {
            "data_type": "UNSPECIFIED",
            "entry_type": "BRANCH"
        }
    },
    {
        "path": "Vehicle.ADAS.SupportedAutonomyLevel",
        "metadata": {
            "data_type": "STRING",
            "entry_type": "ATTRIBUTE",
            "description": "Indicates the highest level of autonomy according to SAE J3016 taxonomy the vehicle is capable of."
        }
    },
    {
        "path": "Vehicle.ADAS.TCS",
        "metadata": {
            "data_type": "UNSPECIFIED",
            "entry_type": "BRANCH"
        }
    },
    {
        "path": "Vehicle.ADAS.ESC",
        "metadata": {
            "data_type": "UNSPECIFIED",
            "entry_type": "BRANCH"
        }
    },
    {
        "path": "Vehicle.ADAS.ActiveAutonomyLevel",
        "metadata": {
            "data_type": "STRING",
            "entry_type": "SENSOR",
            "description": "Indicates the currently active level of autonomy according to SAE J3016 taxonomy."
        }
    },
    {
        "path": "Vehicle.ADAS.PowerOptimizeLevel",
        "metadata": {
            "data_type": "UINT8",
            "entry_type": "ACTUATOR",
            "description": "Power optimization level for this branch/subsystem. A higher number indicates more aggressive power optimization. Level 0 indicates that all functionality is enabled, no power optimization enabled. Level 10 indicates most aggressive power optimization mode, only essential functionality enabled."
        }
    }
]
```

```console
getValue Vehicle.ADAS.*
[
    {
        "path": "Vehicle.ADAS.ABS"
    },
    {
        "path": "Vehicle.ADAS.EBD"
    },
    {
        "path": "Vehicle.ADAS.EBA"
    },
    {
        "path": "Vehicle.ADAS.LaneDepartureDetection"
    },
    {
        "path": "Vehicle.ADAS.ObstacleDetection"
    },
    {
        "path": "Vehicle.ADAS.CruiseControl"
    },
    {
        "path": "Vehicle.ADAS.DMS"
    },
    {
        "path": "Vehicle.ADAS.SupportedAutonomyLevel"
    },
    {
        "path": "Vehicle.ADAS.TCS"
    },
    {
        "path": "Vehicle.ADAS.ESC"
    },
    {
        "path": "Vehicle.ADAS.ActiveAutonomyLevel"
    },
    {
        "path": "Vehicle.ADAS.PowerOptimizeLevel",
        "value": {
            "value": 15,
            "timestamp": "2023-07-26T07:43:35.406717+00:00"
        }
    }
```
```console
getValue Vehicle.ADAS.LaneDepartureDetection
{
    "error": {
        "code": 404,
        "reason": "not_found",
        "message": "Can't read branches"
    },
    "errors": [
        {
            "path": "Vehicle.ADAS.LaneDepartureDetection",
            "error": {
                "code": 404,
                "reason": "not_found",
                "message": "Can't read branches"
            }
        }
    ]
}
```

proto/kuksa/val/v1/types.proto
```diff
@@ -11,278 +11,288 @@
...
-}
-
-message FloatArray {
-  repeated float values = 1;
-}
-
-message DoubleArray {
-  repeated double values = 1;
-}
+ syntax = "proto3";
```
erikbosch, line 14: Changing indentation to 1 for everything? Any special reason?

proto/kuksa/val/v1/types.proto
```diff
@@ -11,278 +11,288 @@
...
+ }
+ 
+ message FloatArray {
+   repeated float values = 1;
+ }
+ 
+ message DoubleArray {
+   repeated double values = 1;
+ }
+
```
erikbosch, line 298: Lacking return at end of file?

kuksa-client/kuksa_client/grpc/__init__.py
```diff
@@ -1,992 +0,0 @@
-########################################################################
```
erikbosch, line 1: A mistake that this file is deleted?

proto/kuksa/val/v1/types.proto
rafaeling: I have pushed a cleaner commit, it seems my proto editor does not work

kuksa-client/kuksa_client/grpc/__init__.py
rafaeling: Yes, check new commits

lukasmittag (changes requested):
Had no look at the code yet.
What I've encounterd is:

```
Test Client> getMetaData Vehicle.ADAS.*
{
    "error": "ValueError in casting the value."
}
```
it seems not to work like you specified. But what's possible is:
```
Test Client> getMetaData Vehicle.ADAS.ABS.*
[
    {
        "path": "Vehicle.ADAS.ABS.IsEngaged",
        "metadata": {
            "data_type": "BOOLEAN",
            "entry_type": "SENSOR",
            "description": "Indicates if ABS is currently regulating brake pressure. True = Engaged. False = Not Engaged."
        }
    },
    {
        "path": "Vehicle.ADAS.ABS.IsEnabled",
        "metadata": {
            "data_type": "BOOLEAN",
            "entry_type": "ACTUATOR",
            "description": "Indicates if ABS is enabled. True = Enabled. False = Disabled."
        }
    },
    {
        "path": "Vehicle.ADAS.ABS.IsError",
        "metadata": {
            "data_type": "BOOLEAN",
            "entry_type": "SENSOR",
            "description": "Indicates if ABS incurred an error condition. True = Error. False = No Error."
        }
    }
]
```
Also getValue seems to work 
```
Test Client> getValue Vehicle.ADAS.*
[
    {
        "path": "Vehicle.ADAS.CruiseControl"
    },
    {
        "path": "Vehicle.ADAS.EBA"
    },
    {
        "path": "Vehicle.ADAS.TCS"
    },
    {
        "path": "Vehicle.ADAS.ObstacleDetection"
    },
    {
        "path": "Vehicle.ADAS.EBD"
    },
    {
        "path": "Vehicle.ADAS.ActiveAutonomyLevel"
    },
    {
        "path": "Vehicle.ADAS.LaneDepartureDetection"
    },
    {
        "path": "Vehicle.ADAS.SupportedAutonomyLevel"
    },
    {
        "path": "Vehicle.ADAS.ABS"
    },
    {
        "path": "Vehicle.ADAS.ESC"
    },
    {
        "path": "Vehicle.ADAS.DMS"
    },
    {
        "path": "Vehicle.ADAS.PowerOptimizeLevel"
    }
]
```
Futhermore I think the expected behavior you specified/said how it would work is okay. For getting all the metadata one could work through all the branches and get all sensors until no branch exists anymore. So I think it is a good starting point and what we need. Maybe some error got sneaked in in your fixing commits.

eriksven:
Sorry for being a bit late to the party. 

For a missing value in the result of `getValue` I find it quite confusing to differentiate whether this path was a branch or an unset value. For that reason, I would not return branches in `getValue`. 

Why are we so interested in branches here anyway? Even for `getMetadata` I would be more interested in all child signals of that branch element instead of the pure existence of that branch. By giving the full signal path we implicitly also know about the existence of the branches leading to those signals.

Not storing the branches as individual elements might make it even easier since we do not have to manage the branches in the same hashmap. I like the opportunity to add further information about a VSS branch, but imho we better wait until there is actually more information for a branch in a later VSS version.

What is the use case and reasoning for only returning one level in the tree? Is this mainly to save computation?
My current use case for the *-feature would be to get an overview of the existence of all signals in a databroker or a given subtree to do things like "Please record all available signals". Therefore, I see more value when `getMetadata` already traverses the tree for me, and I would prioritize this version.

In addition, my general expectation of using a * Operator at the end of the path would have been to get all elements (signals) matching the prefix. Considering that the implementation uses a HashMap and not a tree structure, I would even consider going so far as to allow stuff like `Vehicle.*.IsOpen` to get all signals whose path starts with `Vehicle` and ends with `IsOpen`. But, at the moment, I would not support such in-between *-operators to theoretically allow for changing the internal data model to a tree.

argerus:
Hmm, this PR which was meant to introduce discoverability to the `kuksa.val.v1` in order to support autocompletion in kuksa-client. It seems to have morphed into something else since I last had a look.

That can happen, but I fail to see what is gained by adding branches as an entry type. I do see plenty of reasons _not_ to do this though (which is why it wasn't added in the first place).

Basically, I'm wondering the same thing as Sven (above):

> Why are we so interested in branches here anyway? Even for getMetadata I would be more interested in all child signals of that branch element instead of the pure existence of that branch. By giving the full signal path we implicitly also know about the existence of the branches leading to those signals.

> Not storing the branches as individual elements might make it even easier since we do not have to manage the branches in the same hashmap. I like the opportunity to add further information about a VSS branch, but imho we better wait until there is actually more information for a branch in a later VSS version.

* First of all, branches are not needed in order to implement discoverability as evident by the fact that this feature already exist in `databroker-cli` (which uses the `sdv.databroker.v1` API that doesn't have branches either).
* Second of all, branches where _intentionally_ not included as an "entry type" because databroker is fundamentally used for setting / getting and subscribing to **signals**. It's argued that branches should be added as they are part of the VSS standard. Yes, but they are not signals, and I would argue that the information encoded by branches in VSS are _only_ used to generate the resulting list of signals in the model. Everything encoded by the branches is (or can/should be) encoded in these resulting signals.

Another thing that becomes apparent when reading through the comments here is that the expected behaviour are described in terms of:

* `getValue`
* `getMetadata`

I think it was a mistake in `kuksa.val.v1` to try to put all these different concepts into a single `get` function. I think the API would become a lot more intuitive if it was split into separate functions at the GRPC level (which would make it easier to implement in other languages as well). The same goes for all the other functions that have clearly different expected behaviours associated with them (e.g. setting actuators or providing values).

SebastianSchildt:
Independent of implementation details, there are **two things** I want to achieve here):

# Capability 1
Being able to subscribe to a “whole branch”, “all signals” in the tree/under a branch. This is a request I hear very often, which  - from a clients perspective - is  the support of wildcards in  “paths’s” when requesting.
I will summarize the result from  last f2f discussion here, as I think that captures that aspect from  a users perspective quite well

### For getValue:
* `getValue path.*` should return one level of child signals (excluding branches) that CAN contain values.
*` getValue path.**` should recursively return all child signals (excluding branches) that CAN contain values.

(and similarly when targetValue or metadata is requested)

This is something this PR can do already. And this would be also be *independent* from whether we “understand” branches or not (without branches) we would just not get branch metadata.

# Capability 2
Being able to “discover” children. To learn what signals are available  under a branch. The solution here does it by having branches included in metadata, so that feels natural (to me), having it returned with the same formats and mechanisms you deal with, when handling leaves. Of course, an alternative way could be  having something like. List(Childs) Call as was sketched mentioned in #411



### Arguments...
... for doing it the way it is proposed here: Some of this might come down to a matter of taste, arguing what is “easier to use”.  Consider that argument made. 

The approach here will also help us getting basic VISS support for metadata back. In VISS Read  https://www.w3.org/TR/viss2-core/#read there is the “static-metadata read” filter which is basically returning what we call metdata in databroker, see https://www.w3.org/TR/viss2-core/#static-metadata-request
The VISS spec itself uses this capability to query branches, see “Signal Discovery Read” Example: https://www.w3.org/TR/viss2-transport/#https-service-discovery-read , so they basically achieve discoverability by returning branch metadata. To me the VISS way mades sense here, and while the change in this PR might not be the only way to be able to recreate it, it makes it quite easier.

A second - admittedly  somehwat weaker - argument is, that once we support structs as defined in VSS we might  need t deal with more than a list of signals anyway, as structs are basically over-glorified branches, at least on the model side: https://covesa.github.io/vehicle_signal_specification/rule_set/data_entry/data_types_struct/. That is more of a hunch though, as I have no idea yet how (and when) we would want to support those. They are in the model though, so once the standard catalogue begins using them, that might press the issue...

SebastianSchildt:
> ### What we want to achieve

> > For getValue:
> > 
> > * `getValue path.*` should return one level of child signals (excluding branches) that CAN contain values.
> > * `getValue path.**` should recursively return all child signals (excluding branches) that CAN contain values.
> > 
> > (and similarly when targetValue or metadata is requested)
> 
> Yes, this also seems reasonable.
> 
> _Nitpick: ALL signals can contain values. That's the definition of signals in VSS, right?_

True, in terms of the current implementation this just means NOT returning branches, and in case target values were requested only returning actuators. 

 
> Addition: If we want to match the permission syntax, e.g. `read:Vehicle`, we should also support:
> 
> * `getValue path` (where `path` is a branch)
> 
> that would behave exactly the same as `getValue path.**` above, i.e. "should recursively return all child signals [...]".

Agreed.


> ### Changing kuksa.val.v1
> In order to support these capabilities in `kuksa.val.v1`, it would need to be updated to support paths as specified above in the `Get()` and `Subscribe()` functions:
> 
> ```proto
> message EntryRequest {
>   string path           = 1; // <--- Add support for paths as specified above
>   View view             = 2;
>   repeated Field fields = 3;
> }
> ```
> 
>  ```proto
> message SubscribeEntry {
>   string path           = 1; // <---  Add support for paths as specified above
>   View view             = 2;
>   repeated Field fields = 3;
> }
> ```
> 
>  * Getting `path = "Vehicle.Cabin.Seat.Row2.DriverSide.*"` would return all signals with a path matching that.
> * Subscribing to `path = "Vehicle.Cabin.Seat.Row2.DriverSide.*"` would notify of changes for all signals with a path matching that.

Agreed, although I think the current code does not consider Subscribe yet. But yeah, it should act and feel the same.
> 
> ### Making it complex
> But that is not what is currently being proposed.
> 

I do think, that is exactly the core of what is proposed here, i.e. allowing paths with the wildcards as above

> Instead, the proposal seems to suggest that what is returned should also be based on the contents of `view` and `fields`. Is this correct?

Yes, that happens in _addition_,  I do however think there is no easy way around this. For example targetValues do not make sense for sensors and actuators. (btw. I do think that is still a bug in current master, that you can query them, at least you can not set them anymore without getting an error), so on "wildcard paths", I assume if a branch contains a few actuators and some sensors, you would not want it to error out in case there is target value requested. On the other hand (consider get), I am not sure it is very intuitive if you get the path to a sensor back (which you actually do not know is a sensor), which just "happens" to have no target Value (because actually it can never have one). (Of course same applies if there are branches in the mix, just want to point out just with "signals" alone it is an issue)



> > Of course, an alternative way could be having something like. List(Childs) Call as was sketched mentioned in #411
> 
> This would be an alternative to adding support for wildcards in `Get`, with the logic:
> 
> * `Get([entry1, entry2, entry3])` would _always_ return exactly 3 entries/signals (or an error).
> * `List("wildcard", fields...)` would return an unknown number of entries/signals.
> 
> > The solution here does it by having branches included in metadata, so that feels natural (to me), having it returned with the same formats and mechanisms you deal with, when handling leaves.
> 
> List would also have the same format of the entries/signals. The only difference would be supporting entries / signals not known ahead of time and putting this in a separate function (since the expectation of what is returned is different). With that said, I have nothing against supporting this functionality by adding support for wildcards in the get function itself.

> 
> Point being, I don't see how adding branches really helps in anything listed in **Capability 1** or **Capability 2**.

I am not sure I get it fully, but this feels 90% aligned let me try

  - We allow wildcards in get function as described above (*, **), (for elegance we also define x.y.branch being equivalent to x.y.branch.**)we do NOT need to return Branches (as type), but are still able to get "Not explicitly known/listed signals). So that would give us capability 1
  - Question: I still feel we need to be a bit clever filter output based on views/fields a bit
  - For Capability 2 "discovery", your suggestion implies it is is no use having "branches" as first class members (or querying a "list of children), but just using the wildcard paths as defined above (querying values, or metadata), so you get to know all child or successor signals? Which would also be usable for "autocomplete" (and if you really want, you know about branches as elements in the paths implicitly anyway)?

That at least sounds like a solution I can live with.

Confirm/deny/comment pls

SebastianSchildt:
Note: In the process we should also document how the expansion works exactly, i.e. the semantics of `*`and `**`, i.e. are they allowed to be used in the middle of a  path?

argerus:
`*`  Matches everything until the next `.`

`**` Matches everything

Examples:

| Path                | Matches                              |
|---------------------|--------------------------------------|
| `""`                | Everything                           |
| `"Vehicle"`         | Everything starting with `Vehicle`   |
| `"Vehicle.Cabin.Sunroof"` | `Vehicle.Cabin.Sunroof.Position`<br>`Vehicle.Cabin.Sunroof.Shade.Position`<br>`Vehicle.Cabin.Sunroof.Shade.Switch`<br>`Vehicle.Cabin.Sunroof.Switch` |
| `"Vehicle.Cabin.Sunroof.**"` | `Vehicle.Cabin.Sunroof.Position`<br>`Vehicle.Cabin.Sunroof.Shade.Position`<br>`Vehicle.Cabin.Sunroof.Shade.Switch`<br>`Vehicle.Cabin.Sunroof.Switch`  |
| `"Vehicle.Cabin.Sunroof.*"` | `Vehicle.Cabin.Sunroof.Position`<br>`Vehicle.Cabin.Sunroof.Switch`  |
| `"Vehicle.Cabin.Sunroof.*.Position"` | `Vehicle.Cabin.Sunroof.Position` |
| `"Vehicle.Cabin.Sunroof.**.Position"` | `Vehicle.Cabin.Sunroof.Position`<br>`Vehicle.Cabin.Sunroof.Shade.Position` |
| `"**.Sunroof"`      | `Vehicle.Cabin.Sunroof.Position`<br>`Vehicle.Cabin.Sunroof.Shade.Position`<br>`Vehicle.Cabin.Sunroof.Shade.Switch`<br>`Vehicle.Cabin.Sunroof.Switch` |
| `"*.Sunroof"`       | Nothing|
| `"Sunroof"`         | Nothing|

I created PR #639, adding this as a doc in the repo as well. Discussion / additions around this can be had there perhaps?

kuksa_databroker/databroker/src/glob.rs
```diff
@@ -51,3 +52,64 @@ pub fn to_regex(glob: &str) -> Result<Regex, Error> {
...
+
+        regex.is_match(input) || regex2.is_match(input)
+    } else {
+        false
+    }
+}
+
+#[tokio::test]
+async fn test_valid_request_path() {
+    let inputs = vec![
```
argerus, line 72: Since the input is split from the assertions below, it would be easier to see which paths are supposed to match by including the "expectation" in the same place.

I.e. something like:
```rust
let inputs = &[
    ("String.*", true),
    ...
];
```

argerus (changes requested):

kuksa_databroker/databroker/src/glob.rs
```diff
@@ -51,3 +52,70 @@ pub fn to_regex(glob: &str) -> Result<Regex, Error> {
...
+}
+
+pub fn matches_path_pattern(input: &str) -> bool {
+    let pattern = r"^(?:\w+(?:\.\w+)*)(?:\.\*+)?$";
+    let regex = Regex::new(pattern).unwrap();
+
+    let pattern2 = r"^(\*{2}|\*\.)?[^.]+(\.[^.]+)*\.(\*{2}|\*|[A-Za-z]+)(\.[^.]+)*\.[^.]+$";
+    let regex2 = Regex::new(pattern2).unwrap();
+
+    regex.is_match(input) || regex2.is_match(input)
```
argerus, line 93: These regex:es do not seem to correctly identify valid patterns.

You can have a look at the regex in `src/jwt/scope.rs`, as that covers the same functionality (i.e. the `path` group matches valid patterns).

kuksa_databroker/databroker/src/glob.rs
```diff
@@ -51,3 +52,70 @@ pub fn to_regex(glob: &str) -> Result<Regex, Error> {
...
+
+    re
+}
+
+pub fn to_regex_new(glob: &str) -> Result<Regex, Error> {
+    let re = to_regex_string_new(glob);
+    Regex::new(&re).map_err(|_err| Error::RegexError)
+}
+
+pub fn matches_path_pattern(input: &str) -> bool {
```
argerus, line 86:
```suggestion
pub fn is_valid_pattern(input: &str) -> bool {
```

argerus (changes requested):

rafaeling:
For the use case "**.Sunroof" which returns:

Vehicle.Cabin.Sunroof.Position
Vehicle.Cabin.Sunroof.Shade.Position
Vehicle.Cabin.Sunroof.Shade.Switch
Vehicle.Cabin.Sunroof.Switch

I would use instead the syntax "\*\*.Sunroof.\*\*" since it fits better the regex pattern.

argerus:
> For the use case "**.Sunroof" which returns:
> 
> Vehicle.Cabin.Sunroof.Position Vehicle.Cabin.Sunroof.Shade.Position Vehicle.Cabin.Sunroof.Shade.Switch Vehicle.Cabin.Sunroof.Switch
> 
> I would use instead the syntax "**.Sunroof.**" since it fits better the regex pattern.

Yes, that example has already been changed in #639 (to return nothing, as no signal matches that pattern).

kuksa_databroker/databroker/src/glob.rs
```diff
@@ -51,3 +52,113 @@ pub fn to_regex(glob: &str) -> Result<Regex, Error> {
...
+    false
+}
+
+pub fn to_regex_string_new(glob: &str) -> String {
+    let mut input = String::from(glob);
+
+    //check if asterisk in middle
+    if has_asterisk_in_middle(glob) {
+        input = glob.replace(".*.", ".");
+    }
```
argerus, line 78: Why are we checking if there is an asterisk in the middle?
Why are we replacing `.*.` with `.`. Isn't that just removing the wildcard completely?

kuksa_databroker/databroker/src/glob.rs
```diff
@@ -51,3 +52,113 @@ pub fn to_regex(glob: &str) -> Result<Regex, Error> {
...
+    // Construct regular expression
+
+    // Make sure we match the whole thing
+    // Start from the beginning
+    let mut re = String::from("^");
+
+    if input.ends_with(".*") {
```
argerus, line 85: If it ends with `.*` shouldn't other replacements be done anyway?

kuksa_databroker/databroker/src/glob.rs
```diff
@@ -51,3 +52,113 @@ pub fn to_regex(glob: &str) -> Result<Regex, Error> {
...
+        re.push_str(input.replace(".*", ".[^.]+").as_str());
+    } else if input.ends_with(".**") {
```
argerus, line 87: If it ends with `.**` shouldn't other replacements be done anyway?

kuksa_databroker/databroker/src/glob.rs
```diff
@@ -51,3 +52,113 @@ pub fn to_regex(glob: &str) -> Result<Regex, Error> {
     let re = to_regex_string(glob);
     Regex::new(&re).map_err(|_err| Error::RegexError)
 }
+
+fn has_asterisk_in_middle(input: &str) -> bool {
+    let input = input.trim(); // Remove leading and trailing whitespace
+    if input.starts_with('*') || input.ends_with('*') {
+        return false;
+    }
```
argerus, line 60: Can it not have an "asterisk in the middle" if it starts or ends with an asterisk?

argerus (changes requested):

kuksa_databroker/databroker/src/glob.rs
```diff
@@ -51,3 +52,113 @@ pub fn to_regex(glob: &str) -> Result<Regex, Error> {
...
+    if input.starts_with("*.") {
+        return false;
+    }
+    let pattern = r"^(?:\w+(?:\.\w+)*)(?:\.\*+)?$";
+    let regex = Regex::new(pattern).unwrap();
+
+    let pattern2 = r"^(?:[^.]+\.)*(?:\*{1,2})\.(?:[^.]+\.)*(?:[^.]+)*$";
+    let regex2 = Regex::new(pattern2).unwrap();
+
+    regex.is_match(input) || regex2.is_match(input)
```
argerus, line 130: Ok, so adding validation of paths is obviously a good thing (since this was not done at all before this PR).

The problem is that these two regex:es do not correctly identify whether a path is valid or not.

#### First regex
```rust
let pattern = r"^(?:\w+(?:\.\w+)*)(?:\.\*+)?$";
```
##### Commented
```rust
let pattern = r"(?x)
    # anchor at beginning
    ^
    # one occurrence of
    (?:
        # at least one word character
        \w+
        # followed by zero or more
        (?:       
            \.    # dot
            \w+   # at least one character
        )*
    )
    # followed by zero or one
    (?:
        \.        # dot
        \*+       # at least one asterisk
    )?
    # anchor at end
    $
";
```
This is pretty straight forward, but doesn't cover everything.
It matches paths consisting of word characters only and (optionally) ends with a single `.*`. E.g. `Vehicle.Cabin` or `Vehicle.Cabin.Sunroof.*`

#### Second regex
```rust
let pattern2 = r"^(?:[^.]+\.)*(?:\*{1,2})\.(?:[^.]+\.)*(?:[^.]+)*$";
```
##### Commented
```rust
let pattern = r"(?x)
    # anchor at beginning
    ^
    # zero or more
    (?:
        [^.]+     # at least one non-dot
        \.        # dot
    )*
    # followed by (exactly one)
    (?:
        \*{1,2}   # one or two asterisks
    )
    # followed by (exactly one)
    \.            # dot
    # followed by zero or more
    (?:
        [^.]+     # at least one non-dot
        \.        # dot
    )*
    # followed by any number of
    (?:
        [^.]+     # at least one non-dot
    )*
    #anchor at end
    $
";
```
The second regex is a lot harder to understand and I don't think it's matching the right things (maybe _because_ it's hard to follow the logic). For example, it will allow internal wildcards like `Vehicle.C*.Position` and any number of wildcards like `Vehicle.*.Sunroof.******` or wildcard paths containing invalid characters like `Vehicle.*.#@(&*@#(&#Position`. 

#### Suggested regex
My suggestion is to use something like what is found in `src/jwt/scope.rs` adopted to support `**` as well. Something like this:

##### Commented
```rust 
let re = regex::Regex::new(
    r"(?x)
    ^  # anchor at start (only match full paths)
    
    # At least one subpath (consisting of either of three):
    (?:
        [A-Z][a-zA-Z0-1]*  # An alphanumeric name (first letter capitalized)
        |
        \*                 # An asterisk
        |
        \*\*               # A double asterisk
    )

    # which can be followed by ( separator + another subpath )
    # repeated any number of times
    (?:
        \. # Separator, literal dot
        (?:
            [A-Z][a-zA-Z0-1]*  # Alphanumeric name (first letter capitalized)
            |
            \*                 # An asterisk
            |
            \*\*               # A double asterisk
        )
    )*
    $  # anchor at end (to match only full paths)
    ",
)
.unwrap();
```

or the short version:
```rust
r"^(?:[A-Z][a-zA-Z0-1]*|\*|\*\*)(?:\.(?:[A-Z][a-zA-Z0-1]*|\*|\*\*))*$"
```
(but I really think it's preferable to use the commented version, in the code, as it makes it easier to follow the intended logic, i.e. make the code readable)

kuksa_databroker/databroker/src/grpc/kuksa_val_v1/val.rs
```diff
@@ -54,55 +55,141 @@ impl proto::val_server::Val for broker::DataBroker {
...
+                                match broker
+                                    .get_entries_by_wildcards_with_regex(&regex.unwrap())
+                                    .await
+                                {
+                                    Ok(entries_result) => {
+                                        for data_entry in entries_result {
+                                            if request.view == 2
+                                                && data_entry.metadata.entry_type
+                                                    == broker::EntryType::Actuator
+                                            {
```
argerus, line 92: This is really weird?

```rust
if request.view == 2
    && data_entry.metadata.entry_type
        == broker::EntryType::Actuator
{
    // ... do stuff
} else if request.view != 2 {
    // ... do stuff
}
```

First off, `... do stuff` is _exactly_ the same in both branches (at least I think so).

Secondly, I don't know what `request.view == 2` means (by looking at it, I will now look it up the the proto/generated file). Instead, use the generated enum `proto::View::*` for these comparisons, which you get access to on the very next line:
```rust
let view = proto::View::from_i32(request.view)
```

Thirdly, by doing these conditional things based on `request.view` will probably lead to weird/erroneous behaviour. Requesting View::VIEW_TARGET_VALUE should be the equivalent of requesting Field::FIELD_ACTUATOR_TARGET (most probably).

argerus (changes requested):

kuksa_databroker/databroker/src/glob.rs
rafaeling: Yes it is. This is just to cover the special case of 'Vehicle.Cabin.Sunroof.*.Position'.

If we are looking for a single-level child that has a prefix 'Vehicle.Cabin.Sunroof' and ends with 'Position', the regex would be '^Vehicle\\.Cabin\\.Sunroof\\.Position'. The algorithm [below](https://github.com/eclipse/kuksa.val/pull/605/files#diff-f6fc96afb24437d8ff54b4f051da8a6f0b0224f466a088a7e14a188366a8a774R91), which replaces '*' with '[^.]+', does not cover that case. It matches everything in between leading to wrong results, for example: 'Vehicle.Cabin.Sunroof.Shade.Position'

kuksa_databroker/databroker/src/glob.rs
rafaeling: Following the use case "Vehicle.Cabin.Sunroof.*", if it is not replaced with ".[^.]+", it will return not only one-level child but any possible child, whose functionality belongs to the option "**".

kuksa_databroker/databroker/src/glob.rs
rafaeling: Same here, following the use case 'Vehicle.Cabin.Sunroof.\*\*', must be replaced with '.\*', so it returns any possible child under it.
The reason for these two if else statements is that the split and join per '.' algorithm implemented below doesn't cover the use cases than end with '.\*' or '.\*\*' adding extra '\\'

kuksa_databroker/databroker/src/glob.rs
rafaeling: It seems that such cases do not appear [here](https://github.com/boschglobal/kuksa.val/blob/feature/wildcard_matching_doc/doc/wildcard_matching.md)

kuksa_databroker/databroker/src/glob.rs
```diff
@@ -51,3 +52,113 @@ pub fn to_regex(glob: &str) -> Result<Regex, Error> {
...
+    if input.starts_with('*') || input.ends_with('*') {
+        return false;
+    }
+
+    let asterisk_position = input.find('*');
+    if let Some(pos) = asterisk_position {
+        if input[pos - 1..].contains(".*.") {
+            return true;
+        }
+    }
```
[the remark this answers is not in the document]
rafaeling: Yes you are right, don't remember why I did it this way

kuksa_databroker/databroker/src/glob.rs
```diff
@@ -51,3 +52,113 @@ pub fn to_regex(glob: &str) -> Result<Regex, Error> {
...
+        re.push_str(".*");
+    }
+
+    // And finally, make sure we match until EOL
+    re.push('$');
+
+    re
+}
+
+pub fn to_regex_new(glob: &str) -> Result<Regex, Error> {
```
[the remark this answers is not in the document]
rafaeling: Indeed, the intention was to adapt the existing function, I didn't expect for now that this PR would be also focused on the implementation, but just on the user-output so I didn't want to concern myself with any compatibility in the code for now. My idea was to clean up and refine the code once the output is accurate.

kuksa_databroker/databroker/src/glob.rs
```diff
@@ -51,3 +52,113 @@ pub fn to_regex(glob: &str) -> Result<Regex, Error> {
...
+    if let Some(pos) = asterisk_position {
+        if input[pos - 1..].contains(".*.") {
+            return true;
+        }
+    }
+
+    false
+}
+
+pub fn to_regex_string_new(glob: &str) -> String {
```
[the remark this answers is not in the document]
rafaeling: Same as [before](https://github.com/eclipse/kuksa.val/pull/605#discussion_r1301528640)

kuksa_databroker/databroker/src/glob.rs
argerus: I don't get it.

If the input is:
`Vehicle.Cabin.Sunroof.*.Position`

this code will replace it with:
`Vehicle.Cabin.Sunroof.Position`

That will match `Vehicle.Cabin.Sunroof.Position` (which it shouldn't) and not match `Vehicle.Cabin.Sunroof.Shade.Position` (which it should). Right?

kuksa_databroker/databroker/src/glob.rs
```diff
@@ -51,3 +52,113 @@ pub fn to_regex(glob: &str) -> Result<Regex, Error> {
...
+
+pub fn to_regex_new(glob: &str) -> Result<Regex, Error> {
+    let re = to_regex_string_new(glob);
+    Regex::new(&re).map_err(|_err| Error::RegexError)
+}
+
+pub fn matches_path_pattern(input: &str) -> bool {
+    if input.starts_with("*.") {
+        return false;
+    }
```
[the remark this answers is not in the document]
rafaeling: It was a bit hard to make a really generic regex that matches that case at this stage so that's why a decide to split for now into two regexes and this if clause. I do agree we need elegant regex solution that covers all use cases but since we change constantly opinions and perspectives, I didn't put to much effort into a good generic regex.

kuksa_databroker/databroker/src/broker.rs
```diff
@@ -882,6 +882,48 @@ impl<'a, 'b> DatabaseReadAccess<'a, 'b> {
...
+            }
+        }
+        if entries.is_empty() {
+            return Err(ReadError::NotFound);
+        }
+
+        Ok(entries)
+    }
+
+    pub fn get_entries_by_wildcards(
```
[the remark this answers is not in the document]
rafaeling: Yes, it is not needed anymore, will be removed

kuksa_databroker/databroker/src/glob.rs
argerus: Ok.
But, the solution for that use case should not break other use cases.

My point is that the code that is meant to replace wildcards in the middle of the pattern is never run if `input.ends_with(".*")`.

kuksa_databroker/databroker/src/grpc/kuksa_val_v1/val.rs
rafaeling: That was initially implemented with the intention to cover the 'getTargetValue' which provides the view with FIELD_ACTUATOR_TARGET value to just get actuators.
And yes, it should be refined, there is a repeated code.

kuksa_databroker/databroker/src/glob.rs
rafaeling: In the examples [table](https://github.com/boschglobal/kuksa.val/blob/feature/wildcard_matching_doc/doc/wildcard_matching.md) looks like this:
'Vehicle.Cabin.Sunroof.*.Position' returns 'Vehicle.Cabin.Sunroof.Position'

rafaeling:
One important thing to note about the permissions implementation: Previously, I assumed that, for instance, 'Vehicle.\*' granted access to everything underneath it. However, with the new '**to_regex_string**' implementation, we now need to use 'Vehicle.**' to cover everything, and 'Vehicle.*' for one level of permission."

kuksa_databroker/databroker/src/glob.rs
argerus: Well, that is just wrong (and in turn my mistake) :)

rafaeling:
Official PR for this feature was created -> https://github.com/eclipse/kuksa.val/pull/641

PR #609, opened 2023-07-20
Remove KUKSA-val gRPC client

erikbosch:
Code does not exist, so just remove from documentation. server side support still exists, that is ok but we should not promote it


erikbosch:
This could be a candidate for v0.4.0, no risk for regressions

lukasmittag (approved):
agree, that this should be removed

PR #611, opened 2023-07-20
Fixing build action triggers

erikbosch:
Rationale:
Kuksa-val build was triggered too often due to negative requirement and a spelling error
Clean-images is useless as we do not upload images tagged by PR

But then, should we not trigger when certs and proto files are changed as well (Even if some use symbolic links, do not know if build action is smart enough to follow symbolic links and check if they have been changed))

_Note: Not intended for v0.4!_


.github/workflows/kuksa_val_docker.yml
```diff
@@ -5,15 +5,13 @@ on:
...
-    - "kuksa_databroker/**"
-    - "kuksa_go_client/**"
-    - "kuksa_client/**"
-    - "Cargo.*"
-    - "Cross.toml"
+    paths:
+    - ".github/workflows/kuksa_val_docker.yml"
+    - "kuksa-val-server/**"
+    - "kuksa_certificates/**"
+    - "proto/**"
```
lukasmittag, line 12: do we need proto here? Server doesn't use it or am I wrong?

lukasmittag (changes requested):

.github/workflows/kuksa_val_docker.yml
erikbosch: Correct, good catch, it uses it's own proto file at https://github.com/eclipse/kuksa.val/blob/master/kuksa-val-server/protos/kuksa.proto. Will update

erikbosch:
> I think its my bad the comment above. we have a grpc side for server too. Forgot this

But that is not in the Proto folder, it is in https://github.com/eclipse/kuksa.val/blob/master/kuksa-val-server/protos/kuksa.proto

erikbosch:
The databroker failed build is related to #631

lukasmittag (approved):

SebastianSchildt:
Pls rebase to master  to see it runs through now

Issue #616, opened 2023-07-21
Remove bash dependency from shell scripts


d-s-e:
There are scripts in the repo, that have unnecessarily /bin/bash in the shebang line.
This adds a dependenciy that is not necessary.

PR #617, opened 2023-07-21
Fix bash dependency

d-s-e:
fixes https://github.com/eclipse/kuksa.val/issues/616


erikbosch:
What tests have you performed?

I get an error when trying to run `./genCerts.sh ` after your fix. With bash it works. Are possibly more changes needed in the file to avoid the bash dependency?

```
erik@debian3:~/kuksa.val/kuksa_certificates$ ./genCerts.sh 
./genCerts.sh: 27: Syntax error: "(" unexpected
```

erikbosch:
The newly added pre-commit check also complains, there seems to be a trailing space in one of the files. See https://github.com/eclipse/kuksa.val/blob/master/README.md#pre-commit-set-up on info how to set up pre-commit checks locally

SebastianSchildt:
Haven't checked, but maybe try a space before the " (", not sure POSIX has something to say about it.... When going the vanilla "sh" (which in your system might actually be anything), at least try with busybox, as that is at least a "likely" target ot be encountered in embedded environments

d-s-e:
Sorry, I didn't see that.
Will fix this.

d-s-e:
should work now.

SebastianSchildt:
I have been using `echo` all my life, and was wondering if it is really clever going for printf in shell here. Spoiler alert: It is. I've been **deep** in the rabbit hole just now, and wanted to share: https://unix.stackexchange.com/questions/65803/why-is-printf-better-than-echo/65819#65819

erikbosch (approved):
I ran both scripts locally on my Debian VM, they worked and no regressions.

@SebastianSchildt - I have no problems if this PR gets merged before v0.4.0, it does not affect any release artifacts apart from the source zip

PR #618, opened 2023-07-21
Change kuksa-client version handling

erikbosch:
The reason is that all docker builds on master previously got version v0.1.6 which gave odd behavior if downloading kuksa-client:master or even latest

**Current behavior**

If downloading `kuksa-client:master` and running it then it will report `Welcome to Kuksa Client version 0.1.6`
This also occurs if downloading [kuksa-client:latest](https://github.com/eclipse/kuksa.val/pkgs/container/kuksa.val%2Fkuksa-client/81633962?tag=latest)

**Expected behavior**

Either real tag or something that shows that it is not tagged.

With this PR in CI the version will be like:

`0.3.1.post84+git.3cba9d41`

i.e. 84 commits after v 0.3.1

I have not tested that it reports right version for tagged versions on official repository

**How to reproduce**

```
erik@debian3:~/kuksa.val/kuksa-client$ docker pull ghcr.io/eclipse/kuksa.val/kuksa-client:master
...
Status: Downloaded newer image for ghcr.io/eclipse/kuksa.val/kuksa-client:master

erik@debian3:~/kuksa.val/kuksa-client$ docker run --rm -it --net=host ghcr.io/eclipse/kuksa.val/kuksa-client:master --ip 127.0.0.1 --port 55555 --protocol grpc
Welcome to Kuksa Client version 0.1.6
```

### DIFF
--- a/.github/workflows/kuksa-client.yml
+++ b/.github/workflows/kuksa-client.yml
@@ -34,6 +34,9 @@ jobs:
 
     steps:
     - uses: actions/checkout@v3
+      with:
+        # Fetch everything to get tags working as expected
+        fetch-depth: 0
     - name: Docker meta
       id: meta
       uses: docker/metadata-action@v4
@@ -48,15 +51,15 @@ jobs:
           type=semver,pattern={{version}}
           type=semver,pattern={{major}}.{{minor}}
           type=semver,pattern={{major}}
-          
-    # only needed for runners without buildx setup, will be slow 
+
+    # only needed for runners without buildx setup, will be slow
     #- name: Set up QEMU
     #  uses: docker/setup-qemu-action@v2
-      
+
     #- name: Set up Docker Buildx
     #  id: buildx
     #  uses: docker/setup-buildx-action@v2
-    
+
     - name: Log in to the Container registry
       if: needs.checkrights.outputs.have_secrets == 'true'
       uses: docker/login-action@v2
@@ -94,7 +97,7 @@ jobs:
         push: false
         tags: "ttl.sh/kuksa.val/kuksa-client-${{github.sha}}"
         labels: ${{ steps.meta.outputs.labels }}
-        
+
 
   kuksa-client-test:
     runs-on: ubuntu-latest
@@ -118,4 +121,3 @@ jobs:
           cd kuksa-client
           pip install --upgrade build
           python -m build
-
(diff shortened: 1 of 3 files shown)


lukasmittag:
It seems that https://github.com/eclipse/kuksa.val/blob/master/kuksa-client/kuksa_client/_metadata.py is doing the right thing. I think it originates from here: https://github.com/eclipse/kuksa.val/blob/58dd5d1c46cc337e64652b69601dea84fb8f64ff/kuksa-client/pyproject.toml#L10C1-L12C36 But i do not know how this works.

erikbosch:
> It seems that https://github.com/eclipse/kuksa.val/blob/master/kuksa-client/kuksa_client/_metadata.py is doing the right thing. I think it originates from here: https://github.com/eclipse/kuksa.val/blob/58dd5d1c46cc337e64652b69601dea84fb8f64ff/kuksa-client/pyproject.toml#L10C1-L12C36 But i do not know how this works.

In general the tag filtering seems to work, like when creating a PyPi package. The filtering however does not work. If I understand it correctly the intention is that it shall filter out tags that does not have the wanted syntax, but that does not work and you get a warning, like in the example below where it selects a local tag even if it does not match the filter completely. That is why I removed the tag filter.
```
user@debian:~/kuksa.val/kuksa-client$ python -m setuptools_git_versioning
[2023-04-13 10:51:37,352]  WARNING: tag_filter is not a valid function reference: Wrong reference name: [0-9]+.[0-9]+.[0-9]+
[2023-04-13 10:51:37,481]  WARNING: tag_filter is not a valid function reference: Wrong reference name: [0-9]+.[0-9]+.[0-9]+
0.4.0a1
```

But the real problem is related to how we build the docker image where we had two problems. First we did just a shallow check-out, so we did not get the tags, secondly we copied some directories (but nor `.git`) so we lost all tag information. That is why the docker container for 0.3.1 still reports version 0.1.6. Hopefully, with this PR, when we push 0.4.0 to master it will be able to pick up that version, without any "post" or "dirty" segments

SebastianSchildt:
Checked, works in master now after the merge. "Fetch depth 0" is not very efficient, but currently the only option, there is theoretically a fix here https://github.com/actions/checkout/pull/579 but that guy has been ignored by MS/Github for more than a year. I admire his patience and perseverance.... Anyway, just a reminder, we want to take a look again if hell freezes over.

PR #627, opened 2023-07-27
Update databroker versions to 0.4.0

erikbosch:
Also doing "cargo update" to update *.lock file to get databroker* version updated there as well


erikbosch:
It is queued so not sure license check will pass, but please have a look if changes look ok if the build pass @SebastianSchildt  and @lukasmittag

PR #629, opened 2023-07-28
Testing KUKSA Server Unittest

erikbosch:
This aligns unit-tests with other triggers and now it works. It also removes explicit install of gRPC and Boost. That is included from cmake so not needed. The old warning that it might be needed for "older" versions of cmake shall likely not be relevant any longer, at least not if we assume that Ubuntu 20 or newer is used. 

Still the build takes time. A possible remedy for CI could be to join both builds to the same so we do not need to install boost and gRPC twice.

Now also using 4 threads (`-j 4` in all builds)

### DIFF
--- a/.github/workflows/kuksa_val_unittest.yml
+++ b/.github/workflows/kuksa_val_unittest.yml
@@ -2,24 +2,28 @@ name: kuksa_val_unittests
 
 on:
   push:
-    branches: [ disabled ]
-    tags:
-      - '0.*'
+    branches: [ master ]
   pull_request:
-    branches: [ disabled ]
-    
+    branches: [ master ]
+    paths:
+    - ".github/workflows/kuksa_val_unittest.yml"
+    - "kuksa-val-server/**"
+    - "kuksa_certificates/**"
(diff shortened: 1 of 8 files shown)


.github/workflows/kuksa_val_unittest.yml
```diff
@@ -2,24 +2,28 @@ name: kuksa_val_unittests
...
+    branches: [ master ]
   pull_request:
-    branches: [ disabled ]
-    
+    branches: [ master ]
+    paths:
+    - ".github/workflows/kuksa_val_unittest.yml"
+    - "kuksa-val-server/**"
+    - "kuksa_certificates/**"
+    - "proto/**"
```
lukasmittag, line 12: this is the same issue as in #611 not the right location to look for proto files

lukasmittag (commented):
maybe need to open the PR again.

.github/workflows/kuksa_val_unittest.yml
lukasmittag: fixed by #633

PR #630, opened 2023-07-31
Fix protoc install vulnerability

erikbosch:
See https://cwe.mitre.org/data/definitions/22.html

Theoretical risk in short:
- We download a malicious zip archive
- The zip archive contains items like "../../usr/bin/something"
- We might end up with installing unwanted files at wrong places

Tested that it does not affect normal install
This was detected when running Github code scanning in my own fork

### DIFF
--- a/kuksa_go_client/protocInstall/protocInstall.go
+++ b/kuksa_go_client/protocInstall/protocInstall.go
@@ -8,6 +8,7 @@ import (
 	"os"
 	"path/filepath"
 	"runtime"
+	"strings"
 )
 
 const (
@@ -50,8 +51,11 @@ func UnzipSource(source, destination string) error {
 	}
 
 	for _, f := range reader.File {
+		if strings.Contains(f.Name, "..") {
+			fmt.Println("Ignoring path traversal elements (CWE-22): ", f.Name)
+			continue
+		}
 		filePath := filepath.Join(destination, f.Name)
-
 		if f.FileInfo().IsDir() {
 			err := os.MkdirAll(filePath, os.ModePerm)
 			if err != nil {


kuksa_go_client/protocInstall/protocInstall.go
```diff
@@ -50,8 +51,11 @@ func UnzipSource(source, destination string) error {
 	}
 
 	for _, f := range reader.File {
+		if strings.Contains(f.Name, "..") {
```
lukasmittag, line 54: why .. ? Moving one directory up and installing something weird? I think its not that big of an issue since we get the source from the official website.

lukasmittag (approved):
Codewise seems good. Just one minor comment.

kuksa_go_client/protocInstall/protocInstall.go
erikbosch: As always with security topics it is always a question on how easy it is to take advantage of a vulnerability. We use hard-coded links like `https://github.com/protocolbuffers/protobuf/releases/download/v3.19.4/protoc-3.19.4-linux-x86_32.zip` and https,  but if an attacker either manages to get write access to the protocolbuffers Github account, or manages to do some form of Man-In-The-Middle attack he could theoretically replace the content with something else.

(As we specify specific version we could theoretically do as in https://github.com/eclipse/kuksa.val/blob/master/kuksa-val-server/boost.cmake and add a SHA256 checksum to make sure we get what we expect)

Here one can argue that the severity of the vulnerability is limited, it might be difficult to find a "good business case" for an attacker, but as it was reported by code scanning, mentioned in https://cwe.mitre.org/data/definitions/22.html and is easy to fix I think it is better to fix it than to dismiss it.

PR #631, opened 2023-07-31
Add BCD-3-Clause support

### LINKED ISSUE #632: Cargo license check needs to be extended to cover BSD-3-Clause
We get an error for the BOM check in https://github.com/eclipse/kuksa.val/tree/master/kuksa_databroker/createbom as a new matchit version include BSD3

https://crates.io/crates/matchit/0.7.1

Metadata:  MIT AND BSD-3-Clause 

https://crates.io/crates/matchit/0.7.0

Metadata:  MIT

(As a side note it seems that the root Cargo.lock does not lock all dependencies to a specific version, the lock file specify v0.7.0 but in license check we get 0.7.1)

erikbosch:
Fixes #632


kuksa_databroker/createbom/createbom.py
```diff
@@ -117,7 +117,7 @@ def generate_bom(source_path, target_path, dashout):
     errors = []
     for crate in crates:
         try:
-            print(f"License for {crate['name']}: ", end="")
+            print(f"License for {crate['name']} {crate['version']}: ", end="")
```
erikbosch, line 120: This is to get output with version like:

`2023-07-31T16:02:41.7150426Z #0 6.859 License for matchit 0.7.1: MIT.txt BSD-3-Clause.txt`

Reason is to be able to compare successful/unsuccessful builds and see if the error because of an updated version.

lukasmittag (commented):
lgtm

PR #638, opened 2023-08-11
make tls stuff optional

AkhilTThomas:
the tonic feature "_tls_" is made optional.

Behavior:
`cargo build --no-default-features --features=no-tls` => builds databroker without tonic/tls or rust-tls dependencies
`cargo build or cargo build --features=tls` => build with tls features.

Dependencies can be verified by running and comparing
`cargo tree --features no-tls --no-default-features > feat.txt`
`cargo tree --features tls > feat_tls.txt`

### DIFF
--- a/kuksa_databroker/databroker/Cargo.toml
+++ b/kuksa_databroker/databroker/Cargo.toml
@@ -24,12 +24,7 @@ path = "src/lib.rs"
 
 [dependencies]
 databroker-proto = { workspace = true }
-tonic = { workspace = true, features = [
-    "transport",
-    "channel",
-    "prost",
-    "tls",
-] }
+tonic = { workspace = true, optional = true }
 prost = { workspace = true }
 prost-types = { workspace = true }
 tokio = { workspace = true, features = [
@@ -60,6 +55,9 @@ jemallocator = { version = "0.5.0", optional = true }
 lazy_static = "1.4.0"
 
 [features]
+default = ["tls"]
+tls = ["tonic/transport", "tonic/channel", "tonic/prost", "tonic/tls"]
+no-tls = ["tonic/transport", "tonic/channel", "tonic/prost"]
 # to enable jemalloc use --features jemalloc
 jemalloc = ["dep:jemallocator"]
 libtest = []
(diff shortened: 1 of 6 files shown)


kuksa_databroker/databroker/Cargo.toml
```diff
@@ -60,6 +55,9 @@ jemallocator = { version = "0.5.0", optional = true }
 lazy_static = "1.4.0"
 
 [features]
+default = ["tls"]
+tls = ["tonic/transport", "tonic/channel", "tonic/prost", "tonic/tls"]
+no-tls = ["tonic/transport", "tonic/channel", "tonic/prost"]
```
argerus, line 60: Get rid of the `no-tls` feature, as features should never disable functionality.

Instead, rework the `tls` feature to add anything needed for TLS, i.e. enabling TLS.

See:
https://doc.rust-lang.org/cargo/reference/features.html#feature-unification

> [...] features should be additive. That is, enabling a feature should not disable functionality, and it should usually be safe to enable any combination of features. [...]

> For example, if you want to optionally support [no_std](https://doc.rust-lang.org/reference/names/preludes.html#the-no_std-attribute) environments, do not use a no_std feature. Instead, use a std feature that enables std.

There's even a clippy warning for this:
https://github.com/rust-lang/rust-clippy/pull/7539

argerus (changes requested):

kuksa_databroker/databroker/Cargo.toml
AkhilTThomas: @argerus Thanks you for the comment, the changes are done.

argerus (approved):

argerus (changes requested):
Looks better :)

I just have one more suggestion, which is that command line arguments relating to TLS are removed when the feature is not enabled.

I implemented this on top of your branch, if you want to incorporate that commit directly.

* branch: https://github.com/boschglobal/kuksa.val/tree/argerus/optional_tls
* PR: https://github.com/AkhilTThomas/kuksa.val/pull/1

AkhilTThomas:
> Looks better :)
> 
> I just have one more suggestion, which is that command line arguments relating to TLS are removed when the feature is not enabled.
> 
> I implemented this on top of your branch, if you want to incorporate that commit directly.
> 
>     * branch: https://github.com/boschglobal/kuksa.val/tree/argerus/optional_tls
> 
>     * PR: [Remove TLS args when feature not enabled AkhilTThomas/kuksa.val#1](https://github.com/AkhilTThomas/kuksa.val/pull/1)

I have pulled your changes :)

argerus (approved):

PR #639, opened 2023-08-15
Add wildcard matching doc

argerus:


SebastianSchildt:
(Readable: https://github.com/boschglobal/kuksa.val/blob/feature/wildcard_matching_doc/doc/wildcard_matching.md )

Look quite sensible to me.  Apart from content, some notes:

 Once we have plumbing for that in databroker we should also make gherkin/cucumber tests for it, the table in markdown looks already almost "executable"

Also, instead of just having this file "waiting to be found",  we should then reference it from the authorisation.md document and in the API doc if we *ahem* had any (we do at least have a little bit in wiki https://github.com/eclipse/kuksa.val/wiki/KUKSA.val-gRPC-interfaces)

SebastianSchildt:
Looks good to me, but should only be merged after #641  

Should explicitly mention that the expansion rules are applied in the kuksa.val.v1 API paths as well as Expansion of authorisation tokens

PR #641, opened 2023-08-23
Add discoverability feature to DataBroker

rafaeling:
PR that implements the following examples [table](https://github.com/boschglobal/kuksa.val/blob/feature/wildcard_matching_doc/doc/wildcard_matching.md) for discoverability feature.


kuksa_databroker/databroker/src/broker.rs
```diff
@@ -882,6 +882,24 @@ impl<'a, 'b> DatabaseReadAccess<'a, 'b> {
...
     }
 
+    pub fn get_entries_by_regex(&self, regex: regex::Regex) -> Result<Vec<Entry>, ReadError> {
+        let mut entries: Vec<Entry> = Vec::new();
+        for key in self.db.path_to_id.keys() {
+            if regex.is_match(key) {
+                entries.push(
+                    self.get_entry_by_id(self.db.path_to_id.get(key).unwrap().to_owned())
+                        .unwrap()
+                        .to_owned(),
```
argerus, line 892: As a general rule, there should be no `unwrap()` in databroker. We want to propagate the errors instead of crashing the whole process.

And there is no need to use `unwrap()` here that I can see. Just iterate through the entries directly, instead of the separate `path_to_id` map.

kuksa_databroker/databroker/src/glob.rs
argerus, line 1: General observation.

`pub fn to_regex(glob: &str) -> Result<Regex, Error>`

still doesn't behave according to the matching rules in the [doc](https://github.com/boschglobal/kuksa.val/blob/feature/wildcard_matching_doc/doc/wildcard_matching.md).

It's good that you've added unit tests for

`to_regex_string(glob: &str)`

but these unit tests only check if the produced regex pattern matches the expected _pattern_. It doesn't verify that the pattern is fit for purpose, i.e. that it matches signals according to the rules. Some details have already been pointed out in the previous PR, but before I start listing more, I have a general question:

* Should we continue to refine the glob matching until it works according to the rules?
or
* Should we use the old `to_regex(glob: &str)` in the meantime -  which only supported `*` and would match all children of branches - for the purpose of this PR.

The second option would be sufficient in order to support the "discovery" needed by auto completion for example.

kuksa_databroker/databroker/src/glob.rs
```diff
@@ -51,3 +62,131 @@ pub fn to_regex(glob: &str) -> Result<Regex, Error> {
...
+                |
+                \*                 # An asterisk
+                |
+                \*\*               # A double asterisk
+            )
+        )*
+        $  # anchor at end (to match only full paths)
+        ",
+    )
+    .unwrap();
```
argerus, line 93: Tip. Instead of building the regex every time this function is called, it's better to put it in a static variable using [lazy_static!](https://docs.rs/lazy_static/latest/lazy_static/) which would only be initialized once.

E.g.
```rust
use lazy_static::lazy_static;

lazy_static! {
    static ref REGEX_VALID_PATTERN: regex::Regex = regex::Regex::new(
        r"(?x)

        ...

        ",
    )
    .unwrap();
}

```

kuksa_databroker/databroker/src/grpc/kuksa_val_v1/val.rs
```diff
@@ -54,55 +55,125 @@ impl proto::val_server::Val for broker::DataBroker {
...
+                                            error: Some(proto::Error {
+                                                code: 404,
+                                                reason: "not_found".to_owned(),
+                                                message: "No entries found for the provided path"
+                                                    .to_owned(),
+                                            }),
+                                        });
+                                    }
+                                    Err(ReadError::PermissionExpired) => {}
+                                    Err(ReadError::PermissionDenied) => {}
```
argerus, line 141: Not handled? Not needed?

argerus (changes requested):

kuksa_databroker/databroker/src/glob.rs
argerus: And if _all_ we need from this PR is to support discovery of metadata (to support autocompletion), we don't actually need any support for globs at all.

We would just need to change that a request for an empty path would mean "all paths".

SebastianSchildt: My view - from just following here, I am no deep in the implementation of this PR - we should implement the matching according to the doc, becasue
 - it is for the first time "Documented"
 - It will be similar and useful for API queries as well as tokens
 - We should have (maybe not directly in this PR) some tests, that do exactly that:  Checking all examples from the doc (because that is hopefully where will also document new corner cases if we find them)

I say this, because I feel "How hard can it be" (Doc does not look that complicated), however maybe I am wrong.
NOT doing so, and doing a "half" solution of somethign already there may surely work but it has disadavantages

 - we are again "spec-less" and the knowledge is in the head of people/code only
 - If we still keep the document it agains becomes some sort of "ah this is what wie intend eventually", but beware it is not there yet Id o think it supports inly xxx", that is not good
 - Now you are deep into it, how likely is it we will  without any effort - pick up at this point exactly in a few weeks?


But question really is, are there unsurmountable "technical difficulties" in implementing this https://github.com/boschglobal/kuksa.val/blob/feature/wildcard_matching_doc/doc/wildcard_matching.md ?

kuksa_databroker/databroker/src/glob.rs
SebastianSchildt: if _Lazy_ static is doing what I think it does, is an "unwrap()" ok here, according to your no  `unwrap()` rule? I understand, that for stuff  during initialisation it is fine (because well, either it works, or databroker does not start), but wouldn't the lazy here indicate this might be executed anytime later during runtime? Or we can force the issue by using it already once somewhere during init?

argerus: > if Lazy static is doing what I think it does
It is

> is an "unwrap()" ok here
Good point! Generally, no. But as the input for the regex will _never_ change (at least not at runtime) we can guarantee<sup>1</sup> that it will always succeed.

Now that you point it out, it should probably be changed to
```diff
--    .unwrap();
++    .expect("regex compilation (of static pattern) should always succeed")
```
or something. We could also force [initialization](https://docs.rs/lazy_static/latest/lazy_static/fn.initialize.html) of the static at startup (instead of first use).

`1` Guaranteed because it's used in our unit tests.

kuksa_databroker/databroker/src/glob.rs
rafaeling: I've given it another chance before going for second option, I'm a bit stubborn. I have made some changes that hopefully will cover all cases now.

argerus: It's a lot easier to follow what is going on now!

And it works a lot better as well (perhaps because of it ;) )

I noticed a couple of (hopefully easily fixable) things not working:

1. An empty path doesn't match everything (an early return for that case perhaps).
2. It doesn't handle if the input is the full path of a signal (which is not a branch).
3. It doesn't handle a single wildcard, i.e. "`*`". Limitied practical implication given the current structure of VSS...

I'll push some unit tests as well that you can use / extend.

> it's not enough with simple substring replacement.

So, perhaps I was wrong! Substring replacement might be enough.

kuksa_databroker/databroker/src/grpc/kuksa_val_v1/val.rs
```diff
@@ -54,55 +55,125 @@ impl proto::val_server::Val for broker::DataBroker {
...
+                                let fields = combine_view_and_fields(view, fields);
+                                debug!("Getting fields: {:?}", fields);
+                                let proto_entry = proto_entry_from_entry_and_fields(entry, fields);
+                                debug!("Getting datapoint: {:?}", proto_entry);
+                                entries.push(proto_entry);
+                            }
+                            Err(ReadError::NotFound) => {
+                                let sub_path_new = request.path.clone();
+                                let regex = glob::to_regex(sub_path_new.as_ref());
+                                match broker.get_entries_by_regex(regex.unwrap()).await {
```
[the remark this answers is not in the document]
rafaeling: Done

kuksa_databroker/databroker/src/grpc/kuksa_val_v1/val.rs
```diff
@@ -54,55 +55,156 @@ impl proto::val_server::Val for broker::DataBroker {
...
-                        errors.push(proto::DataEntryError {
-                            path: request.path,
-                            error: Some(proto::Error {
-                                code: 401,
-                                reason: "unauthorized".to_owned(),
-                                message: "Authorization expired".to_owned(),
-                            }),
-                        });
+                match glob::is_valid_pattern(&request.path) {
+                    true => {
```
argerus, line 59: It's probably better to "short circuit" if the path is invalid, to avoid at least one level of nesting. I.e.

```rust
for request in requested {
    if glob::is_valid_pattern(&request.path) {
        errors.push(proto::DataEntryError {
            path: request.path,
            error: Some(proto::Error {
                code: 400,
                reason: "bad_request".to_owned(),
                message: "Bad Path Request".to_owned(),
            }),
        });
        continue;
    }
```

(and as a side note, matching on `true` / `false` is not idiomatic Rust as that's what `if` / `else` are for... But you quickly get used to using `match` for everything in Rust :) )

kuksa_databroker/databroker/src/grpc/kuksa_val_v1/val.rs
```diff
@@ -54,55 +55,156 @@ impl proto::val_server::Val for broker::DataBroker {
...
+                                                        reason: "forbidden".to_owned(),
+                                                        message: "Permission denied".to_owned(),
+                                                    }),
+                                                });
+                                            }
+                                        }
+                                    }
+                                    Err(_error) => {
+                                        println!("Error: Invalid Regex");
+                                    }
```
argerus, line 172: Return error to client?

argerus (changes requested):

kuksa_databroker/databroker/src/grpc/kuksa_val_v1/val.rs
rafaeling: You are right, should not return anything

kuksa_databroker/databroker/src/grpc/kuksa_val_v1/val.rs
argerus: The update still nests everything in a big if / else. Still think it would be better to short circuit at the top of the `for request in requested` loop.

It's probably better to "short circuit" if the path is invalid, to avoid at least one level of nesting. I.e.

```rust
for request in requested {
    if !glob::is_valid_pattern(&request.path) {
        errors.push(proto::DataEntryError {
            path: request.path,
            error: Some(proto::Error {
                code: 400,
                reason: "bad_request".to_owned(),
                message: "Bad Path Request".to_owned(),
            }),
        });
        continue;
    }

    let view = ...;
    let fields = ...;
    match broker.get_entry_by_path(&request.path).await {
```

rafaeling: Done

argerus (approved):
I haven't actually tested it, but it looks good to me now.

PR #642, opened 2023-08-28
Add (optional) VISSv2 interface

argerus:
Add VISSv2 interface to databroker.
Gated by feature flag `viss` (not built by default).

Supports:
- `get`, `set`, `subscribe` and `unsubscribe`
- Authorization (using same jwt scope format as the other interfaces)

Not implemented:
- Filters
- TLS

### DIFF
--- a/doc/protocol/support.md
+++ b/doc/protocol/support.md
@@ -1,60 +1,81 @@
-# What do the KUKSA.val Server and databroker support?
-This file contains an overview what the KUKSA.val Server and databroker each supports. It focuses on gRPC and VISS support and also what feeders are supported.
-
 ## Supported protocols
 
-
-| Protocol   |      KUKSA.val server      |  KUKSA.val databroker   |
-|------------|:-----------------:|----------:|
-| VISS V1    |      o           |     o     |
-| VISS V2    |     x/o          |     o     |
-| gRPC       |      x           |     x     |
-
-x = supported; x/o = partly supported; o = not supported
+This file contains an overview what the KUKSA.val Server and databroker each supports. It focuses on gRPC and VISS support and also what feeders are supported.
 
 
-### Server and VISS 
+| Protocol                  | KUKSA.val server | KUKSA.val databroker |
+|---------------------------|:----------------:|:--------------------:|
+| VISS V1                   |         -        |           -          |
+| VISS V2                   |        x/-       |          x/-         |
+| gRPC (kuksa)              |         x        |           -          |
+| gRPC (kuksa.val.v1)       |         -        |           x          |
+| gRPC (sdv.databroker.v1)  |         -        |           x          |
 
-KUKSA.val server supports the semantic of [VISS v1](https://www.w3.org/TR/vehicle-information-service/) using the new syntax of [VISS v2](https://www.w3.org/TR/viss2-core/).
+x = supported; x/- = partially supported; - = not supported
 
-KUKSA.val only supports the Websocket transport defined in VISS v2.
 
-Support of VISSv2:
+### VISSv2 support (websocket transport)
 
-### Security model
-KUKSA.val does not support the VISS V2 security model and currently we are not planning to support. KUKSA.val server does support authenticated access to VSS resources. For details check [here.](../KUKSA.val_server/jwt.md).
+| Feature                       | KUKSA.val server  | KUKSA.val databroker |
+|-------------------------------|:-----------------:|:--------------------:|
+| Read                          |                   |                      |
+|   - Authorized Read           | x<sup>1,2</sup>   |           x          |
+|   - Search Read               | -                 |           -          |
+|   - History Read              | -                 |           -          |
+|   - Static Metadata Read      | -                 |           x          |
+|   - Dynamic Metadata Read     | -                 |           -          |
+| Update                        |                   |                      |
+|   - Authorized Update         | x<sup>1,2</sup>   |           x          |
+| Subscribe                     |                   |                      |
+|   - Authorized Subscribe      | x<sup>1,2</sup>   |           x          |
+|   - Curve Logging Subscribe   | -                 |           -          |
+|   - Range Subscribe           | -                 |           -          |
+|   - Change Subscribe          | -                 |           -          |
+| Unsubscribe                   | x                 |           x          |
+| Subscription                  | x                 |           x          |
+| Error messages                | x                 |           x          |
+| Timestamps                    | x                 |           x          |
 
-### Supported VISS (v2) calls
+x = supported
 
-| method                        |   Support                     | Comment                                                                                                           |
-|-------------------------------|:-----------------------------:|-------------------------------------------------------------------------------------------------------------------|
-| Read                          |            x/o                ||
-|   - Authorized Read           |             x                 | Authorization must be performed as standalone call, do not support "in-lining" authorization in read call         |
-|   - Search Read               |             o                 ||
-|   - History Read              |             o                 ||
-|   - Signal Discovery Read     |             o                 ||
-|   - Dynamic Metadata Read     |             o                 ||
-| Update                        |             x                 ||
-|   - Authorized Update         |             x                 | Authorization must be performed as standalone call, do not support "in-lining" authorization in update call       |
-| Subscribe                     |             x                 ||
-|   - Authorized Subscribe      |             x                 | Authorization must be performed as standalone call, do not support "in-lining" authorization in subscribe call    |
-|   - Curve Logging Subscribe   |             o                 ||
-|   - Range Subscribe           |             o                 ||
-|   - Change Subscribe          |             o                 ||
-| Unsubscribe                   |             x                 ||
-| Subscription                  |             x                 ||
-| Error messages                |             x                 ||
-| Timestamps                    |             x                 ||
+x<sup>1</sup> Authorization is done using a non-standard standalone call which is incompatible with standards compliant clients.
 
-x = implemented; x/o = partly implemented; o = not implemented
+x<sup>2</sup> Relies on the non-standard `attribute` values which doesn't work with standards compliant clients.
 
 For a more detailed view of the supported JSON-schemas [click here](https://github.com/eclipse/kuksa.val/blob/master/kuksa-val-server/include/VSSRequestJsonSchema.hpp)
 
+### VISSv2 in KUKSA.val server
+KUKSA.val server supports the semantics of [VISS v1](https://www.w3.org/TR/vehicle-information-service/) using the new syntax of [VISS v2](https://www.w3.org/TR/viss2-core/). It implements a modified version of VISSv2 which introduces the concept of `attributes` which makes it incompatible with standards compliant VISSv2 clients.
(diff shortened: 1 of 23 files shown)


lukasmittag:
Isn't it already implemented through https://github.com/eclipse-kuksa/kuksa-viss ?

argerus:
> Isn't it already implemented through https://github.com/eclipse-kuksa/kuksa-viss ?

Sure. But that requires running a separate service (written in python). Useful for demonstration purposes, but not very resource efficient or easy to make "production ready".

With that said, I haven't really evaluated the need for a VISSv2 service, just wanted to check the feasibility of doing it (and took it up as a fun weekend project :) ).

Conclusion is that it seems very feasible. And apart from the lower footprint, it would IMHO also be a lot easier to make production ready thanks to the type safety / thread safety provided by Rust.

kuksa_databroker/databroker/src/main.rs
```diff
@@ -296,6 +299,41 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
...
     }
 
+    #[cfg(feature = "viss")]
+    {
+        parser = parser
+            .arg(
+                Arg::new("enable-viss")
+                    .display_order(30)
+                    .long("enable-viss")
+                    .help("Enable VISSv2 service")
```
erikbosch, line 309: Should we possibly mention "websocket", like "Enable VISSv2 websocket service", as VISS include specifications for HTTPS, Websocket and MQTT, but it is only websocket we support.

erikbosch:
Should we not add a short documentation somewhere. Like how to build it, how to test it. What limitations are there. Shall it for now be considered as "alpha" (see https://github.com/eclipse/kuksa.val/wiki/KUKSA.val-Component-Maturity), i.e. not suitable for production use? I.e. for now we will not test the interface as part of release testing?

How much have been tested? Have we for instance tested with kuksa-client (python) using websocket towards the Databroker VISS interface?

argerus:
> How much have been tested? Have we for instance tested with kuksa-client (python) using websocket towards the Databroker VISS interface?

It has been tested with kuksa-client. Feel free to give it a go.

Since kuksa-client isn't fully VISSv2 compliant (or compliant with any VISS version really?), a small change is needed in kuksa-client to support VISSv2 authorization (I have a local branch with that).

> Should we not add a short documentation somewhere. Like how to build it, how to test it. What limitations are there.

If anyone is up for that, sure. This was just a "weekend project" which I think would be a useful addition to the project. But it is not a priority at the moment.

With that said, my impression is that it would - even in it's current state - be a better option for "supporting VISS" than the available alternatives.

* It implements standards compliant VISSv2 (which I don't think is the case with kuksa-val-server)
* It seems to be "more" standards compliant and feature complete than kuksa-viss referenced above.
* It would be a more realistic thing to make production ready at some point.

And as the whole thing is gated behind a feature flag, one option would be to have it "hidden" for now, but easily enabled by `cargo build --features viss` for anyone who wants to. If we decide it would be a useful thing to have going forward, documentation etc can be added.

I do think it would be useful to merge though, as that would prevent the code from rotting.

kuksa_databroker/databroker/src/main.rs
argerus: done

erikbosch:
I support merging. One of the problem with VISSv2 is the lack of (commonly used) reference implementations. One which possibly could be considered as the "official one" exists at https://github.com/w3c/automotive-viss2, and if VISSv2 gets higher priority it could be an idea to verify that "their solution" and "our solution" interprets the VISSv2 standard in the same way.

kuksa_databroker/databroker/src/viss/v2/conversions.rs
```diff
@@ -0,0 +1,372 @@
...
+            broker::DataType::Int16 => "int16",
+            broker::DataType::Int32 => "int32",
+            broker::DataType::Int64 => "int64",
+            broker::DataType::Uint8 => "uint8",
+            broker::DataType::Uint16 => "uint16",
+            broker::DataType::Uint32 => "uint32",
+            broker::DataType::Uint64 => "uint64",
+            broker::DataType::Float => "float",
+            broker::DataType::Double => "double",
+            broker::DataType::Timestamp => "timestamp",
```
erikbosch, line 356: Not specifically related to this PR but timestamp/timestamp[] was removed from VSS some time ago, so we could possibly remove it as well from KUKSA

erikbosch (approved):
Looks ok, as it is intended as an experimental/optional feature I think it can be merged without exhaustive testing.

SebastianSchildt:
I am still not sure, what the "correct" way is to move forward here (in terms of maintainability/usage), but letting this rot would also be sad. Moving this behind a feature flag for now is a good first step. I think to merge this, there are few other things we need to do and should not do

 - Even when compiled (with feature flag), it should need to be enabled explicitly by command line argument and when that is enabled should print a warning along the lines of "You enabled VISS; this is an experimental feature"
    -  The rationale here is, _if_ we would - at a later point -  decide to include this in our default/released builds and something "is wrong", i.e. not secure, or can crash/confuse db, it would not be running by default. And it is a good idea anyway not unnecessarily exposing additional network ports
  - Write a small doc how to enable it _and_ what VISS is supported: Not a "book", but I would expect something like "VISS get call, no history, no filters, no wildcards, and some example message (like in the VISS transport spec does)
  - Haven't checked here, but in general: Do NOT extend VISS. Have a ssubset of VISS, not something that somehow "overlaps", i.e. we do not need to be able to do providers (i.e. set sensors) in VISS. Let's keep this an "application" protocol as ~~god~~ ~~W3C~~ Ulf intended it to be

argerus:
> * Even when compiled (with feature flag), it should need to be enabled explicitly by command line argument and when that is enabled should print a warning along the lines of "You enabled VISS; this is an experimental feature"

That's how it's implemented now (except for the warning). Even if the `viss` feature is enabled, no "VISSv2" code will run unless `--enable-viss` is also supplied as an argument at startup.

> Haven't checked here, but in general: Do NOT extend VISS.

The code is strictly following the standard (barring bugs). And I agree that it should continue to do so.

> Write a small doc how to enable it and what VISS is supported

That is missing, but I can add it.

kuksa_databroker/databroker/src/viss/v2/conversions.rs
argerus: I just removed them here. They were not supported by databroker anyway and would just throw an error if anyone tried to use them.

argerus:
- Rebased.
- Added documentation.
- Included VISSv2 support for kuksa-client (which lived in a separate branch previously).
  VISSv2 in kuksa-client is only enabled if the negotiated subprotocol is VISSv2. If no subprotocol is negotiated, it reverts to it's previous behaviour in order to remain backwards compatible with kuksa-val-server which isn't compatible with VISSv2 (and doesn't negotiate a subprotocol).

SebastianSchildt:
@erikbosch  Can you check this (again). Also I assume we want to merge it fast, but after "release"/tag?

argerus:
> Also I assume we want to merge it fast, but after "release"/tag?

Before or after release tag shouldn't matter as it's not included in the release artifacts anyway.

doc/protocol/support.md
```diff
@@ -1,68 +1,87 @@
...
-
+## Supported protocols
+
+This file contains an overview what the KUKSA.val Server and databroker each supports. It focuses on gRPC and VISS support and also what feeders are supported.
+
+
+| Protocol                  | KUKSA.val server | KUKSA.val databroker |
+|---------------------------|:----------------:|:--------------------:|
+| VISS V1                   |         -        |           -          |
+| VISS V2                   |        x/-       |          x/-         |
```
erikbosch, line 9: Isn't it a bit confusing that we in kuksa-client code specifies the two subprotocols:

`subprotocols = ["VISSv2", "VISSv1"]`

but here we say that both server and databroker supports (parts of) VISSv2

erikbosch: Would it be better to call them (in kuksa-client) something like "VISSv2" and "VISSv2-extended" or something similar.

doc/protocol/support.md
```diff
@@ -1,68 +1,87 @@
...
+
+x = supported; x/- = partially supported; - = not supported
+
+
+### VISSv2 support (websocket transport)
+
+| Feature                       | KUKSA.val server  | KUKSA.val databroker |
+|-------------------------------|:-----------------:|:--------------------:|
+| Read                          |                   |                      |
+|   - Authorized Read           | x<sup>1,2</sup>   |           x          |
```
erikbosch, line 22: Would it be an idea to duplicate the "Authorized Read" and "Authorized Update" lines so that we have lines for:

- Authorized Read of current value (supported by both)
- Authorized Read of target value (only supported by server)
- Authorized Update of current value (only supported by server)
- Authorized Update of target value (supported by both)

To make it more clear what we have in server WS protocol

doc/protocol/support.md
argerus: Yes, that is confusing.

When I added it, I assumed that the "not VISSv2" that kuksa-val-server and kuksa-client speaks was VISSv1. So if any server were to return VISSv1, kuksa-client could just use the that.

But reading the documentation here, it seems that it is actually a modified version of VISSv2 they use. Which is also confusing, because I don't think a VISSv2 would work with kuksa-val-server.

kuksa-client/kuksa_client/cli_backend/ws.py
```diff
@@ -153,18 +160,33 @@ def updateMetaData(self, path, jsonStr, timeout=5):
...
 
     # Set value to a given path
     def setValue(self, path, value, attribute="value", timeout=5):
         if 'nan' == value:
             return json.dumps({"error": path + " has an invalid value " + str(value)}, indent=2)
         req = {}
-        req["action"] = "set"
-        req["path"] = path
-        req["attribute"] = attribute
+        if self.subprotocol == "VISSv2":
```
erikbosch, line 180: Standard VISSv2 only supports setting target value and reading current value, so I suggest we give an error if someone use "setValue" and instead support "setTargetValue". Similar give an error if someone call "getTargetValue"

argerus: Pushed an update to this effect.

doc/protocol/support.md
argerus: Pushed an update which removed `"VISSv1"` from the requested subprotocols.

argerus: > Would it be better to call them (in kuksa-client) something like "VISSv2-std" and "VISSv2-extended" or something similar.

The VISSv2 [specification](https://www.w3.org/TR/viss2-transport/#wss-initialisation) explicitly states that it SHALL be "VISSv2", so no.
> The sub-protocol name SHALL be 'VISSv2' with the digit 2 being the version number.

I guess support for negotiating subprotocol can be added to kuksa-val-server, and that it then would be possible for kuksa-client to identify a server "speaking kuksa-val-server protocol". But I don't know the effort required to do that.

erikbosch (approved):

PR #643, opened 2023-08-31
Use shared SPDX action

SebastianSchildt:
This iuses the shared SPDX action in the new Orga. Should not change anyhting functionality wise.

I am however not sure if deleting the action itself has neagative side effects in case it is used elsewhere. If other repos refer to a tag, no issue, if they refer to no reference or master it needs further PRs in the other Repos. Any idea @erikbosch ?


erikbosch:
Don't you need to change https://github.com/eclipse/kuksa.val/blob/master/.github/workflows/check_license.yaml so that it uses the action from the new location (i.e. https://github.com/eclipse-kuksa/kuksa-actions)

Yes, it is true that other workflows in other repos might need to be changed (like https://github.com/eclipse/kuksa.val.feeders/blob/main/.github/workflows/check_license.yaml) but that is wanted. Using latest master/main is always risky :-)

SebastianSchildt:
> Don't you need to change https://github.com/eclipse/kuksa.val/blob/master/.github/workflows/check_license.yaml so that it uses the action from the new location (i.e. https://github.com/eclipse-kuksa/kuksa-actions)
> 

I need, I did, I didn't push, hopefully fixed now

.github/workflows/check_license.yaml
```diff
@@ -21,8 +21,8 @@ jobs:
     - name: Set license
       run: |
         echo "licenses=Apache-2.0" >> $GITHUB_ENV
-        
-    - uses: ./.github/actions/spdx
+
```
erikbosch, line 24: You need to remove trailing whitespaces on line 10 as well

erikbosch (approved):

PR #644, opened 2023-09-04
[databroker] Add top level read interface.

argerus:
This is a first step towards a better internal API.

Adding "top level" read guard will enable users to reuse an
already aquired quard for subsequent interaction.

The same functionality can be accomplished by adding more and more
wrapper functions whose main purpose is to wrap the underlying lock
(for example adding wrapper functions that would take a closure and
apply it in a certain way on an aquired lock guard).

Adding access using a top level read guard seems better.

Future steps could be to gradually removing most of the existing top
level wrapper functions.


argerus:
Closing this for now, as there might be advantages in using wrappers.

PR #645, opened 2023-09-04
Add iterator access for entries

argerus:
A wrapper function that made it easy to iterate through metadata was already available in the broker interface:
```rust
broker
    .map_metadata(|metadata| proto::Metadata::from(metadata))
    .await
```

This has been generalized to iterate over entries instead (i.e. not only metadata) as that will be useful when in the `kuksa.val.v1` interface for example:
```rust
broker
    .map_entries(|entry| {
        // Extract stuff from all entries (including metadata)
    }).await

broker
    .filter_map_entries(|entry| {
        // Extract stuff from some entries (including metadata)
    }).await
```

### DIFF
--- a/kuksa_databroker/databroker/src/broker.rs
+++ b/kuksa_databroker/databroker/src/broker.rs
@@ -845,16 +845,16 @@ pub struct DatabaseWriteAccess<'a, 'b> {
     permissions: &'b Permissions,
 }
 
-pub struct MetadataIterator<'a> {
-    inner: std::collections::hash_map::Iter<'a, i32, Entry>,
+pub struct EntryIterator<'a> {
+    inner: std::collections::hash_map::Values<'a, i32, Entry>,
 }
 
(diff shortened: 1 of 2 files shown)


lukasmittag (commented):
LGTM just one question. Will ```filter_map_entries``` get used cause currently I see not how it gets used. And just to confirm it is only used in grpc/sdv_databroker_v1 because in the kuksa_val_v1 the feature is not implemented yet am I right?

argerus:
> LGTM just one question. Will `filter_map_entries` get used cause currently I see not how it gets used. And just to confirm the adaptions get only used in `grpc/sdv_databroker_v1` because in the `grpc/kuksa_val_v1` the feature is not implemented yet - am I right?

Correct (on both counts). The reason for adding it (`filter_map_entries` etc) is that it would have made implementing #641 and similar functionality easier to implement. That PR is already merged, and added very interface specific code to `broker.rs`. That code can be moved out to live along the rest of the interface specific code when this PR is merged.

PR #646, opened 2023-09-04
Support setting value "NotAvailable"

argerus:
Add the ability to set a value as being not available (i.e. the initial state of a signal that is not set).


SebastianSchildt (approved):
Looks ok to me,

I do think there is the Corner case that a string signal can not be set to literal "NotAvailable" anymore, but as this is just a choice made by the client and not an API issue as such I think it is fine (alternative would be an "unset" call)

PR #648, opened 2023-09-04
[kuksa-client] Show subscription events in client.

argerus:
- Show subscription events in client instead of writing them to a file (and suggesting the user tails them manually).

### DIFF
--- a/kuksa-client/kuksa_client/__main__.py
+++ b/kuksa-client/kuksa_client/__main__.py
@@ -26,7 +26,6 @@ import os
 import pathlib
 import sys
 import threading
-import time
 
 from pygments import highlight
 from pygments import lexers
@@ -128,9 +127,9 @@ class TestClient(Cmd):
 
         return basic_complete(text, line, begidx, endidx, self.pathCompletionItems)
 
-    def subscribeCallback(self, logPath, resp):
-        with logPath.open('a', encoding='utf-8') as logFile:
-            logFile.write(resp + "\n")
+    def subscribeCallback(self, resp):
+        self.async_alert(highlight(json.dumps(json.loads(resp), indent=2), lexers.JsonLexer(),
+                  formatters.TerminalFormatter()))
 
     def subscriptionIdCompleter(self, text, line, begidx, endidx):
         self.pathCompletionItems = []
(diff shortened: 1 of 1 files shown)


kuksa-client/kuksa_client/__main__.py
```diff
@@ -401,18 +406,11 @@ def do_subscribe(self, args):
...
             resp = self.commThread.subscribe(
-                args.Path, callback, args.attribute)
+                args.Path, self.subscribeCallback, args.attribute)
             resJson = json.loads(resp)
             if "subscriptionId" in resJson:
                 self.subscribeIds.add(resJson["subscriptionId"])
-                logPath.touch()
-                print(f"Subscription log available at {logPath}")
-                print(
-                    f"Execute tail -f {logPath} on another Terminal instance")
```
erikbosch, line 415: If we change in this method, should we not also change `do_subscribeMultiple`

argerus: Yes, wasn't aware of that one.

erikbosch:
Looks good to me.

What we possibly could discuss, not necessarily part of this PR, is if we would like to align output format for `subscribe` and `getValue`. Today for subscribe you always get a list including "entry" and "fields", but you do not get that for  `getValue` which strictly may not be needed as you know from the command if it is "value" or something else. On the other hand, if we assume that output not will be parsed it does not matter. Also, we do not document how output will look like for subscribe, also a possible future improvement. But nothing that needs to be done for this PR I think.

erikbosch (approved):

SebastianSchildt:
The tail CAN be annoying sometimes, but how does this Pr influence the ability to "keep working" while several subscriptions running? Maybe this should be more like an option?

argerus:
> The tail CAN be annoying sometimes, but how does this Pr influence the ability to "keep working" while several subscriptions running? Maybe this should be more like an option?

Good point. I just got annoyed by this indirection, but I guess it can be feature in some cases.

I added an option to enable old behaviour.

SebastianSchildt (approved):
Works. Nice little quality-of-life-improvement for some use cases without breaking others 🥳

PR #651, opened 2023-09-07
Add cancel-in-progress

### LINKED ISSUE #626: CI cancel in progress
Currently we spend time building PR versions that already are obsolete as a new commit has been added. That is a problem as our build resources are limited. We should investigate if we should add `cancel-in-progress` so that we do not need to wait for it to finish. We already have it in some places:

https://github.com/eclipse/kuksa.val.services/blob/main/.github/workflows/lint.yml#L25

```
concurrency:
  group: ${{ github.ref }}-${{ github.workflow }}
  cancel-in-progress: true
```

 See discussion at https://stackoverflow.com/questions/66335225/how-to-cancel-previous-runs-in-the-pr-when-you-push-new-commitsupdate-the-curre

erikbosch:
Intention is to cancel ongoing/queued jobs if a new version of a PR is uploaded

Fixes #626

### DIFF
--- a/.github/workflows/check_license.yml
+++ b/.github/workflows/check_license.yml
@@ -3,9 +3,14 @@ name: spdx
 on:
   pull_request
 
+concurrency:
+      group: ${{ github.ref }}-${{ github.workflow }}
+      cancel-in-progress: true
+
 jobs:
   check-spdx-headers:
     runs-on: ubuntu-latest
+
     steps:
 
     - name: Checkout code
--- a/.github/workflows/check_push_rights.yml
+++ b/.github/workflows/check_push_rights.yml
@@ -5,10 +5,14 @@ on:
         description: "In possession of ghcr.io tokens?"
         value: ${{ jobs.check_push_rights.outputs.have_secrets }}
 
+# No concurrency group or cancel-in-progress here as this workflow is called from other workflows
+# so if you have a concurrency definition here it will be considered as a deadlock and job dismissed
+# The job below is also very fast, so no real need for cancel-in-progress
 
 jobs:
   check_push_rights:
     runs-on: ubuntu-latest
+
     outputs:
       have_secrets: ${{ steps.check-secrets.outputs.have_secrets }}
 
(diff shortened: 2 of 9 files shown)


.github/workflows/check_license.yml
```diff
@@ -6,6 +6,11 @@ on:
 jobs:
   check-spdx-headers:
     runs-on: ubuntu-latest
+
+    concurrency:
+      group: ${{ github.ref }}-${{ github.workflow }}-check-spdx-headers
```
erikbosch, line 11: General idea is to have a unique group-id for each job. I first tried including `${{ github.job }}` but that is not set when this is evaluated, so now the pattern is ref-workflow-<job title>

erikbosch:
This is an example of what happens if a job is cancelled because a new PR version has been uploaded

image

lukasmittag (commented):
couldn't we put it one layer up on the same level as jobs? because if one workflow will have one with a higher priority it will cancel all and we need less code. Should work too according to the docs.

erikbosch:
> couldn't we put it one layer up on the same level as jobs? because if one workflow will have one with a higher priority it will cancel all and we need less code. Should work too according to the docs.

I tried to put the concurrent on workflow level like below, but then the cancel-in-progress seemed to cancel all but one of the jobs in that workflow. Will test a bit more

```
 concurrency:
      group: ${{ github.ref }}-${{ github.workflow }}
      cancel-in-progress: true
```

erikbosch:
> couldn't we put it one layer up on the same level as jobs? because if one workflow will have one with a higher priority it will cancel all and we need less code. Should work too according to the docs.

I think I found the problem of having concurrency defined on workflow level - we can have it for all but the `check_push_right` workflows, as that is a reusable workflow. Also no need on that one as it is a very quick one, the biggest gain is in our "lengthy" workflows, especially the ARM ones

SebastianSchildt (approved):
Looks good to me, and everything is green

If an actual (sequence of) cancel(s) later leads to some unforeseen problems, we will see and can fix it then :)

PR #654, opened 2023-09-13
Use entries iterator that handles permissions correctly

rafaeling:
**New expected behaviour**

1. Client remains as it was, just pre-commit style was updated.
2. Databroker responses works as follow:
      - ```getValue(path)```: return value or error if there is any.
      - ```getValues(path1, path2, ... , n)``` return all path values if there are no errors, if there is/are any error/errors it would return no **partial values** and just its corresponding error for each error path.

      - ```getValue(wildcard)```: return value or error if there is any but **no partial values**
      - ```getValues(wildcard1, wildcard2, ... , n)``` return all wildcard values if there are no errors, if there is/are any error/errors it would return no **partial values of any wildcard** and just its corresponding error for each error wildcard.
      
      - ```getValues(wildcard1, path2, ... , n)``` return all path/wildcard values if there are no errors, if there is/are any error/errors it would return no **partial values** and just its corresponding error for each error path/wildcard.
      
      - ```getMetaData(wildcard or path)``` return values without permission errors.

**(Outdated)**
**How wildards and request errors are handled?**

- When a wildcard pattern is requested and there are permission errors for some items, the response will include entries for which permission was granted, along with just one permission error indicating that "access was denied for some entries" -> Lazy developers should find out which those entries are
- If a wildcard pattern is requested, but it doesn't match any items, the response will consist of a single error, "No entries found for the provided path."
- When a request is made for an entry that doesn't exist, the response will also contain a single error, "No entries found for the provided path."
- In cases where a request is made for a value for which the requester lacks permission, the response will include one error indicating "Permission Denied" for the access attempt.

### DIFF
--- a/kuksa-client/kuksa_client/grpc/__init__.py
+++ b/kuksa-client/kuksa_client/grpc/__init__.py
@@ -315,7 +315,6 @@ class Datapoint:
             ) if message.HasField('timestamp') else None,
         )
 
-
     def cast_array_values(cast, array):
         """
         Parses array input and cast individual values to wanted type.
@@ -339,7 +338,7 @@ class Datapoint:
             # My Way
             # ... without quotes
             if item.strip() == '':
-                #skip
+                # skip
                 pass
             else:
                 yield cast(item)
@@ -365,7 +364,7 @@ class Datapoint:
         new_val = new_val.replace('\\\"', '\"')
         new_val = new_val.replace("\\\'", "\'")
         return new_val
-            
+
     def to_message(self, value_type: DataType) -> types_pb2.Datapoint:
         message = types_pb2.Datapoint()
 
@@ -373,7 +372,6 @@ class Datapoint:
             array = getattr(obj, attr)
             array.Clear()
             array.values.extend(values)
-
 
         field, set_field, cast_field = {
             DataType.INT8: ('int32', setattr, int),
@@ -388,29 +386,29 @@ class Datapoint:
             DataType.DOUBLE: ('double', setattr, float),
             DataType.BOOLEAN: ('bool', setattr, Datapoint.cast_bool),
             DataType.STRING: ('string', setattr, Datapoint.cast_str),
-            DataType.INT8_ARRAY: ('int32_array', set_array_attr, 
+            DataType.INT8_ARRAY: ('int32_array', set_array_attr,
                                   lambda array: Datapoint.cast_array_values(int, array)),
-            DataType.INT16_ARRAY: ('int32_array', set_array_attr, 
+            DataType.INT16_ARRAY: ('int32_array', set_array_attr,
                                    lambda array: Datapoint.cast_array_values(int, array)),
-            DataType.INT32_ARRAY: ('int32_array', set_array_attr, 
+            DataType.INT32_ARRAY: ('int32_array', set_array_attr,
                                    lambda array: Datapoint.cast_array_values(int, array)),
-            DataType.UINT8_ARRAY: ('uint32_array', set_array_attr, 
+            DataType.UINT8_ARRAY: ('uint32_array', set_array_attr,
                                    lambda array: Datapoint.cast_array_values(int, array)),
-            DataType.UINT16_ARRAY: ('uint32_array', set_array_attr, 
+            DataType.UINT16_ARRAY: ('uint32_array', set_array_attr,
                                     lambda array: Datapoint.cast_array_values(int, array)),
-            DataType.UINT32_ARRAY: ('uint32_array', set_array_attr, 
+            DataType.UINT32_ARRAY: ('uint32_array', set_array_attr,
                                     lambda array: Datapoint.cast_array_values(int, array)),
-            DataType.UINT64_ARRAY: ('uint64_array', set_array_attr, 
+            DataType.UINT64_ARRAY: ('uint64_array', set_array_attr,
                                     lambda array: Datapoint.cast_array_values(int, array)),
-            DataType.INT64_ARRAY: ('int64_array', set_array_attr, 
+            DataType.INT64_ARRAY: ('int64_array', set_array_attr,
                                    lambda array: Datapoint.cast_array_values(int, array)),
-            DataType.FLOAT_ARRAY: ('float_array', set_array_attr, 
+            DataType.FLOAT_ARRAY: ('float_array', set_array_attr,
                                    lambda array: Datapoint.cast_array_values(float, array)),
-            DataType.DOUBLE_ARRAY: ('double_array', set_array_attr, 
+            DataType.DOUBLE_ARRAY: ('double_array', set_array_attr,
                                     lambda array: Datapoint.cast_array_values(float, array)),
-            DataType.BOOLEAN_ARRAY: ('bool_array', set_array_attr, 
+            DataType.BOOLEAN_ARRAY: ('bool_array', set_array_attr,
                                      lambda array: Datapoint.cast_array_values(Datapoint.cast_bool, array)),
-            DataType.STRING_ARRAY: ('string_array', set_array_attr, 
+            DataType.STRING_ARRAY: ('string_array', set_array_attr,
                                     lambda array: Datapoint.cast_array_values(Datapoint.cast_str, array)),
         }.get(value_type, (None, None, None))
         if self.value is not None:
@@ -523,6 +521,7 @@ class ServerInfo:
     def from_message(cls, message: val_pb2.GetServerInfoResponse):
         return cls(name=message.name, version=message.version)
 
+
 class BaseVSSClient:
     def __init__(
         self,
@@ -536,7 +535,6 @@ class BaseVSSClient:
         connected: bool = False,
         tls_server_name: Optional[str] = None
     ):
-    
 
         self.authorization_header = self.get_authorization_header(token)
         self.target_host = f'{host}:{port}'
@@ -559,11 +557,10 @@ class BaseVSSClient:
                 logger.info("Using client private key and certificates, mutual TLS supported if supported by server")
                 return grpc.ssl_channel_credentials(root_certificates, private_key, certificate_chain)
             else:
-                logger.info(f"No client certificates provided, mutual TLS not supported!")
+                logger.info("No client certificates provided, mutual TLS not supported!")
                 return grpc.ssl_channel_credentials(root_certificates)
-        logger.info(f"No Root CA present, it will not be posible to use a secure connection!")
+        logger.info("No Root CA present, it will not be posible to use a secure connection!")
         return None
-        
 
     def _prepare_get_request(self, entries: Iterable[EntryRequest]) -> val_pb2.GetRequest:
         req = val_pb2.GetRequest(entries=[])
@@ -649,7 +646,7 @@ class BaseVSSClient:
         return "Bearer " + token
 
     def generate_metadata_header(self, metadata: list, header=None) -> list:
-        if header == None:
+        if header is None:
             header = self.authorization_header
         if metadata:
             metadata = dict(metadata)
@@ -686,7 +683,7 @@ class VSSClient(BaseVSSClient):
         creds = self._load_creds()
         if target_host is None:
             target_host = self.target_host
-            
+
         if creds is not None:
             logger.info("Establishing secure channel")
             if self.tls_server_name:
@@ -694,12 +691,12 @@ class VSSClient(BaseVSSClient):
                 options = [('grpc.ssl_target_name_override', self.tls_server_name)]
                 channel = grpc.secure_channel(target_host, creds, options)
             else:
-                logger.debug(f"Not providing explicit TLS server name")
+                logger.debug("Not providing explicit TLS server name")
                 channel = grpc.secure_channel(target_host, creds)
         else:
             logger.info("Establishing insecure channel")
             channel = grpc.insecure_channel(target_host)
-            
+
         self.channel = self.exit_stack.enter_context(channel)
         self.client_stub = val_pb2_grpc.VALStub(self.channel)
         self.connected = True
@@ -973,7 +970,6 @@ class VSSClient(BaseVSSClient):
             else:
                 raise VSSClientError.from_grpc_error(exc) from exc
         return None
-        
 
     def get_value_types(self, paths: Collection[str], **rpc_kwargs) -> Dict[str, DataType]:
         """
--- a/kuksa-client/kuksa_client/cli_backend/grpc.py
+++ b/kuksa-client/kuksa_client/cli_backend/grpc.py
@@ -28,7 +28,6 @@ from typing import Iterable
 from typing import Optional
 import uuid
 import os
-import re
 
 from kuksa_client import cli_backend
 import kuksa_client.grpc
@@ -36,14 +35,17 @@ import kuksa_client.grpc.aio
 from kuksa_client.grpc import EntryUpdate
 from kuksa.val.v1 import types_pb2
 
+
 def callback_wrapper(callback: Callable[[str], None]) -> Callable[[Iterable[EntryUpdate]], None]:
     def wrapper(updates: Iterable[EntryUpdate]) -> None:
         callback(json.dumps([update.to_dict() for update in updates]))
     return wrapper
 
+
 class DatabrokerEncoder(json.JSONEncoder):
     def default(self, obj):
-        if isinstance(obj, (types_pb2.StringArray, types_pb2.BoolArray, types_pb2.Uint32Array, types_pb2.Uint64Array, types_pb2.FloatArray, types_pb2.Int32Array, types_pb2.Int64Array, types_pb2.DoubleArray)):
+        if isinstance(obj, (types_pb2.StringArray, types_pb2.BoolArray, types_pb2.Uint32Array, types_pb2.Uint64Array,
+                            types_pb2.FloatArray, types_pb2.Int32Array, types_pb2.Int64Array, types_pb2.DoubleArray)):
             string_values = []
             for value in obj.values:
                 value = str(value)
@@ -145,7 +147,7 @@ class Backend(cli_backend.Backend):
         return json.dumps({"error": "Invalid Attribute"})
 
     # Function for authorization
-    def authorize(self, token_or_tokenfile:Optional[str] =None, timeout=5):
+    def authorize(self, token_or_tokenfile: Optional[str] = None, timeout=5):
         if token_or_tokenfile is None:
             token_or_tokenfile = self.token_or_tokenfile
         if os.path.isfile(token_or_tokenfile):
@@ -251,7 +253,7 @@ class Backend(cli_backend.Backend):
                 responseQueue.put((resp, None))
             except kuksa_client.grpc.VSSClientError as exc:
                 responseQueue.put((None, exc.to_dict()))
-            except ValueError as exc:
+            except ValueError:
                 responseQueue.put(
                     (None, {"error": "ValueError in casting the value."}))
 
--- a/kuksa-client/kuksa_client/grpc/aio.py
+++ b/kuksa-client/kuksa_client/grpc/aio.py
@@ -75,7 +75,7 @@ class VSSClient(BaseVSSClient):
                 options = [('grpc.ssl_target_name_override', self.tls_server_name)]
                 channel = grpc.aio.secure_channel(target_host, creds, options)
             else:
-                logger.debug(f"Not providing explicit TLS server name")
+                logger.debug("Not providing explicit TLS server name")
                 channel = grpc.aio.secure_channel(target_host, creds)
         else:
             logger.info("Establishing insecure channel")
@@ -363,7 +363,7 @@ class VSSClient(BaseVSSClient):
         except AioRpcError as exc:
             if exc.code() == grpc.StatusCode.UNAUTHENTICATED:
                 logger.info("Unauthenticated channel started")
-            else: 
+            else:
                 raise VSSClientError.from_grpc_error(exc) from exc
         return None
 
--- a/kuksa-client/tests/test_grpc.py
+++ b/kuksa-client/tests/test_grpc.py
@@ -435,7 +435,9 @@ class TestVSSClient:
         client = VSSClient('127.0.0.1', unused_tcp_port)
         mocker.patch.object(client, 'set')
         await client.set_current_values({
-            'Vehicle.Speed': Datapoint(42.0, datetime.datetime(2022, 11, 7, 16, 18, 35, 247307, tzinfo=datetime.timezone.utc)),
+            'Vehicle.Speed': Datapoint(42.0,
+                                       datetime.datetime(2022, 11, 7, 16, 18, 35, 247307,
+                                                         tzinfo=datetime.timezone.utc)),
             'Vehicle.ADAS.ABS.IsActive': Datapoint(True),
             'Vehicle.Chassis.Height': Datapoint(666),
         })
@@ -518,7 +520,9 @@ class TestVSSClient:
                            View.CURRENT_VALUE, (Field.VALUE,)),
         ]
         assert received_updates == {
-            'Vehicle.Speed': Datapoint(42.0, datetime.datetime(2022, 11, 7, 16, 18, 35, 247307, tzinfo=datetime.timezone.utc)),
+            'Vehicle.Speed': Datapoint(42.0,
+                                       datetime.datetime(2022, 11, 7, 16, 18, 35, 247307,
+                                                         tzinfo=datetime.timezone.utc)),
             'Vehicle.ADAS.ABS.IsActive': Datapoint(True),
             'Vehicle.Chassis.Height': Datapoint(666),
         }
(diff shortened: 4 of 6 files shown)


rafaeling:
A separate PR should be created to fix the actuator filter on kuksa-client for getTargetValue method

argerus (commented):
Code wise, I don't have anything of substance to add.

I haven't tested the actual behaviour that much though. I did try testing this though:
1. Start databroker with authorization enabled.
2. Start kuksa-client using jwt token `jwt/read-vehicle-speed.token`.
3. Run: `getValues Vehicle.Speed`
4. Run: `getValues Vehicle.Speed Vehicle.Service.* Vehicle.Asd`

The first command correctly returns:
```
{
    "path": "Vehicle.Speed",
    "value": {
        "value": 10.0,
        "timestamp": "2023-09-20T08:51:12.047766+00:00"
    }
}
```

But the second command returns:
```
{
    "error": {
        "code": 403,
        "reason": "forbidden",
        "message": "Permission denied for some entries"
    },
    "errors": [
        {
            "path": "Vehicle.Service.*",
            "error": {
                "code": 403,
                "reason": "forbidden",
                "message": "Permission denied for some entries"
            }
        },
        {
            "path": "Vehicle.Asd",
            "error": {
                "code": 404,
                "reason": "not_found",
                "message": "No entries found for the provided path"
            }
        }
    ]
}
```

which I guess is wrong, as i doesn't include what was in the first response.

Not sure if this is a bug in the client or this implementation (or my understanding of how it _should_ work).

lukasmittag (approved):
Works as expected. Tried:
```
Test Client> authorize ../jwt/provide-vehicle-speed.token 
"Authenticated"

Test Client> getValues Vehicle.Speed Vehicle.PowerOptimizeLevel 
[
    {
        "path": "Vehicle.Speed"
    },
    {
        "path": "Vehicle.PowerOptimizeLevel",
        "error": {
            "code": 403,
            "reason": "forbidden",
            "message": "Permission denied"
        }
    }
]

Test Client> setValue Vehicle.Speed 12
OK

Test Client> getValues Vehicle.Speed Vehicle.PowerOptimizeLevel 
[
    {
        "path": "Vehicle.Speed",
        "value": {
            "value": 12.0,
            "timestamp": "2023-09-21T07:23:20.029219+00:00"
        }
    },
    {
        "path": "Vehicle.PowerOptimizeLevel",
        "error": {
            "code": 403,
            "reason": "forbidden",
            "message": "Permission denied"
        }
    }
]

Test Client> getValue Vehicle.Speed
[
    {
        "path": "Vehicle.Speed",
        "value": {
            "value": 12.0,
            "timestamp": "2023-09-21T07:23:20.029219+00:00"
        }
    }
]

Test Client> getValue Vehicle.Power
PowerOptimizeLevel    Powertrain.           
Test Client> getValue Vehicle.PowerOptimizeLevel 
{
    "error": {
        "code": 403,
        "reason": "forbidden",
        "message": "Permission denied"
    },
    "errors": [
        {
            "path": "Vehicle.PowerOptimizeLevel",
            "error": {
                "code": 403,
                "reason": "forbidden",
                "message": "Permission denied"
            }
        }
    ]
}

Test Client> getMetaData Vehicle.PowerOptimizeLevel 
{
    "error": {
        "code": 403,
        "reason": "forbidden",
        "message": "Permission denied"
    },
    "errors": [
        {
            "path": "Vehicle.PowerOptimizeLevel",
            "error": {
                "code": 403,
                "reason": "forbidden",
                "message": "Permission denied"
            }
        }
    ]
}

Test Client> getMetaData Vehicle.Speed 
[
    {
        "path": "Vehicle.Speed",
        "metadata": {
            "data_type": "FLOAT",
            "entry_type": "SENSOR",
            "description": "Vehicle speed."
        }
    }
]

```
Just one remark: Have we agreed upon that metadata needs permission now too?

SebastianSchildt:
My understanding was, last discussion went to having metaData free for all for now, or has that changed? Espiecially wouldnt this break "Learning" with "*" what data is available?

rafaeling:
> My understanding was, last discussion went to having metaData free for all for now, or has that changed? Espiecially wouldnt this break "Learning" with "*" what data is available?
It is a bug, getMetaData should work anyways, have to fix it, thanks Lukas :)

kuksa-client/kuksa_client/grpc/__init__.py
```diff
@@ -576,10 +573,14 @@ def _prepare_get_request(self, entries: Iterable[EntryRequest]) -> val_pb2.GetRe
         logger.debug("%s: %s", type(req).__name__, req)
         return req
 
-    def _process_get_response(self, response: val_pb2.GetResponse) -> List[DataEntry]:
+    def _process_get_response(self, response: val_pb2.GetResponse) -> (List[DataEntry], List[str]):
```
argerus, line 576: I don't think this change is a good idea.

It breaks every client using the `kuksa.grpc` library.
It creates a very weird interface. Previously, when doing:
```python
client.get_current_values(["Vehicle.Speed", "Vehicle.Services"])
```

you would either get:
* a list of entries
* an exception (for example permission denied)

with this change, you would either get:
* a list of entries
* a list of entries and a list of errors
* an exception (for example permissions denied)

The previous behaviour was a lot more pythonic, and easier to work with.

I guess the main culprit, is that the the grpc function should either be successful and return the expected result **or** fail and return an error. Otherwise, I think the end result will be exceptionally confusing.

argerus (changes requested):

kuksa-client/kuksa_client/grpc/__init__.py
SebastianSchildt: I would agree that "half-errored" is maybe unexpected. If we don't want that, I think the main question is, when there is a problem with "some" entries, do we want to
 - just return an error
 - silenty ignore the problematic cases and continue with the working ones

Next question, are these two questions
 - Using a wildcard
 - Listing all entries a wildcard would match by hand
 equivalent in that regard, or should we treat one differently from the other?

Opinions?

rafaeling: The only way I can think of for now, assuming we don't want to break anything by modifying ```kuksa.grpc``` would be John's suggestion today during the daily.

We disable erros for wildcards. That leads to the following:

- If the request is wildcard the response contains only allowed entries (silenty ignore the problematic cases)
- If the request is for a signal path without permission an error is returned.
- If the request is a list of entries (with and without permission), only the error permission is returned. The client should be aware that it cannot query that entry and should modify the request.

argerus: Yes, maybe.

I'm having trouble foreseeing all the implications. The "just return an error" would probably be the best in terms of having a predictable behaviour.

But something a long the line of what @rafaeling listed above might be more pragmatic in some sense, but would require implementing special handling which would probably make it harder to understand and predict (as a user).

I can easily imagine how "silently ignoring errors" for wildcards might lead to unintended behaviour in clients.

The fundamental issue, I believe, is that what can be requested is very unrestricted. You can request:
* paths together with wildcard paths.
* different fields for every path or wildcard path.

And in combination with a global error + a global (optional) list of per path errors (next to the successful results), it is not at all clear how this _should_ behave.

rafaeling: I also see how silently ignoring errors can greatly confuse clients. Another option that could work and could be to return just an error for each wildcard (would be treated as signal path), while also including a permission field for each signal in the metadata. This way, the clients are fully aware upfront of what they have access to or not.

The client should start by calling ```getMetaData``` to retrieve the entire tree with permissions included and then build the requests from there. Additionally, this feature could allow our kuksa-client to, for instance, add colors to differentiate between entries with permissions and those without when using autocompletion.

argerus: > I also see how silently ignoring errors can greatly confuse clients. Another option that could work could be to return just an error for each wildcard (would be treated as signal path), while also including a permission field for each signal in the metadata.

So, if I understand correctly, that would be "return only error", but there would be one error per "path" requested.

> while also including a permission field for each signal in the metadata. This way, the clients are fully aware upfront of what they have access to or not.

That's an interesting idea.

argerus: To follow up on my comment above:
> The fundamental issue, I believe, is that what can be requested is very unrestricted. You can request:
> * paths together with wildcard paths.
> * different fields for every path or wildcard path.

and dig into why this is problematic.

### Mixing "listing" of signals with "getting" signals.
As identified above, there is the question of whether wildcards should be handled differently than when requesting explicit paths. If the answer is yes, the ability to combine paths and wildcards in the same request is causing issues. One way to solve this is to have separate functions for these separate behaviours, e.g.
* `get(...)`  (get explicitly listed signals)
* `list(...)` (list available signals)

The new function `list(...)` would then list all signals that is available to be listed (it would be the "wildcard handling" function).
The `get(...)` function would return an error if permission is denied for an explicitly listed signal. Or (since it would be a 1:1 relationship between the number of paths requested and the number returned, it could also return a list with either a successful response or an error, i.e. `[(path, success_response), (path, success_response), (path, error), (path, success_response)]`)


### Mixing retrieval of metadata with retrieval of data
The other point is that both metadata and data can be requested for the same path. That leads to the question of what should happen if there is permission to read one of them but not the other.

One way to solve this would be to have different functions for these things.
* `get_values(...)` (and `list_values(...)`)
* `get_metadata(...)` (and/or `list_metadata(...)`)

`get_values(...)` would return an error if there is no permission to read the value for an explicitly listed path. Since it would be a 1:1 relationship between the number of paths requested and the number returned, it could also return a list with either a successful response or an error, i.e. `[(path, success_response), (path, success_response), (path, error), (path, success_response)]`)
`get_metadata(...)` or `list_metadata(...)` would not fail based on lacking permissions for the actual data.

SebastianSchildt: I think the metadata idea is nice, but let's decouple it form the discussion here.

So I created this token

```json
{
  "sub": "local dev",
  "iss": "createToken.py",
  "aud": [
    "kuksa.val"
  ],
  "iat": 1516239022,
  "exp": 1767225599,
  "scope": "read:Vehicle.Acceleration.Lateral read:Vehicle.Acceleration.Vertical"
}
```


and running this example:

```python
from kuksa_client.grpc import EntryRequest
from kuksa_client.grpc import Field
from kuksa_client.grpc import View
from kuksa_client.grpc import VSSClient

with VSSClient('127.0.0.1', 55556) as client:
    path = 'Vehicle.Body.Windshield.Front.Wiping.System.TargetPosition'
    client.authorize("eyJ0eXAiOiJKV1QiLCJhbGciOiJSUzI1NiJ9.eyJzdWIiOiJsb2NhbCBkZXYiLCJpc3MiOiJjcmVhdGVUb2tlbi5weSIsImF1ZCI6WyJrdWtzYS52YWwiXSwiaWF0IjoxNTE2MjM5MDIyLCJleHAiOjE3NjcyMjU1OTksInNjb3BlIjoicmVhZDpWZWhpY2xlLkFjY2VsZXJhdGlvbi5MYXRlcmFsIHJlYWQ6VmVoaWNsZS5BY2NlbGVyYXRpb24uVmVydGljYWwifQ.00PTCsHWocFEn5PjFqfvUwVo400WleuR4IK7SJu8MDCi0MZe19Pg3BFjCKefybQfwxgHpR_vuwxFtQPigvWTSVbQFMdn1zbOBulF9x_g7KGBjVcLNkw3MOQU5eswM-9e6rRyj5qWuL97tczOK21f3QEVsGacHWO9wYvPmw-Fb4PWOJUrx82ans7iuMPt3SCFjPa_of-NAPotgA14ahXHRzfLrlLFtHmgIeDthNGt8Fd-wPd5sl8r2vb7XQq1-eOb-v01nTwp2WFXwa3ZJV_EegNZzvM8Wf4nJl8zfRLNfmIqHlvpcaOnIVIHUBsfteJS4dICZhs6Ub3xscnEEEF9g37hIrQkLpXa96K4YcaxDmJr_MQH3qfhYX2Q0U_74KPxKzDv7fL5LYsLUB2RQ2114rxyD4jRVPheNwB56ytjLBDXTGuCZvVLpxouhCIn7cFfoSf24djy65BOhD7vFAE_ET1fIqvycNEl83o6QkKjKuLvJwC6omHlD1cTG99nOJBlNj3fJer4hw8iFy3xQjSE2eh0WVqeEmD7FjpfnlauovvSOndK5Qrk_Ih89OypvrgOqPE2n9EQuQ-UgHVB7ip2fU1Mo1GzpdxXgn7luRIx4K4fKESzhWrI90nkuIh2hb7pKhYWy8jIxrh2m2nSWld8Dr9Y9MCoYxd0uWoA7gxLASQ")
    entries = client.get_current_values([
        'Vehicle.Acceleration.Vertical', 'Vehicle.Acceleration.Longitudinal', 'Vehicle.Acceleration.Lateral'
    ])  

    print(str(entries))
```

This yields

```
 python test.py
  File "/Users/scs2rng/Documents/Dev/testkc/test.py", line 9, in <module>
    entries = client.get_current_values([
  File "/Users/scs2rng/Documents/Dev/kuksa.val/kuksa-client/kuksa_client/grpc/__init__.py", line 727, in get_current_values
    entries = self.get(
  File "/Users/scs2rng/Documents/Dev/kuksa.val/kuksa-client/kuksa_client/grpc/__init__.py", line 679, in wrapper
    return func(self, *args, **kwargs)
  File "/Users/scs2rng/Documents/Dev/kuksa.val/kuksa-client/kuksa_client/grpc/__init__.py", line 889, in get
    return self._process_get_response(resp)
  File "/Users/scs2rng/Documents/Dev/kuksa.val/kuksa-client/kuksa_client/grpc/__init__.py", line 581, in _process_get_response
    self._raise_if_invalid(response)
  File "/Users/scs2rng/Documents/Dev/kuksa.val/kuksa-client/kuksa_client/grpc/__init__.py", line 644, in _raise_if_invalid
    raise VSSClientError(error, errors)
kuksa_client.grpc.VSSClientError: ({'code': 403, 'reason': 'forbidden', 'message': 'Permission denied'}, [{'path': 'Vehicle.Acceleration.Longitudinal', 'error': {'code': 403, 'reason': 'forbidden', 'message': 'Permission denied'}}])
```

That was my baseline, and I think this is definitely one valid choice: My intention was "get values for three paths", the API telling me "Nope, that is not possible, here is one hint as to why"

When I modify this exmaple to 

```python
from kuksa_client.grpc import EntryRequest
from kuksa_client.grpc import Field
from kuksa_client.grpc import View
from kuksa_client.grpc import VSSClient

with VSSClient('127.0.0.1', 55556) as client:
    path = 'Vehicle.Body.Windshield.Front.Wiping.System.TargetPosition'
    client.authorize("eyJ0eXAiOiJKV1QiLCJhbGciOiJSUzI1NiJ9.eyJzdWIiOiJsb2NhbCBkZXYiLCJpc3MiOiJjcmVhdGVUb2tlbi5weSIsImF1ZCI6WyJrdWtzYS52YWwiXSwiaWF0IjoxNTE2MjM5MDIyLCJleHAiOjE3NjcyMjU1OTksInNjb3BlIjoicmVhZDpWZWhpY2xlLkFjY2VsZXJhdGlvbi5MYXRlcmFsIHJlYWQ6VmVoaWNsZS5BY2NlbGVyYXRpb24uVmVydGljYWwifQ.00PTCsHWocFEn5PjFqfvUwVo400WleuR4IK7SJu8MDCi0MZe19Pg3BFjCKefybQfwxgHpR_vuwxFtQPigvWTSVbQFMdn1zbOBulF9x_g7KGBjVcLNkw3MOQU5eswM-9e6rRyj5qWuL97tczOK21f3QEVsGacHWO9wYvPmw-Fb4PWOJUrx82ans7iuMPt3SCFjPa_of-NAPotgA14ahXHRzfLrlLFtHmgIeDthNGt8Fd-wPd5sl8r2vb7XQq1-eOb-v01nTwp2WFXwa3ZJV_EegNZzvM8Wf4nJl8zfRLNfmIqHlvpcaOnIVIHUBsfteJS4dICZhs6Ub3xscnEEEF9g37hIrQkLpXa96K4YcaxDmJr_MQH3qfhYX2Q0U_74KPxKzDv7fL5LYsLUB2RQ2114rxyD4jRVPheNwB56ytjLBDXTGuCZvVLpxouhCIn7cFfoSf24djy65BOhD7vFAE_ET1fIqvycNEl83o6QkKjKuLvJwC6omHlD1cTG99nOJBlNj3fJer4hw8iFy3xQjSE2eh0WVqeEmD7FjpfnlauovvSOndK5Qrk_Ih89OypvrgOqPE2n9EQuQ-UgHVB7ip2fU1Mo1GzpdxXgn7luRIx4K4fKESzhWrI90nkuIh2hb7pKhYWy8jIxrh2m2nSWld8Dr9Y9MCoYxd0uWoA7gxLASQ")

    entries = client.get_current_values([
        'Vehicle.Acceleration.*'
    ])  

    print(str(entries))
```
I get with current master

```
{'Vehicle.Acceleration.Longitudinal': None, 'Vehicle.Acceleration.Vertical': None, 'Vehicle.Acceleration.Lateral': None}
```

that is not good, have not tested this branch, but my _expectation_ would be, that this behaves exactly as if I had requested all matching paths "by hand".

That would seem consistent to me.

The suggested metadata extension is than one interesting way to get around this for "generic"/"exploratory" apps, but as I said that is an interesting suggestion anyway, apart from the current issue

argerus: > that is not good, have not tested this branch, but my expectation would be, that this behaves exactly as if I had requested all matching paths "by hand".

That particular response is caused by the "permissions bug" in the first implementation of wildcard support.

rafaeling:
PR description was updated with the new behaviour -> https://github.com/eclipse/kuksa.val/pull/654#issue-1893980884

argerus:
This is perhaps overly pedantic, but the server is still returning "partial success" and silently ignoring permission errors.

```python
from kuksa_client.grpc import VSSClient, EntryRequest, View, Field

read_vehicle_speed_token = "eyJ0eXAiOiJKV1QiLCJhbGciOiJSUzI1NiJ9.eyJzdWIiOiJsb2NhbCBkZXYiLCJpc3MiOiJjcmVhdGVUb2tlbi5weSIsImF1ZCI6WyJrdWtzYS52YWwiXSwiaWF0IjoxNTE2MjM5MDIyLCJleHAiOjE3NjcyMjU1OTksInNjb3BlIjoicmVhZDpWZWhpY2xlLlNwZWVkIn0.QyaKO-vjjzeimAAyUMjlM_jAvhlmYVguzXp76jAnW9UUiWowrXeYTa_ERVzZqxz21txq1QQmee8WtCO978w95irSRugTmlydWNjS6xPGAOCan6TEXBryrmR5FSmm6fPNrgLPhFEfcqFIQm07B286JTU98s-s2yeb6bJRpm7gOzhH5klCLxBPFYNncwdqqaT6aQD31xOcq4evooPOoV5KZFKI9WKkzOclDvto6TCLYIgnxu9Oey_IqyRAkDHybeFB7LR-1XexweRjWKleXGijpNweu6ATsbPakeVIfJ6k3IN-oKNvP6nawtBg7GeSEFYNksUUCVrIg8b0_tZi3c3E114MvdiDe7iWfRd5SDjO1f8nKiqDYy9P-hwejHntsaXLZbWSs_GNbWmi6J6pwgdM4XI3x4vx-0Otsb4zZ0tOmXjIL_tRsKA5dIlSBYEpIZq6jSXgQLN2_Ame0i--gGt4jTg1sYJHqE56BKc74HqkcvrQAZrZcoeqiKkwKcy6ofRpIQ57ghooZLkJ8BLWEV_zJgSffHBX140EEnMg1z8GAhrsbGvJ29TmXGmYyLrAhyTj_L6aNCZ1XEkbK0Yy5pco9sFGRm9nKTeFcNICogPuzbHg4xsGX_SZgUmla8Acw21AAXwuFY60_aFDUlCKZh93Z2SAruz1gfNIL43I0L8-wfw"

client = VSSClient(host="127.0.0.1", port=55555)
client.connect()
client.authorize(token=read_vehicle_speed_token)

# This will give no error / exception. I.e. permission denied is silently ignored and
# all metadata is returned + the value for the one thing we have permission for ("Vehicle.Speed").
resp1 = client.get([EntryRequest(path="Vehicle", view=View.UNSPECIFIED, fields=[Field.VALUE, Field.METADATA])])

# Causes an exception
# kuksa_client.grpc.VSSClientError: ({'code': 403, 'reason': 'forbidden', 'message': 'Permission denied for some entries'}, [{'path': 'Vehicle', 'error': {'code': 403, 'reason': 'forbidden', 'message': 'Permission denied for some entries'}}])
resp2 = client.get([EntryRequest(path="Vehicle", view=View.UNSPECIFIED, fields=[Field.VALUE])])

```

PS
Unrelated, but the python client just returns `None` if one fails to call connect before making a call.

```python
client = VSSClient(host="127.0.0.1", port=55555)
client.authorize(token=read_vehicle_speed_token)

# client.get(...) returns None
```

and fails to use the access token provided to it if `client.connect()` is called after `client.authorize(...)`

```python
client = VSSClient(host="127.0.0.1", port=55555)
client.authorize(token=read_vehicle_speed_token)
client.connect()

# client.get(...) fails with
# kuksa_client.grpc.VSSClientError: ({'code': 16, 'reason': 'unauthenticated', 'message': 'Invalid auth token'}, [])
```

SebastianSchildt:
I tested the behaviors lined out, and those work for me as I expect.

for Johns first comment: If several things are requested, i.e. in that case Metadata and Value, would it be possible @rafaeling  to also just deny the whole request, i.e. that always "deny" wins, i.e. if requesting value, target and metaData for a single data point, but you miss e.g. read rights, it will just deny the whole request? or is this hard with how this is currently architected?


The other comment, i.e. returning `none` on get when not connected, might be worth a separate issue. Not ideal, but let's not fix it here.

I think, that the token is not automatically used after "connect" is maybe ok, because as I see it, `connect()` resets the whole state client side, and I understand `authorize` more like  a function, and not an inherent property fo the client object. I _know_ this is not how it *is* in reality, but for waht you expect @argerus I would maybe expect the function not be called "authorize", but rather `set_authorization_token` or something like that, but same -> Sounds like material for a separate issue

rafaeling:
> This is perhaps overly pedantic, but the server is still returning "partial success" and silently ignoring permission errors.
> 
> ```python
> from kuksa_client.grpc import VSSClient, EntryRequest, View, Field
> 
> read_vehicle_speed_token = "eyJ0eXAiOiJKV1QiLCJhbGciOiJSUzI1NiJ9.eyJzdWIiOiJsb2NhbCBkZXYiLCJpc3MiOiJjcmVhdGVUb2tlbi5weSIsImF1ZCI6WyJrdWtzYS52YWwiXSwiaWF0IjoxNTE2MjM5MDIyLCJleHAiOjE3NjcyMjU1OTksInNjb3BlIjoicmVhZDpWZWhpY2xlLlNwZWVkIn0.QyaKO-vjjzeimAAyUMjlM_jAvhlmYVguzXp76jAnW9UUiWowrXeYTa_ERVzZqxz21txq1QQmee8WtCO978w95irSRugTmlydWNjS6xPGAOCan6TEXBryrmR5FSmm6fPNrgLPhFEfcqFIQm07B286JTU98s-s2yeb6bJRpm7gOzhH5klCLxBPFYNncwdqqaT6aQD31xOcq4evooPOoV5KZFKI9WKkzOclDvto6TCLYIgnxu9Oey_IqyRAkDHybeFB7LR-1XexweRjWKleXGijpNweu6ATsbPakeVIfJ6k3IN-oKNvP6nawtBg7GeSEFYNksUUCVrIg8b0_tZi3c3E114MvdiDe7iWfRd5SDjO1f8nKiqDYy9P-hwejHntsaXLZbWSs_GNbWmi6J6pwgdM4XI3x4vx-0Otsb4zZ0tOmXjIL_tRsKA5dIlSBYEpIZq6jSXgQLN2_Ame0i--gGt4jTg1sYJHqE56BKc74HqkcvrQAZrZcoeqiKkwKcy6ofRpIQ57ghooZLkJ8BLWEV_zJgSffHBX140EEnMg1z8GAhrsbGvJ29TmXGmYyLrAhyTj_L6aNCZ1XEkbK0Yy5pco9sFGRm9nKTeFcNICogPuzbHg4xsGX_SZgUmla8Acw21AAXwuFY60_aFDUlCKZh93Z2SAruz1gfNIL43I0L8-wfw"
> 
> client = VSSClient(host="127.0.0.1", port=55555)
> client.connect()
> client.authorize(token=read_vehicle_speed_token)
> 
> # This will give no error / exception. I.e. permission denied is silently ignored and
> # all metadata is returned + the value for the one thing we have permission for ("Vehicle.Speed").
> resp1 = client.get([EntryRequest(path="Vehicle", view=View.UNSPECIFIED, fields=[Field.VALUE, Field.METADATA])])
Good you found it but, what value should be expected here? An error or the metadata?

argerus:
> Sounds like material for a separate issue

Yes, definitely not in this PR.

> Good you found it but, what response should be expected here? An error or the metadata?

An error. Partial success are:
1. Not supported by kuksa-client.
2. Confusing in general.

PR #658, opened 2023-09-15
Better cmdline args specifying server in kuksa-client

### LINKED ISSUE #657: kuksa-client uses cumbersome way to specify server
The current kuksa-client has a very verbose, non-intuitive way to specify where to connect to, e.g.

```
kuksa-client --ip 127.0.0.1 --port 8090 --protocol ws 
```

The reason seems, that this corresponds to some internal dictionary used to set up the client. From a users perspective something like this

```
kuksa-client ws://127.0.01:8090
```

would seem much easier.

As the CLI is not (intended) to be used in any scripts, I think this can be changed without any kind of backwards compatibility

SebastianSchildt:
Will close #657 

This tries to simplify connecting to VSS servers using kuksa-client CLI.  It uses URIs now, exmaples of valid URIs

```
grpc://127.0.0.1:55555
grpcs://127.0.0.1:55555
ws://127.0.0.1:8090
wss://127.0.0.1:8090
```

Other changes

 - Changed default (when no command line parameters are given) to connecting to databroker instead of val server
 - Removed the "set/getServerAddress" commands within the CLI, instead use `connect <URI>
 - Updated Readme

When checking this, please also double check whether all variants (TLS/noTLS databroker, val-server) still work.

This is a change in the CLI part only, so the API itself is unaffected

### DIFF
--- a/kuksa-client/README.md
+++ b/kuksa-client/README.md
@@ -16,7 +16,7 @@ KUKSA Python Client and SDK provides both a command-line interface (CLI) and a s
 The fastest way to start using KUKSA Python Client is to install a pre-built version from pypi.org:
 
 ```console
-$ pip install kuksa-client
+pip install kuksa-client
 ```
 
 If you want to install from sources instead see [Building and running a local version](#building-and-running-a-local-version).
@@ -24,15 +24,15 @@ If you want to install from sources instead see [Building and running a local ve
 After you have installed the kuksa-client package via pip you can run the test client CLI directly by executing:
 
 ```console
-$ kuksa-client
+kuksa-client
 ```
 
-With default CLI arguments, the client will try to connect to a local VISS server e.g. `kuksa-val-server`.
-If you wish to connect to a gRPC server e.g. `kuksa-databroker`, you should instead run:
+With default CLI arguments, the client will try to connect to a local Databroker, e.g. a server supporting the `kuksa.val.v1` protocol without using TLS. This is equivalent to executing
 
 ```console
-$ kuksa-client --ip 127.0.0.1 --port 55555 --protocol grpc
+kuksa-client grpc://127.0.0.1:55555
 ```
+
 
 If everything works as expected and the server can be contacted you will get an output similar to below.
 
@@ -55,36 +55,42 @@ Welcome to Kuksa Client version <some_version>
 
 Default tokens directory: /some/path/kuksa_certificates/jwt
 
-connectj to wss://127.0.0.1:8090
-Websocket connected securely.
+Connecting to VSS server at 127.0.0.1 port 55555 using KUKSA GRPC protocol.
+TLS will not be used.
+INFO 2023-09-15 18:48:13,415 kuksa_client.grpc No Root CA present, it will not be posible to use a secure connection!
+INFO 2023-09-15 18:48:13,415 kuksa_client.grpc.aio Establishing insecure channel
+gRPC channel connected.
 Test Client>
 ```
 
-If the connected KUKSA Server or KUKSA Databroker require authorization the next step is to authorize.
-KUKSA Server and KUKSA Databroker use different token formats.
+If you wish to connect to a VISS server e.g. `kuksa-val-server` (not using TLS), you should instead run:
 
-### Connecting to KUKSA Databroker
+```console
+kuksa-client ws://127.0.0.1:8090
+```
 
-A gRPC connection to KUKSA Databroker is started by specifying address and port for the Databroker and giving
-`--protocol grpc` as argument.
-KUKSA Client use TLS by default, it only run in insecure mode if `--insecure` is given as argument.
-By default the KUKSA example Root CA and Client keys are used, but client keys have no effect as mutual authentication is not supported by KUKSA Databroker or KUKSA Server.
+### TLS with databroker
+
+KUKSA Client uses TLS to connect to databroker when the schema part of the server URI is `grpcs`, i.e. a valid command to connect to a TLS enabled local databroker is
 
 ```
-~/kuksa.val/kuksa-client$ kuksa-client --ip localhost --port 55555 --protocol grpc
+kuksa-client grpcs://localhost:55555
 ```
+
+By default the KUKSA example Root CA and Client keys are used, but client keys have no effect currently as mutual authentication is not supported by KUKSA Databroker or KUKSA Server.
+
 
 This call with all parameters specified give same effect:
 
 ```
-~/kuksa.val/kuksa-client$ kuksa-client --ip localhost --port 55555 --protocol grpc --certificate ../kuksa_certificates/Client.pem --keyfile ../kuksa_certificates/Client.key --cacertificate ./kuksa_certificates/CA.pem
+kuksa-client --certificate ../kuksa_certificates/Client.pem --keyfile ../kuksa_certificates/Client.key --cacertificate ./kuksa_certificates/CA.pem grpcs://localhost:55555
 ```
 
 There is actually no reason to specify client key and certificate, as mutual authentication is not supported in KUKSA Databroker,
 so the command can be simplified like this:
 
 ```
-~/kuksa.val/kuksa-client$ kuksa-client --ip localhost --port 55555 --protocol grpc --cacertificate ./kuksa_certificates/CA.pem
+kuksa-client  --cacertificate ./kuksa_certificates/CA.pem grpcs://localhost:55555
 ```
 
 The example server protocol list 127.0.0.1 as an alternative name, but the TLS-client currently used does not accept it,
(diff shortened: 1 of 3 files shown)


kuksa-client/README.md
```diff
@@ -27,11 +27,11 @@ After you have installed the kuksa-client package via pip you can run the test c
 $ kuksa-client
 ```
 
-With default CLI arguments, the client will try to connect to a local VISS server e.g. `kuksa-val-server`.
-If you wish to connect to a gRPC server e.g. `kuksa-databroker`, you should instead run:
+With default CLI arguments, the client will try to connect to a local databroker, e.g. a server supporting the `kuksa.val.v1` protocol..
```
erikbosch, line 30: `Databroker` with capital `D`?
Shall we also mention that it is "without TLS" ( I prefer not to use the term "insecure" as I see that as a bit ambiguous)

kuksa-client/README.md
```diff
@@ -27,11 +27,11 @@ After you have installed the kuksa-client package via pip you can run the test c
...
+If you wish to connect to a VISS server e.g. `kuksa-val-server`, you should instead run:
```
erikbosch, line 31: Move this part down to below the output log below (as the output log is for gRPC) to avoid misunderstandings

kuksa-client/README.md
```diff
@@ -55,36 +55,48 @@ Welcome to Kuksa Client version <some_version>
 
 Default tokens directory: /some/path/kuksa_certificates/jwt
 
-connectj to wss://127.0.0.1:8090
-Websocket connected securely.
+Connecting to VSS server at 127.0.0.1 port 55556 using KUKSA GRPC protocol.
```
erikbosch, line 58: use the default port 55555, not your own mac port

kuksa-client/kuksa_client/__main__.py
```diff
@@ -527,11 +511,35 @@ def connect(self):
...
+        elif srv.scheme in ["ws", "wss"]:
+            config["protocol"] = "ws"
+        else:
+            print(f"Invalid server URI. Unsupported protocol: {srv.scheme} ")
+            return
+
+        if srv.port is not None:
+            config["port"] = srv.port
+
+        if srv.scheme in ["grpcs", "wss"]:
```
erikbosch, line 530: As default value for `config["insecure"]` may change in the future, shall we also set it explicitly if "grpc" or "ws" is used, i.e. add an else branch where we set it True?

erikbosch:
I have tested Databroker functionality, works fine. When this is merged we should update https://github.com/eclipse/kuksa.val/wiki/Release-Testing, or at least add a note that it needs to be updated (as part of preparation for next release)

kuksa-client/README.md
```diff
@@ -55,36 +55,48 @@ Welcome to Kuksa Client version <some_version>
...
+TLS will not be used.
+INFO 2023-09-15 18:48:13,415 kuksa_client.grpc No Root CA present, it will not be posible to use a secure connection!
+INFO 2023-09-15 18:48:13,415 kuksa_client.grpc.aio Establishing insecure channel
+gRPC channel connected.
```
lukasmittag, line 62: shouldn't it connect to websocket in this example?

kuksa-client/README.md
```diff
@@ -27,11 +27,11 @@ After you have installed the kuksa-client package via pip you can run the test c
...
 ```
 
-With default CLI arguments, the client will try to connect to a local VISS server e.g. `kuksa-val-server`.
-If you wish to connect to a gRPC server e.g. `kuksa-databroker`, you should instead run:
+With default CLI arguments, the client will try to connect to a local databroker, e.g. a server supporting the `kuksa.val.v1` protocol..
+If you wish to connect to a VISS server e.g. `kuksa-val-server`, you should instead run:
 
 ```console
-$ kuksa-client --ip 127.0.0.1 --port 55555 --protocol grpc
+$ kuksa-client ws://127.0.0.1:8090
```
lukasmittag, line 34: show ws connection and grpc connection command. currently I only see ws connection

lukasmittag (changes requested):

kuksa-client/README.md
SebastianSchildt: Fixed

kuksa-client/README.md
SebastianSchildt: I reordered it know, according to Eriks comment: It first shows the GRPC output, and AFTERWARDs mentions the command to connect to WS (without giving any output). Should be less confusing

SebastianSchildt:
Addressed findings. Somebody should still test if TLS and val-server connection works as expected

kuksa-client/README.md
SebastianSchildt: Done

kuksa-client/README.md
```diff
@@ -55,36 +50,51 @@ Welcome to Kuksa Client version <some_version>
...
+
+```
+grpcs://127.0.0.1:55555
+```
+
+To contact a TLS enabled local val-server the uri is
 
-### Connecting to KUKSA Databroker
+```
+wss://127.0.0.1:8090
```
lukasmittag, line 78: thats doubled

kuksa-client/README.md
```diff
@@ -95,23 +105,22 @@ Currently `Server` and `localhost` are valid names from the example certificates
 ~/kuksa.val/kuksa-client$ kuksa-client --ip 127.0.0.1 --port 55555 --protocol grpc  --cacertificate ../kuksa_certificates/CA.pem --tls-server-name Server
 ```
 
-### Connecting to KUKSA Server
+### TLS with val-server
```
lukasmittag, line 108: double

kuksa-client/README.md
```diff
@@ -95,23 +105,22 @@ Currently `Server` and `localhost` are valid names from the example certificates
...
 ```
-~/kuksa.val/kuksa-client$ kuksa-client
+~/kuksa.val/kuksa-client$ kuksa-client wss://127.0.0.1:8090
 ```
 
 This corresponds to this call:
 
 ```
-kuksa-client --ip 127.0.0.1 --port 8090 --protocol ws --cacertificate ./kuksa_certificates/CA.pem
+kuksa-client --cacertificate ./kuksa_certificates/CA.pem wss://127.0.0.1:8090
```
lukasmittag, line 119:
```suggestion
kuksa-client --cacertificate ../kuksa_certificates/CA.pem --tls-server-name server
```
otherwise it fails for me at least

kuksa-client/README.md
```diff
@@ -55,36 +50,51 @@ Welcome to Kuksa Client version <some_version>
...
-~/kuksa.val/kuksa-client$ kuksa-client --ip localhost --port 55555 --protocol grpc --certificate ../kuksa_certificates/Client.pem --keyfile ../kuksa_certificates/Client.key --cacertificate ./kuksa_certificates/CA.pem
+~/kuksa.val/kuksa-client$ kuksa-client --certificate ../kuksa_certificates/Client.pem --keyfile ../kuksa_certificates/Client.key --cacertificate ./kuksa_certificates/CA.pem grpcs://localhost:55555
 ```
 
 There is actually no reason to specify client key and certificate, as mutual authentication is not supported in KUKSA Databroker,
 so the command can be simplified like this:
 
 ```
-~/kuksa.val/kuksa-client$ kuksa-client --ip localhost --port 55555 --protocol grpc --cacertificate ./kuksa_certificates/CA.pem
+~/kuksa.val/kuksa-client$ kuksa-client  --cacertificate ./kuksa_certificates/CA.pem grpcs://localhost:55555
```
lukasmittag, line 97:
```suggestion
~/kuksa.val/kuksa-client$ kuksa-client  --cacertificate ./kuksa_certificates/CA.pem --tls-server-name server grpcs://localhost:55555
```
otherwise it fails but I do not know if we want to add :)

lukasmittag (commented):
So basicallay looks better but two inor comments (see below)

kuksa-client/README.md
erikbosch: And line 105 below must be changed to new syntax

kuksa-client/kuksa_client/__main__.py
```diff
@@ -149,10 +147,11 @@ def subscriptionIdCompleter(self, text, line, begidx, endidx):
     VSS_COMMANDS = "Kuksa Interaction Commands"
     INFO_COMMANDS = "Info Commands"
 
-    ap_getServerAddr = argparse.ArgumentParser()
     ap_connect = argparse.ArgumentParser()
     ap_connect.add_argument(
-        '-i', "--insecure", default=False, action="store_true", help='Connect in insecure mode')
+        'server', help=f"VSS server to connect to. Format: protocol://host[:port]. \
```
erikbosch, line 152: Do we need to state somewhere that default port is 55555 regardless of protocol.
Or do we want it to be smarter and let default port vary depending on protocol to match default port for databroker and val-server

kuksa-client/kuksa_client/__main__.py
```diff
@@ -527,11 +511,35 @@ def connect(self):
...
                 self.commThread = None
-        config = {'ip': self.serverIP,
-                  'port': self.serverPort,
-                  'insecure': self.insecure,
-                  'protocol': self.serverProtocol
-                  }
+
+        # Check we have a valid server URI
+        srv = urlparse(self.server)
+        config = {"port": 55555, "insecure": True}
```
erikbosch, line 517: See comment above, do we want different default for port depending on protocol, i.e. use 8090 if protocol is ws?

kuksa-client/README.md
lukasmittag: Ah yes with localhost it works seems to be this problem :)

kuksa-client/kuksa_client/__main__.py
SebastianSchildt: I did not state it but choose sensible default ports now. AS the client tells you exactly what it is doing before it connects, I think no need to explicitly spell it out in usage message

SebastianSchildt:
Comments addressed. I added commits, so it is easier to see what I did. Once this is greenlighted might squash a little bit

lukasmittag (approved):
LGTM

kuksa-client/README.md
```diff
@@ -144,18 +148,19 @@ Test Client> help -v
 
 Documented commands (use 'help -v' for verbose/'help <topic>' for details):
 
+Documented commands (use 'help -v' for verbose/'help <topic>' for details):
```
erikbosch, line 151: Duplicate of line 149?

PR #662, opened 2023-09-21
port databroker cli to kuksa.val.v1

lukasmittag:
This uses the new API kuksa.val.v1 for the a rust sdk. It adds target value handeling and the ability to set metadata. Subscribe is left on old API because of the query ability that is not (yet/ever) supported by the new kuksa.val.v1. The other parts are ported to kuksa.val.v1 without changing the behaviour. The only change is that in the new API ChangeType is not existent so it can't be displayed for metadata query.


SebastianSchildt:
@lukasmittag , is this replaced by #675 ? If yes, maybe close this, if not, disregard this message :D

PR #663, opened 2023-09-21
kuksa-client: exit is an alias for quit now

SebastianSchildt:
Sometimes people - including me -  try to type `exit` instead of `quit` to exit kuksa-client.

This PR makes sure exit and quit are equivalent.


lukasmittag (approved):
looks good and works as expected

PR #664, opened 2023-09-21
Documentation update

### LINKED ISSUE #659: Links broken
The link to the TLS documentation in kuksa_certiicate/README.md is broken and should be updated. The link in doc/tls.md to kuksa_certificate/README.md is broken too

erikbosch:
Fixes #659

Also clarifies jwt token naming for server and need to specify servername for tls

How to check links is described as part of release process at https://github.com/eclipse/kuksa.val/wiki/Release-Process#checking-links, I just improved it a bit.

### DIFF
--- a/NOTICE.md
+++ b/NOTICE.md
@@ -17,7 +17,7 @@ source code repository logs.
 ## Declared Project Licenses
 
 This program and the accompanying materials are made available under the terms
-of the  Apache License, Version 2.0 
+of the  Apache License, Version 2.0
 
 You may obtain a copy of the License at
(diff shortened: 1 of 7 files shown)


lukasmittag (approved):
looks good couldn't verify links by clicking but had a look if they make sense.

SebastianSchildt (approved):
Content-wise looks good to me.

I noticed, `NOTICE.md` might need to go away eventually, as it is not really current, and we are doing the Dashfiles in release and BOMs in the artifacts. But for no taking the editorial changes

This also means the logic in #661 is fine, right?

Issue #652, opened 2023-09-12
Subscription: Updates on each setValue


nayakned:
For quite a few usecases (e.g. heartbeat signals) it is essential that the update messages to a subscription are sent each time the corresponding datapoint is set (irrespective of whether the value changes or not). Currently this is by default the case with target values (which is also not good for applications which need to triggered only on change of values). 

Suggestion: Add an optional flag during subscription request to figure out if the subscribing application needs the information when set(Target)Value is called or only when the value changes.

argerus:
I agree with the problem description.

So I see two distinct observations here. 

### Values
> Suggestion: Add an optional flag during subscription request to figure out if the subscribing application needs the information when set(Target)Value is called or only when the value changes.

This can make sense. But, a related question is whether this is something that should be decided by the client, or if this is an inherent property of the signal/datapoint. If it is an inherent property of the signal, subscribing to it should _always_ behave in accordance with that.

The VHAL implementation in Android has decided that this is an inherent property of the signal, and this is also how the implementation works internally in databroker. If a signal/datapoint is `ChangeType::OnChange` it will send notifications only when the signal/datapoint changes. If it is `ChangeType::Continous`, it will send notifications periodically* (regardless of whether the value changes).

Firstly, this information is not part of VSS though. But since VSS supports adding additional metadata to signals, the only thing missing is adding support for reading this extended metadata in databroker. Right now, it defaults to `ChangeType::OnChange` as you've noticed, (maybe that's the wrong default).

The second part is that a client should be able to supply a preferred update frequency when subscribing to `ChangeType::Continuous` type of signals, as to not introduce more load to the client or system than necessary.

This "solution" might seem more complicated than just letting the client decide. But another way to look at it, is that this is metadata (just as specifying the data type) that is needed in order to implement correctly behaving providers of these signals. And in a sense, also correctly behaving clients.

### Target values
> update messages to a subscription are sent each time [...] (irrespective of whether the value changes or not) [...] is by default the case with target values (which is also not good for applications which need to triggered only on change of values).

Setting a target value is distinct from setting a value. It is more akin to calling a function implemented by an actuator provider. If the provider is not available at the time of the call, it will not be notified. Calling the same "function" again, with the same value, when the provider _is_ available should notify the provider even if the target value doesn't change. Because of that, I think it makes sense to send a notification every time (regardless of what the previous **target value** was). It might make sense to avoid this notification if the target value doesn't differ from the current **value**, though.

[...]

PR #669, opened 2023-09-22
Configure change-type from VSS JSON

SebastianSchildt:
Work in progress.

Tries to let user determine change_type (continous/on_change) via VSS model as suggested in #652 

The behavior is as follows:

With a standard VSS all sensor and actuator signals are registered with ChangeType `continuous` (I feel, as VSS itself does not specify this, continuous is a more "safe" choice in terms of expectations. Everything that comes in via provider is forwarded to subscribers. But this is just a default.) 

Attributes are set to ChangeType `static` - if have no idea what implications that has in databroker, is this correct?

If a signal defines the extra proper x-kuksa-changetype being either `onchange`, `continuous` or `static` the ChangeType is set respectively. 

To test: Start databroker, start to kuksa-clients, subscribe e.g. `Vehicle.Speed` in one of them, and do a `setValue` for the same in another client. You will see all updates on the subscriber side, even when you repeatedly send the same value. This is different from the behavior in master.

You can change this behavior for any sensor by setting `x-kuksa-changetype`in the metadata.json (this also works through vss-tools, using e.g. an overlay), as an example consider this snippet

```json
"Speed": {"datatype": "float", "description": "Vehicle speed.", "x-kuksa-changetype": "onchange" , "type": "sensor", "unit": "km/h", "uuid": "efe50798638d55fab18ab7d43cc490e9"}
```

If the experiment is repeated, as above, if you send the same value the subscriber is only informed once.

Should this potentially also be returned in metadata? If so, where would be the place to put it into databroker, and would doing so break API compatibility?

### DIFF
--- a/kuksa_databroker/doc/user_guide.md
+++ b/kuksa_databroker/doc/user_guide.md
@@ -15,6 +15,7 @@ The following sections provide information for running and configuring Databroke
     <li><a href="#query-syntax">Query Syntax</a></li>
     <li><a href="#using-custom-vss-data-entries">Using Custom VSS Data Entries</a></li>
     <li><a href="#configuration-reference">Configuration Reference</a></li>
+    <li><a href="#signal-change-types">Signal Change Types</a></li>
     <li><a href="#api">API</a></li>
     <li><a href="#known-limitations">Known Limitations</a></li>
   </ol>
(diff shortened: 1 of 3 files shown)


erikbosch:
Just a thought, do we see a future need to provide additional data to control update, like min/max-frequency? If so, would we then prefer them to be grouped by a single keyword? I have no real problem with any of the alternatives below, and if we agree that the first alternative would be the best for the future then I have no problem with the current PR.


```
# Send value whenever it changes, but only if it differs more than 0.5 from previously sent value
{
"x-kuksa-changetype": "onchange",
"x-kuksa-threshold": 0.5 
}
```

... or ...

```
# Send value whenever it changes, but only if it differs more than 0.5 from previously sent value
{
"x-kuksa": {
       "changetype": "onchange",
       "threshold": 0.5
    } 
}
```

SebastianSchildt:
my personal opinion is, to keep it flat, i.e. option 1. Firstly it is not clear if/when/what we will introduce here, secondly if other adopt it, it might be easier to have a flat strucutre, i.e. maybe someone decide in his implementation to only support a subset of x-kuksa  attributes, or maybe one day COVESA decides to make "changetype" offcial/recommended, then we just remove the prefix, and not need to move it "out" of a KUKSA data structure

erikbosch:
Tested all possible combinations. To me it seems that "onchange" and "static" always gives the same behavior. And as hinted by Sebastian, default is onchange/static for attributes, continuous for sensor/actuator.

Found a possible bug - when subscribing for targetvalue like `Test Client> subscribe -a targetValue A.Actuator` you always seem to get "continuous", regardless of `x-kuksa-changetype`. Or do we possibly see a need to be able to specify different setting for target value, like  `x-kuksa-changetype-target`?

SebastianSchildt:
I think the thing with "target" being "continuous is currently "by design", I'd say currently we just merge this to make the default behaviour more consistent/expected, and check what we can do  API-wise, and how the extra attributes should influence things later.

erikbosch:
> I think the thing with "target" being "continuous is currently "by design", I'd say currently we just merge this to make the default behaviour more consistent/expected, and check what we can do API-wise, and how the extra attributes should influence things later.

But does this need to be documented somewhere, maybe a table like this that defines expected behavior:


signal Type | attribute | default | static | onchange | continuous
-------------|-----------|---------|-------|-------------|-------------
attribute  | value | onchange | onchange| onchange | continuous
sensor  | value | continuous | onchange| onchange | continuous
actuator  | value | continuous | onchange| onchange | continuous
actuator | targetValue | continuous |  continuous|  continuous | continuous

SebastianSchildt:
Documented the behavior now

kuksa_databroker/doc/user_guide.md
```diff
@@ -168,6 +169,50 @@ docker run --rm -it -p 55555:55555 ghcr.io/eclipse/kuksa.val/databroker:master -
 ```
 <p align="right">(<a href="#top">back to top</a>)</p>
 
+## Signal Change Types
+
+Internally, databroker knows different change types for VSS signals. There are three change-types
+
+ - **Continuos**: This are usually sensor values that are continuos, such as vehicle speed. Whenever a continuos signal is updated by a provider, all subscribers are notified.
```
erikbosch, line 176: Check spelling - "continuous"

kuksa_databroker/doc/user_guide.md
```diff
@@ -168,6 +169,50 @@ docker run --rm -it -p 55555:55555 ghcr.io/eclipse/kuksa.val/databroker:master -
...
+  datatype: string
+  type: attribute
+  x-kuksa-changetype: static
+  description: 17-character Vehicle Identification Number (VIN) as defined by ISO 3779.
+
+Vehicle.Speed:
+  datatype: float
+  type: sensor
+  unit: km/h
+  x-kuksa-changetype: continuos
```
erikbosch, line 204: "continuous"

kuksa_databroker/doc/user_guide.md
```diff
@@ -168,6 +169,50 @@ docker run --rm -it -p 55555:55555 ghcr.io/eclipse/kuksa.val/databroker:master -
 ```
 <p align="right">(<a href="#top">back to top</a>)</p>
 
+## Signal Change Types
+
+Internally, databroker knows different change types for VSS signals. There are three change-types
+
+ - **Continuos**: This are usually sensor values that are continuous, such as vehicle speed. Whenever a continuous signal is updated by a provider, all subscribers are notified.
```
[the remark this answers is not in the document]
SebastianSchildt: Definitely 110% fixed now. 🤞

erikbosch (approved):

Issue #589, opened 2023-06-30
Configuration policy


erikbosch:
The various KUKSA.val components as of today (Broker, Server, Kuksa-client, dbcfeeder, ...) today use quite different approaches to configuration. They typically support a combination of config-file, command line arguments and env-variables, but it is not consistently implemented and precedence is not always well defined and may differ. Supporting all 3 options (cli-argument, env-variable, config file) makes code somehwat complex as we then need to consider order of precedence 

Would it be an idea to try agree on a general concept? And in the long run - or at least for new components - try to use the same principle for all.

Some ideas:

**Config file "only"**

No env-variables, as few cli-arguments as possible, typically only help, version and `--config`. Use a config file for the rest

SebastianSchildt:
Having a unified concept would be good.

Normally I am not a fan of environment variables, but those are a pretty common concept in the container landscape. Although a pattern might be, that the containers have some extra scripting to inject those into config files.

argerus:
> Supporting all 3 options (cli-argument, env-variable, config file) makes code somehwat complex as we then need to consider order of precedence

This might be true, but I think the more important question to ask is whether **using** our applications would get easier by any proposed change.

We implement the configuration handling once (and maybe we should look into more reuse between different applications), but users of our applications will configure them countless of times.

> Some ideas:
>
>Config file "only"
>
>No env-variables, as few cli-arguments as possible, typically only help, version and --config. >Use a config file for the rest

Removing support for both environment variables and command line arguments would definitely make our applications **harder** to configure in a lot of use cases.

> Although a pattern might be, that the containers have some extra scripting to inject those into config files.

Yeah, definitely not *easier* than the current situation. Editing a file is a lot more effort than either supplying a command line argument or an environment variable (and more permanent). And as Sebastian also points out, it's pretty much standard practice to inject configuration using environment variables in container environments.

However, config files are a great way to be able to configure *default values*. But it's only great if it's also easy to also override them.

Regarding priority, I think it's pretty well defined what every "posix sysadmin" would expect.

From highest to lowest:
* command line arguments (overrides everything else)
* environment variables
* local config file
* system config file (i.e. in /etc/)
* default values

argerus:
> I have no problem with the "triple" approach. But do we want it for all possible configurations, or do we consider it OK to have cli-argument + env-variable only for configs we see as likely that someone wants to change or for all.

The problem is determining what a user wants to be able to change easily. But sure, that's a possibility.

[...]

PR #672, opened 2023-09-25
Update configuration of python-cli to configuration policy

lukasmittag:
To meet the discussed configuration policy here https://github.com/eclipse/kuksa.val/issues/589 this adds in a first step some environment variables for the cli arguments. Futhermore this should be a place to discuss the removal of default tokens. The default tokens were not used anymore because they got overwritten. This gets fixed with this PR. 

The question is do we want to keep default tokens as default usage in or do we just want to keep them in the container that they can manually be used.

@argerus @erikbosch @SebastianSchildt @rafaeling

### DIFF
--- a/kuksa-client/kuksa_client/__main__.py
+++ b/kuksa-client/kuksa_client/__main__.py
@@ -43,9 +43,15 @@ import kuksa_certificates
 from kuksa_client import KuksaClientThread
 from kuksa_client import _metadata
 
-DEFAULT_SERVER_ADDR = "grpc://127.0.0.1:55555"
-
 scriptDir = os.path.dirname(os.path.realpath(__file__))
+
+DEFAULT_SERVER_ADDR = os.environ.get("SERVER_ADDR", "grpc://127.0.0.1:55555")
+DEFAULT_LOGGING_CONFIG = os.environ.get("LOGGING_CONFIG", os.path.join(scriptDir, 'logging.ini'))
+DEFAULT_TOKEN_OR_TOKENFILE = os.environ.get("TOKEN_OR_TOKENFILE", None)
+DEFAULT_CERTIFICATE = os.environ.get("CERTIFICATE", None)
+DEFAULT_KEYFILE = os.environ.get("KEYFILE", None)
+DEFAULT_CACERTIFICATE = os.environ.get("CACERTIFICATE", None)
+DEFAULT_TLS_SERVER_NAME = os.environ.get("TLS_SERVER_NAME", None)
 
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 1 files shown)


kuksa-client/kuksa_client/__main__.py
```diff
@@ -43,10 +43,16 @@
...
 from kuksa_client import _metadata
 
-DEFAULT_SERVER_ADDR = "grpc://127.0.0.1:55555"
-
 scriptDir = os.path.dirname(os.path.realpath(__file__))
 
+DEFAULT_SERVER_ADDR = os.environ.get("SERVER_ADDR", "grpc://127.0.0.1:55555")
+DEFAULT_LOGGING_CONFIG = os.environ.get("LOGGING_CONFIG", os.path.join(scriptDir, 'logging.ini'))
+DEFAULT_TOKEN_OR_TOKENFILE = os.environ.get("TOKEN_OR_TOKENFILE", None)
+DEFAULT_CERTIFICATE = os.environ.get("CERTIFICATE", None)
```
erikbosch, line 51: What is our view on mutual authentication? AFAIK we do not currently support it, as broker does not check/request client cert. I would prefer if we either spend some effort to support it, or conclude that it likely never will be supported and remove all client certs

kuksa-client/kuksa_client/__main__.py
```diff
@@ -43,10 +43,16 @@
 from kuksa_client import KuksaClientThread
 from kuksa_client import _metadata
 
-DEFAULT_SERVER_ADDR = "grpc://127.0.0.1:55555"
-
 scriptDir = os.path.dirname(os.path.realpath(__file__))
 
+DEFAULT_SERVER_ADDR = os.environ.get("SERVER_ADDR", "grpc://127.0.0.1:55555")
```
erikbosch, line 48: Should we try to agree on a long term (ideal) pattern. Like here:

* [Databroker](https://github.com/eclipse/kuksa.val/tree/master/kuksa_databroker) uses `KUKSA_DATA_BROKER_ADDR`
* [Dbcfeeder](https://github.com/eclipse/kuksa.val.feeders/tree/main/dbc2val) uses `KUKSA_ADDRESS`

Here we have a third option. Another topic - are we fine with names that possibly could clash with other env variables, or do we prefer to always have a `KUKSA_ `prefix?

kuksa-client/kuksa_client/__main__.py
SebastianSchildt: I would vote for removing all that. I don't think they even enable additional security goals over authorisation tokens, which also need to be signed by a trusted authority.
The only advantage is, that TLS cleint auth kicks in "earlier" in the stack on transport level, but even though: We did not have it implemented and it is not on the roadmap, so let's rather clean up here. IMO not much we loose if at a later point we DO decide to implement this.

kuksa-client/kuksa_client/__main__.py
lukasmittag: I like the ```KUKSA_``` option. Will change. No strong opinion on what comes after it.

erikbosch: If the address can be used for both Server And Broker then I think `KUKSA_ADDRESS` is good

erikbosch (commented):
What is the conclusion of client certificates (for mutual auth). Remove as part of this PR?

If so we need to decide how "deep" we want to remove them as part of this PR. Only in client, but also the client key/cert itself and the script generating it?

SebastianSchildt:
I think remove them, but potentially not necessarily in this PR. @erikbosch as you have reviewed this, you have a better overview, if it is "trivial" or we rather do it separately (I think we would also need to scour documentation, and remove informaton there)

erikbosch:
> I think remove them, but potentially not necessarily in this PR. @erikbosch as you have reviewed this, you have a better overview, if it is "trivial" or we rather do it separately (I think we would also need to scour documentation, and remove informaton there)

Lets do it separately later. Could possibly require some extra work here and there to remove all traces of it.

erikbosch (approved):

erikbosch:
@lukasmittag @SebastianSchildt - Is this a merge candidate to include in v0.4.1?

erikbosch:
@lukasmittag - do you want to rebase and fix conflict? So it can be included in 0.4.1

lukasmittag:
will do

lukasmittag:
@erikbosch

erikbosch (approved):

Issue #668, opened 2023-09-22
Databroker: Hardcoded metadata always added. Remove it?


SebastianSchildt:
Do we still want the hardcoded metadata? I think it is ieasy enough to load a VSS model, and our containers ship with batteires included.

Furthermore, there seems to be a bug, were metadata is always loaded (I think the intention is to only do it, when `--dummy-metadata` is set, i.e. see

```
docker run -it --rm --net=host ghcr.io/eclipse/kuksa.val/databroker:master  --insecure --port 55556
2023-09-22T23:00:16.960912Z  INFO databroker: Init logging from RUST_LOG (environment variable not found)
2023-09-22T23:00:16.960937Z  INFO databroker: Starting Kuksa Databroker 4d758d4967cfa6af583cab61457fd30c13499dbe
2023-09-22T23:00:16.960941Z  INFO databroker: Populating (hardcoded) metadata <------- WHY?
2023-09-22T23:00:16.960968Z  INFO databroker: Populating metadata from file 'vss_release_4.0.json'
2023-09-22T23:00:17.063353Z  WARN databroker: Authorization is not enabled.
2023-09-22T23:00:17.063427Z  INFO databroker::broker: Starting housekeeping task
2023-09-22T23:00:17.063436Z  INFO databroker::grpc::server: TLS is not enabled
2023-09-22T23:00:17.063438Z  INFO databroker::grpc::server: Authorization is not enabled.
2023-09-22T23:00:17.063453Z  INFO databroker::grpc::server: Listening on 0.0.0.0:55556
```

Also I think, when the JSON is loaded it will not override (all) settings that may be in the hardcoded metadata, but I am not sure. as I was hacking on the  JSON parsing anway (so my code might be buggy),

In any case IMO the cleanest "solution" here is getting rid of the hard coded metadata, as it is not a very obvious (helpful?) feature anyway

argerus:
> In any case IMO the cleanest "solution" here is getting rid of the hard coded metadata

Agreed.

erikbosch:
Fully fixed now @SebastianSchildt ?

SebastianSchildt:
It is. Unless somebody says otherwise

PR #673, opened 2023-09-26
Remove dummy VSS signals...

SebastianSchildt:
... and adding some databroker version info to VSS tree instead. This might be useful, and avoids corner cases in databroker, where it has an empty/no tree

Currently added data

```console
getTest Client> getMetaData Vehicle.KUKSA.Databroker
[
    {
        "path": "Vehicle.KUKSA.Databroker.GitVersion",
        "metadata": {
            "data_type": "STRING",
            "entry_type": "ATTRIBUTE",
            "description": "Databroker version as reported by GIT"
        }
    },
    {
        "path": "Vehicle.KUKSA.Databroker.CargoVersion",
        "metadata": {
            "data_type": "STRING",
            "entry_type": "ATTRIBUTE",
            "description": "Databroker version as reported by GIT"
        }
    },
    {
        "path": "Vehicle.KUKSA.Databroker.CommitSha",
        "metadata": {
            "data_type": "STRING",
            "entry_type": "ATTRIBUTE",
            "description": "Commit Sha of current version"
        }
    }
]
```

If we consider this databroker data useful, I'd prefer to have it under a "special" path such as `Vehicle._kuksa` but there is currently a bug (#674 ) in databroker that prevents accessing such paths in all conditions.

This will solve #668

### DIFF
--- a/kuksa_databroker/databroker/src/main.rs
+++ b/kuksa_databroker/databroker/src/main.rs
@@ -31,86 +31,6 @@ use clap::{Arg, ArgAction, Command};
 
 use databroker::{broker, grpc, permissions, vss};
 
-// Hardcoded datapoints
-const DATAPOINTS: &[(
-    &str,
-    broker::DataType,
-    broker::ChangeType,
-    broker::EntryType,
-    &str,
-)] = &[
-    (
-        "Vehicle.Cabin.Seat.Row1.Pos1.Position",
-        broker::DataType::Uint32,
-        broker::ChangeType::OnChange,
-        broker::EntryType::Actuator,
-        "Vehicle cabin seat position. Row 1 Position 1",
-    ),
-    (
-        "Vehicle.Cabin.Seat.Row1.Pos2.Position",
-        broker::DataType::Uint32,
-        broker::ChangeType::OnChange,
-        broker::EntryType::Actuator,
-        "Vehicle cabin seat position. Row 1 Position 2",
-    ),
-    (
-        "Vehicle.Speed",
-        broker::DataType::Float,
-        broker::ChangeType::Continuous,
-        broker::EntryType::Sensor,
-        "Vehicle speed",
-    ),
-    (
-        "Vehicle.ADAS.ABS.Error",
-        broker::DataType::String,
-        broker::ChangeType::OnChange,
-        broker::EntryType::Attribute,
-        "ADAS ABS error message",
-    ),
-    (
-        "Vehicle.ADAS.ABS.IsActive",
-        broker::DataType::Bool,
-        broker::ChangeType::OnChange,
-        broker::EntryType::Actuator,
-        "Whether ADAS ABS is active",
-    ),
-    (
-        "Vehicle.ADAS.ABS.IsEngaged",
-        broker::DataType::Bool,
-        broker::ChangeType::OnChange,
-        broker::EntryType::Actuator,
-        "Whether ADAS ABS is engaged",
-    ),
-    (
-        "Vehicle.ADAS.CruiseControl.Error",
-        broker::DataType::String,
-        broker::ChangeType::OnChange,
-        broker::EntryType::Attribute,
-        "ADAS Cruise control error message",
-    ),
-    (
-        "Vehicle.ADAS.CruiseControl.IsActive",
-        broker::DataType::Bool,
-        broker::ChangeType::OnChange,
-        broker::EntryType::Actuator,
-        "Whether ADAS Cruise control is active",
-    ),
-    (
-        "Vehicle.ADAS.CruiseControl.SpeedSet",
-        broker::DataType::Bool,
-        broker::ChangeType::OnChange,
-        broker::EntryType::Actuator,
-        "Whether ADAS Cruise control has the speed set",
-    ),
-    (
-        "Vehicle.TestArray",
-        broker::DataType::StringArray,
-        broker::ChangeType::OnChange,
-        broker::EntryType::Sensor,
-        "Run of the mill test array",
-    ),
-];
-
 async fn shutdown_handler() {
     let mut sigint =
         signal(SignalKind::interrupt()).expect("failed to setup SIGINT signal handler");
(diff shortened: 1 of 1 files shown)


argerus:
I don't think corner cases are an issue really. But it could make sense to be able to access this information using the same methods as when accessing other signals.

It does seem weird to place them under `Vehicle` though, as they are not really "Vehicle signals".

`Kuksa.Databroker.Version` or `Vendor.Kuksa.Databroker.Version` might make more sense. And if they are namespaced with `Kuksa`, is there really a need to add a leading underscore in there?

And should these signals require specific authorization in order to access them?

argerus:
> If we consider this databroker data useful, I'd prefer to have it under a "special" path such as Vehicle._kuksa

See discussion in #674, but to address this specific point here:

Instead of creating a "special" path by introducing a leading underscore (which as I understand it, you think should be allowed for any externally defined signal as well), an alternative is to make this a "special" path by placing it under a completely separate root (as suggested above).

That would guarantee that it would not clash with any externally defined signal if all of those are assumed to have Vehicle as the root.

SebastianSchildt:
I know moved the data to `Kuksa.Databroker` and confirmed also, there are no issues when loading a VSS model starting at `Vehicle`

I think currently there is no need for any "special" authorisation required, and since it is now neatly separated, with authorisation enabled, it works "just as usual",  so if somebody wants it, he can add the prefix to the token scope.

erikbosch:
I triggered a re-run of the failed check and then it succeeded. It seemed to be some temporary hickup, a 403 error when trying to download something

erikbosch:
* I downloaded the PR.  Then `getValue Kuksa.Databroker.GitVersion` does not work. The information is available, as both `git describe --tags` and kuksa-client reports version.
* Also - should we possibly document the 3 default signals somewhere
* And would it be an idea to print the values as INFO when starting Databroker? Could be useful as log info if someone reports that Databroker does not work as expected

```
erik@debian3:~/kuksa.val/kuksa-client$ kuksa-client

     ⢀⣤⣶⣾⣿⢸⣿⣿⣷⣶⣤⡀
    ⣴⣿⡿⠋⣿⣿   ⠈⠙⢿⣿⣦
   ⣾⣿⠋  ⣿⣿  ⣶⣿  ⠙⣿⣷
  ⣸⣿⠇   ⣿⣿⠠⣾⡿⠃   ⠸⣿⣇  ⣶ ⣠⡶⠂ ⣶  ⢰⡆ ⢰⡆⢀⣴⠖ ⢠⡶⠶⠶⡦   ⣰⣶⡀
  ⣿⣿    ⠿⢿⣷⣦⡀     ⣿⣿  ⣿⢾⣏   ⣿  ⢸⡇ ⢸⡷⣿⡁  ⠘⠷⠶⠶⣦  ⢠⡟⠘⣷
  ⢹⣿⡆   ⣿⣶⠈⢻⣿⡆   ⢰⣿⡏  ⠿ ⠙⠷⠄ ⠙⠷⠶⠟⠁ ⠸⠇⠈⠻⠦ ⠐⠷⠶⠶⠟ ⠠⠿⠁ ⠹⠧
   ⢿⣿⣄  ⣿⣿  ⠿⣿  ⣠⣿⡿
    ⠻⣿⣷⡄⣿⣿   ⢀⣠⣾⣿⠟    kuksa-client CLI
     ⠈⠛⠇⢿⣿⣿⣿⣿⡿⠿⠛⠁     0.4.0.post34+git.5f4e01db.dirty

Default tokens directory: /home/erik/kuksa.val/kuksa_certificates/jwt

Connecting to VSS server at 127.0.0.1 port 55555 using KUKSA GRPC protocol.
TLS will not be used.
INFO 2023-09-29 11:13:05,230 kuksa_client.grpc No Root CA present, it will not be posible to use a secure connection!
INFO 2023-09-29 11:13:05,230 kuksa_client.grpc.aio Establishing insecure channel
gRPC channel connected.
Test Client> getValue Kuksa.Databroker.GitVersion  
{
    "path": "Kuksa.Databroker.GitVersion",
    "value": {
        "value": "N/A",
        "timestamp": "2023-09-29T08:45:20.334727+00:00"
    }
}

Test Client> quit
gRPC channel disconnected.
erik@debian3:~/kuksa.val/kuksa-client$ git describe --tags
0.4.0-34-g5f4e01d
erik@debian3:~/kuksa.val/kuksa-client$ 
```

SebastianSchildt:
Whether the information is available depends on "in which state" the working copy is, when building, currently this is the same you you get when doing --help . (I did not look deeper into what we are doing when buiding here, and waht vergen is doing to get the information as opposed to the py build system)

As we had a discussion before how to "determine"/report versions (tags? Cargo?) before, here I opted the report verbatim what  was available at build time. Currently I would see it as "debugging/engineering info, whre likely for "released" version Cargo version number is all that is needed, whereas the has unambiguously identifies where you were no matter the build.

SebastianSchildt:
Put the CI "weirdness" in a separate issue https://github.com/eclipse/kuksa.val/issues/677

Changing this/or using it in clients potentially can be in other PRs.

Merging this for now

erikbosch:
> Whether the information is available depends on "in which state" the working copy is, when building, currently this is the same you you get when doing --help . (I did not look deeper into what we are doing when buiding here, and waht vergen is doing to get the information as opposed to the py build system)
> 
> As we had a discussion before how to "determine"/report versions (tags? Cargo?) before, here I opted the report verbatim what was available at build time. Currently I would see it as "debugging/engineering info, whre likely for "released" version Cargo version number is all that is needed, whereas the has unambiguously identifies where you were no matter the build.

For info @SebastianSchildt , I tested also on the "official master build" and also there I get N/A:
```

erik@debian3:~/kuksa.val/kuksa-client$ docker run --rm -it --net=host  ghcr.io/eclipse/kuksa.val/databroker:master --help
Commit Date:      2023-09-29T09:59:05.000000000Z
  Commit SHA:       f716273b74f7a2612307d0a810f3e581f35c893c
  Commit Branch:    master

  Package version:  0.4.0
  Debug build:      false

Usage: databroker [OPTIONS]

Options:
      --address <IP>            Bind address [env: KUKSA_DATA_BROKER_ADDR=0.0.0.0] [default: 127.0.0.1]
      --port <PORT>             Bind port [env: KUKSA_DATA_BROKER_PORT=55555] [default: 55555]
      --vss <FILE>              Populate data broker with VSS metadata from (comma-separated) list of files [env: KUKSA_DATA_BROKER_METADATA_FILE=vss_release_4.0.json]
      --jwt-public-key <FILE>   Public key used to verify JWT access tokens
      --tls-cert <FILE>         TLS certificate file (.pem)
      --tls-private-key <FILE>  TLS private key file (.key)
      --disable-authorization   Disable authorization
      --insecure                Allow insecure connections
  -h, --help                    Print help
  -V, --version                 Print version
```


What client reports:

```
Test Client> getValue Kuksa.Databroker.GitVersion
{
    "path": "Kuksa.Databroker.GitVersion",
    "value": {
        "value": "N/A",
        "timestamp": "2023-09-29T12:40:06.867812+00:00"
    }
}

Test Client> getValue Kuksa.Databroker.CargoVersion
{
    "path": "Kuksa.Databroker.CargoVersion",
    "value": {
        "value": "0.4.0",
        "timestamp": "2023-09-29T12:40:06.867818+00:00"
    }
}

Test Client> getValue Kuksa.Databroker.CommitSha   
{
    "path": "Kuksa.Databroker.CommitSha",
    "value": {
        "value": "f716273b74f7a2612307d0a810f3e581f35c893c",
        "timestamp": "2023-09-29T12:40:06.867820+00:00"
    }
}

Test Client> 

```

But maybe there are cases where `Kuksa.Databroker.GitVersion` actually return something

SebastianSchildt:
> But maybe there are cases where `Kuksa.Databroker.GitVersion` actually return something

Seems no in our current build

```
% docker run -it --rm   ghcr.io/eclipse/kuksa.val/databroker:0.4.0   -V                       
Kuksa Data Broker a1c9d5083033273b861771cd291ea3f4d7189df4
```

(I think the "-V" is conditional, it tries to get "GIT Version" when possible, but falls back to commit hash if not). So in general: Nothing really "critical" but probably something too look into during a rainy winter night)

PR #675, opened 2023-09-27
Rust client library and plugable databroker cli

lukasmittag:
This PR addresses the need of a client library for rust for the new API (kuksa.val.v1) and makes the databroker-cli plugable. TODO: Documentation update.

To try this PR you need to know that per default it will be built with the old API client lib. If you want to use the new one use cargo ```
run --bin databroker-cli --features feature_kuksa --no-default-features
```

### CODE REVIEW

**Comment by erikbosch** [comment]
How does Rust compiler handle features? Does it ignore features not included (like old fashioned `#ifdef`) or does it anyway compile them even if it does not include them in result. 

If it ignore it totally - would it then be an idea to extend our CI so that it also builds the "feature_kuksa" variant, to make sure that we do not introduce any regressions that can be detected at compile time, i.e. that all code is compiled by CI.

**Comment by lukasmittag** [comment]
> How does Rust compiler handle features? Does it ignore features not included (like old fashioned `#ifdef`) or does it anyway compile them even if it does not include them in result.
> 
> If it ignore it totally - would it then be an idea to extend our CI so that it also builds the "feature_kuksa" variant, to make sure that we do not introduce any regressions that can be detected at compile time, i.e. that all code is compiled by CI.

Yes that needs to be added. Good point!

**Comment by lukasmittag** [comment]
```
cargo clippy --all-targets -- -W warnings -D warnings
```
would this be enough or would you actually fully build it

File: kuksa_databroker/databroker-cli/src/cli.rs [hunk 1]
```diff
@@ -1,5 +1,5 @@
 /********************************************************************************
-* Copyright (c) 2022 Contributors to the Eclipse Foundation
+* Copyright (c) 2023 Contributors to the Eclipse Foundation
```
erikbosch, line 2: Why changing year? Eclipse seems to recommend either just stating the year the file was created, or a range "creation year - year of latest change", see https://www.eclipse.org/projects/handbook/#ip-copyright-headers

kuksa_databroker/README.md
```diff
@@ -118,6 +118,8 @@ Currently, to run databroker-cli (see below), you do need to change the port it
 
 ## Test the Databroker using CLI
 
+The databroker supports both of ```sdv.databroker.v1``` and ```kuksa.val.v1``` as an API. Per default the databroker-cli uses the ```sdv.databroker.v1``` interface. To change it use ```--protocol``` option when starting. Chosse eihter one of ```kuksa-val-v1``` and ```sdv-databroker-v1```.
```
erikbosch, line 121: spelling error "eihter"

kuksa_databroker/databroker-cli/src/help.rs
```diff
@@ -0,0 +1,1029 @@
...
+            //     (_, _) => {}
+            // }
+        }
+        None => return Err("No server uri specified".to_owned()),
+    }
+    parts.path_and_query = Some("".parse().expect("uri path should be empty string"));
+    tonic::transport::Uri::from_parts(parts).map_err(|err| format!("{err}"))
+}
+
+fn print_logo(version: impl fmt::Display) {
```
erikbosch, line 67: Is this identical to "print_logo" in cli.rs?

kuksa_databroker/lib/Cargo.toml
```diff
@@ -0,0 +1,37 @@
...
+# terms of the Apache License 2.0 which is available at
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# SPDX-License-Identifier: Apache-2.0
+#*******************************************************************************/
+
+[package]
+name = "common"
+version = "0.1.0"
+authors = ["Robert Bosch GmbH"]
```
erikbosch, line 17: As part of the release I replaced most instances of this line with "Eclipse KUKSA Project" after a discussion with Sebastian. Could be good to do that also for the new files

kuksa_databroker/databroker-cli/src/help.rs
lukasmittag: ah yes :) forgot to remove

Issue #674, opened 2023-09-26
Not all valid VSS paths supported by databroker


SebastianSchildt:
Consider the following VSS tree (you can test this by hacking PR in #673 or by loading a generated VSS.json as `--metadata` with such paths

```
[
    {
        "path": "Vehicle._kuksa.databroker.CommitSha",
        "metadata": {
            "data_type": "STRING",
            "entry_type": "ATTRIBUTE",
            "description": "Commit SHA of current version"
        }
    },
    {
        "path": "Vehicle._kuksa.databroker.CargoVersion",
        "metadata": {
            "data_type": "STRING",
            "entry_type": "ATTRIBUTE",
            "description": "Databroker version as reported by GIT"
        }
    },
    {
        "path": "Vehicle._kuksa.databroker.GitVersion",
        "metadata": {
            "data_type": "STRING",
            "entry_type": "ATTRIBUTE",
            "description": "Databroker version as reported by GIT"
        }
    }
]

```

one will notice that direct access, e.g. `getValue Vehicle._kuksa.databroker.GitVersion` or `getMetadata getValue Vehicle._kuksa.databroker.GitVersion` is not possible leading to an error

```
Test Client> getValue Vehicle._kuksa.databroker.GitVersion
{
    "error": {
        "code": 400,
        "reason": "bad_request",
        "message": "Bad Path Request"
    },
    "errors": [
        {
            "path": "Vehicle._kuksa.databroker.GitVersion",
            "error": {
                "code": 400,
                "reason": "bad_request",
                "message": "Bad Path Request"
            }
        }
    ]
}
```

While at the same time it is possible to get the data using wildcards

```
Test Client> getValue Vehicle.**
[
    {
        "path": "Vehicle._kuksa.databroker.CommitSha",
        "value": {
            "value": "3cca2ba197316b31180c9d597bfbbd8d405fe467",
            "timestamp": "2023-09-26T17:16:11.420578+00:00"
        }
    },
    {
        "path": "Vehicle._kuksa.databroker.CargoVersion",
        "value": {
            "value": "0.4.0",
            "timestamp": "2023-09-26T17:16:11.420569+00:00"
        }
    },
    {
        "path": "Vehicle._kuksa.databroker.GitVersion",
        "value": {
            "value": "N/A",
            "timestamp": "2023-09-26T17:16:11.420394+00:00"
        }
    }
]

Test Client> 
```

This is definitely not consistent, so: Bug.

[...]

PR #678, opened 2023-10-02
Support all valid vss paths

rafaeling:
Databroker now accepts any path composed of any character except : and whitespace, i.e.
```
- Vehicle._kuksa.databroker.GitVersion
- Vehicle.Do_you_not_like_smörgåstårta
- Vehicle.tschö_mit_ö
- Vehicle.東京_Москва_r#true
```

### DIFF
--- a/kuksa_databroker/databroker/src/broker.rs
+++ b/kuksa_databroker/databroker/src/broker.rs
@@ -29,6 +29,8 @@ use std::time::SystemTime;
 
 use tracing::{debug, info, warn};
 
+use crate::glob;
+
 #[derive(Debug)]
 pub enum UpdateError {
     NotFound,
@@ -46,7 +48,7 @@ pub enum ReadError {
     PermissionExpired,
 }
 
-#[derive(Debug)]
+#[derive(Debug, PartialEq)]
 pub enum RegistrationError {
     ValidationError,
     PermissionDenied,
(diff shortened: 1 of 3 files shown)


argerus (changes requested):
These changes still include validation of paths "sprinkled" throughout the code.

As discussed and concluded in the corresponding issue (#674), a better approach would be to add validation *only* at the point of adding new entries (i.e. [here](https://github.com/boschglobal/kuksa.val/blob/fed6a1d148efb5f944a92bc81bdd6eb22c914858/kuksa_databroker/databroker/src/broker.rs#L987)).

kuksa_databroker/databroker/src/broker.rs
```diff
@@ -1024,6 +1026,10 @@ impl<'a, 'b> DatabaseWriteAccess<'a, 'b> {
         allowed: Option<types::DataValue>,
         datapoint: Option<Datapoint>,
     ) -> Result<i32, RegistrationError> {
+        if !glob::is_valid_pattern(name.as_str()) {
```
argerus, line 1029: ### 1. Validate path
This is not correct, as what we want to validate is that a new entry has a valid **path**.

As currently implemented, it will allow `*` and `**` in the path (as `is_valid_pattern(...)` validates a **pattern**).

### 2. Validate pattern
And when `DatabaseWriteAccess::add` has been updated to accept only valid **paths**, we know that any entry in the database has a valid path.

Then, when later constructing "a matcher" that can be used to match paths with a pattern, `is_valid_pattern(...)` can be used to verify that the pattern is valid. But the pattern validation should **not** duplicate the validation of what a valid path can contain (as we already know that all entries we are matching against have a valid path). Instead, it should only validate that the pattern is valid in terms of wildcards, i.e. it should allow any character in a path (which is what this PR is about).

argerus (changes requested):

kuksa_databroker/databroker/src/broker.rs
rafaeling: is_valid_path method added to split different validations

argerus (approved):

PR #681, opened 2023-10-04
Use common action for dash

### LINKED ISSUE #677: Dash check runs in the "wrong place" in CI for Rust builds
For databroker(-cli) we

 - create an input file for Dash and a BOM (createbom)
 - Run the dash tool agains the input file and archive outcomes

This currently is done in the "Lint" step, which seems wrong (was likely done to "recylce" the builder").

See https://github.com/eclipse/kuksa.val/actions/runs/6328141013/job/17246390201 

Way forward seems
 - do not do the Createbom in the "lint" step but have something in buold or seperate
 - The DASH tool check (e.g. processing a Dash file and archiving the outputs) should be spun out in a separate action probably in the shared actions repository, as this is the same no matter we do it for pyton, rust, Kotlin etc....

erikbosch:
This PR does two things:

- It starts using a common action for dash in kuksa-actions. For now not using the "official" eclipse-kuksa repo as the action have not been merged there yet
- It puts the BOM generation and dash check in a separate job. Disadvantage is that that possibly causes the job to take somewhat longer as we may need to "redo" som setup steps like checkout. I put the job as follow up to build-container. It could run in parallel, but maybe we save some resource only running this job if build-container succeeds

Fixes #677


erikbosch:
Replaced by #680

PR #683, opened 2023-10-06
Simplify kuksa-client docker build and reducing resulting image size

### LINKED ISSUE #682: kuksa-client docker > 400MB on ARM
kuksa-client docker is around 480MiB on ARM, while the "normal" size on x64 is around 95MiB.

Root cause seems to be a fault in the GRPC build process that leaves an unstripped .so

see inside the container

```
/kuksa-client # du -hcs *
[...]
242.2M	grpc
150.3M	grpc_tools
```

There is an issue https://github.com/grpc/grpc/issues/29935

the easiest solution is probably just stripping manually when building on ARM

SebastianSchildt:
This fixes #682 

On  a high level

 - switch to a glibc build
 - this reduced build times as grpc wheels exist
 - as a side effect the existing glibc wheels are already small
 - using pyinstaller only ship the parts of python necessary to run the client, thus reducing size over using a python base image

This fixes the huge size of the arm image, while at the same time even reducing the amd64 image compared to master

Arm64 size now

```
ttl.sh/kuksa.val/kuksa-client-ff43b39d56fff605c6330c078dea4b383cade50f        latest                 e2dce0b872fd   About a minute ago   73.1MB
```

### DIFF
--- a/kuksa-client/Dockerfile
+++ b/kuksa-client/Dockerfile
@@ -8,23 +8,41 @@
 
 # Note: This dockerfile needs to be executed one level above in the root folder
 
-FROM python:3.10-alpine as build
-RUN apk update && apk add git alpine-sdk linux-headers
-RUN pip install --upgrade pip build
+
+FROM python:3.10-slim-bookworm as build
+# binutils is required by pyinstaller to strip any .so libs that are collected
+# git is used to determine & embed version information during build time
+RUN apt update && apt -yy install binutils git
+RUN pip install --upgrade pip build pyinstaller
+
 # We must copy the whole repository otherwise version lookup by tag would not work
 COPY . /kuksa.val/
 WORKDIR /kuksa.val/kuksa-client
-RUN rm -rf dist
+
 RUN python3 -m build
-RUN mkdir /kuksa-client
-RUN pip install --target /kuksa-client --no-cache-dir dist/*.whl
+# We install globally on build container, so pyinstaller can easily gather all files
+RUN pip install --no-cache-dir dist/*.whl
 
-FROM python:3.10-alpine
+WORKDIR /
+RUN rm -rf dist
+# Letting pyinstaller collect everything that is required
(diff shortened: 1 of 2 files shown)


kuksa-client/Dockerfile
```diff
@@ -8,23 +8,41 @@
 
 # Note: This dockerfile needs to be executed one level above in the root folder
 
-FROM python:3.10-alpine as build
-RUN apk update && apk add git alpine-sdk linux-headers
-RUN pip install --upgrade pip build
+
+FROM python:3.10-slim-bookworm as build
```
erikbosch, line 12: Should we possibly document our strategy for Docker environments somewhere. Like, should we stop using alpine everywhere? Today we have a quite big mix, and are there not some possible benefits of using the same environment as often as possible?

erikbosch (approved):
I did a very basic check:

- Download container
- Start it and verify that KUKSA client starts and connect successfully to Databroker (insecure mode)
That worked, that is good enough for me

kuksa-client/Dockerfile
SebastianSchildt: "probably", although I think it might differ depending on technology

So "historically" there was Alpine used, as for the C++ Val-Server it really made a difference size-wise.

For Python based things the savings are not that large, and using glibc builds are just easiest. Current stable debian just seems a common choice in many docker workloads. The solution here basically already includes learnings from the past , e.g. the pyinstaller technique from DBC feeder, except it does not condense all pyinstaller collected things in one file (as it does not offer benefits inside a container, but a runtime penalty unpacking it)

For rust it is a bit different, as we are using "from scratch" images (e.g. only  one binary in the container, no distro), and glibc isn't really suited to be statically linked  - you CAN do it, but then some rust crates/dependecies break that, hence we have MUSL there.

tldr; We should, I think this on represents our current "best practice" in pyhton, and maybe we should adapt our python stuff

SebastianSchildt:
I tested secure. lgtm.

PR #688, opened 2023-10-19
Subscribe to wildcard

rafaeling:
Output after setting Vehicle.PowerOptimizeLevel and Vehicle.IsMoving signals:

```
Test Client> subscribeMultiple Vehicle.*
{
    "subscriptionId": "f5c6544f-f93d-4ef8-9532-5f537eb5a179"
}

[
  {
    "entry": {
      "path": "Vehicle.IsMoving",
      "value": {
        "value": false,
        "timestamp": "2023-10-19T08:38:21.467527+00:00"
      }
    },
    "fields": [
      "VALUE"
    ]
  }
]

[
  {
    "entry": {
      "path": "Vehicle.PowerOptimizeLevel",
      "value": {
        "value": 5,
        "timestamp": "2023-10-19T08:38:53.230746+00:00"
      }
    },
    "fields": [
      "VALUE"
    ]
  }
]


```

### DIFF
--- a/kuksa_databroker/databroker/src/broker.rs
+++ b/kuksa_databroker/databroker/src/broker.rs
@@ -706,9 +706,13 @@ impl ChangeSubscription {
                         }
                         notifications
                     };
-                    match self.sender.send(notifications).await {
-                        Ok(()) => Ok(()),
-                        Err(_) => Err(NotificationError {}),
+                    if notifications.updates.is_empty() {
+                        Ok(())
+                    } else {
+                        match self.sender.send(notifications).await {
+                            Ok(()) => Ok(()),
+                            Err(_) => Err(NotificationError {}),
+                        }
                     }
                 } else {
                     Ok(())
--- a/kuksa_databroker/databroker/src/broker.rs
+++ b/kuksa_databroker/databroker/src/broker.rs
@@ -79,7 +79,7 @@ pub struct Entry {
     pub metadata: Metadata,
 }
 
-#[derive(Debug, PartialEq, Eq, Hash)]
+#[derive(Debug, PartialEq, Eq, Hash, Clone)]
 pub enum Field {
     Datapoint,
     ActuatorTarget,
@@ -1328,21 +1328,8 @@ impl<'a, 'b> AuthorizedAccess<'a, 'b> {
 
     pub async fn subscribe(
         &self,
-        entries: HashMap<String, HashSet<Field>>,
+        valid_entries: HashMap<i32, HashSet<Field>>,
     ) -> Result<impl Stream<Item = EntryUpdates>, SubscriptionError> {
-        let valid_entries = {
-            let mut valid_entries = HashMap::new();
-            for (path, fields) in entries {
-                match self.get_id_by_path(path.as_ref()).await {
-                    Some(id) => {
-                        valid_entries.insert(id, fields);
-                    }
-                    None => return Err(SubscriptionError::NotFound),
-                }
-            }
-            valid_entries
-        };
-
         if valid_entries.is_empty() {
             return Err(SubscriptionError::InvalidInput);
         }
@@ -2803,10 +2790,7 @@ mod tests {
             .expect("Register datapoint should succeed");
 
         let mut stream = broker
-            .subscribe(HashMap::from([(
-                "test.datapoint1".into(),
-                HashSet::from([Field::Datapoint]),
-            )]))
+            .subscribe(HashMap::from([(id1, HashSet::from([Field::Datapoint]))]))
             .await
             .expect("subscription should succeed");
 
--- a/kuksa_databroker/databroker/src/grpc/kuksa_val_v1/val.rs
+++ b/kuksa_databroker/databroker/src/grpc/kuksa_val_v1/val.rs
@@ -375,31 +375,61 @@ impl proto::val_server::Val for broker::DataBroker {
             ));
         }
 
-        let mut entries = HashMap::new();
+        let mut valid_requests: HashMap<String, (regex::Regex, HashSet<broker::Field>)> =
+            HashMap::new();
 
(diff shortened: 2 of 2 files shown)


argerus (changes requested):

kuksa_databroker/databroker/src/broker.rs
```diff
@@ -1333,11 +1332,16 @@ impl<'a, 'b> AuthorizedAccess<'a, 'b> {
...
+                let regex_exp = glob::to_regex(&path);
+                if let Ok(regex) = regex_exp {
+                    self.for_each_entry(|entry| {
+                        let entry_fields = fields.clone();
+                        let entry_path = &entry.metadata().path;
+                        if regex.is_match(entry_path) {
+                            valid_entries.insert(entry.metadata().id, entry_fields);
+                        }
+                    })
+                    .await;
```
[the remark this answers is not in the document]
rafaeling: Code updated with your approach

kuksa_databroker/databroker/src/grpc/kuksa_val_v1/val.rs
```diff
@@ -375,31 +375,60 @@ impl proto::val_server::Val for broker::DataBroker {
...
+                            match entries.get(id) {
+                                Some(fields) => {
+                                    let joined_set: HashSet<_> =
+                                        fields.union(&fields_set).cloned().collect();
+                                    entries.insert(*id, joined_set);
+                                }
+                                None => {
+                                    entries.insert(*id, fields_set.clone());
+                                }
+                            }
```
argerus, line 427: I think extend works better for this case. And perhaps using the "entry" API to do it...
```suggestion
                            entries
                                .entry(entry.metadata().id)
                                .and_modify(|existing_fields| {
                                    existing_fields.extend(fields.clone());
                                })
                                .or_insert(fields.clone());
```

kuksa_databroker/databroker/src/broker.rs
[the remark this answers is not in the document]
argerus: Looks better.

It still doesn't handle the case when trying to subscribe to a non existing path.
Since everything is treated as a wildcard pattern, I guess it would have to be solved by checking if "wildcard pattern matches nothing" -> error "not found".

kuksa_databroker/databroker/src/grpc/kuksa_val_v1/val.rs
```diff
@@ -375,31 +375,55 @@ impl proto::val_server::Val for broker::DataBroker {
...
+
+        let mut entries: HashMap<i32, HashSet<broker::Field>> = HashMap::new();
+
+        if !valid_requests.is_empty() {
+            for (_, (regex, fields_set)) in valid_requests {
+                broker
+                    .for_each_entry(|entry| {
+                        let entry_path = &entry.metadata().path;
+                        if regex.is_match(entry_path) {
+                            entries
```
argerus, line 417: > Looks better.
> 
> It still doesn't handle the case when trying to subscribe to a non existing path.
> Since everything is treated as a wildcard pattern, I guess it would have to be solved by checking if "wildcard pattern matches nothing" -> error "not found".

Something along these lines.
```diff
+           let mut requested_path_found = false;
            for (_, (regex, fields_set)) in valid_requests {
                broker
                    .for_each_entry(|entry| {
                        let entry_path = &entry.metadata().path;
                        if regex.is_match(entry_path) {
+                           requested_path_found = true;
                            entries
            ...
+           if !requested_path_found {
+               // Return error "Not found"
+           }
```

argerus (changes requested):

kuksa_databroker/databroker/src/grpc/kuksa_val_v1/val.rs
rafaeling: it should be include inside the for loop I suppose, right? We want to return an error in case any path dont have any entry

SebastianSchildt (changes requested):
get and subscribe are inconsistent, if get is missing access to any datapoint it fails, while subscribe doesn't

```
Test Client> getValue Vehicle
{
    "error": {
        "code": 403,
        "reason": "forbidden",
        "message": "Permission denied for some entries"
    },
    "errors": [
        {
            "path": "Vehicle",
            "error": {
                "code": 403,
                "reason": "forbidden",
                "message": "Permission denied for some entries"
            }
        }
    ]
}

Test Client> subscribe Vehicle
{
    "subscriptionId": "7cdc6447-6bbd-4693-8e32-76bce9976d5c"
}
```

I think wildcard subscribe should only work of all datapoints can be read and compeltely fail otherwise. If there are reasons, why this is not the case, the behaviour still is not ideal. When I did the subscribe with a token that only allows to read `Vehicle.Speed` , setting this value works, however setting any other value, we are not allwoed to read, still triggers a response on the subscribe channel. That is defintely wrong, see how it looks in Client, when after subscribe setting `Vehicle.Speed`, and then some values in `Vehicle.CurrentLocation`


Doing this

```
"Authenticated"

Test Client> setValue Vehicle.Speed 22
OK

Test Client> setValue Vehicle.CurrentLocation.Latitude 27
OK

Test Client> setValue Vehicle.CurrentLocation.Latitude 24
OK

```

yields 

```

Test Client> subscribe Vehicle
{
    "subscriptionId": "7cdc6447-6bbd-4693-8e32-76bce9976d5c"
}

[
  {
    "entry": {
      "path": "Vehicle.Speed",
      "value": {
        "value": 22.0,
        "timestamp": "2023-10-24T06:42:16.499781+00:00"
      }
    },
    "fields": [
      "VALUE"
    ]
  }
]

[]

[]

Test Client> 

```

SebastianSchildt (approved):
Nice. Behaves as expected. Checked wildcards and vanilla subscriptions and handling authorisation.

PR #689, opened 2023-10-20
Remove `optional` from kuksa.val.v1 proto

argerus:
Remove `optional` from `kuksa.val.v1` as it's causing unnecessary issues building it on various platforms depending on which version of protoc is available.

It should have minimal effects downstream, as the python library API remains backwards compatible and the few clients using their own generated files can be updated with minimal effort.

### DIFF
--- a/kuksa-client/kuksa_client/grpc/__init__.py
+++ b/kuksa-client/kuksa_client/grpc/__init__.py
@@ -161,7 +161,7 @@ class Metadata:
         metadata = cls(data_type=DataType(message.data_type),
                        entry_type=EntryType(message.entry_type))
         for field in ('description', 'comment', 'deprecation', 'unit'):
-            if message.HasField(field):
+            if getattr(message, field):
                 setattr(metadata, field, getattr(message, field))
         if message.HasField('value_restriction'):
             value_restriction = getattr(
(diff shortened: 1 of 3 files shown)


kuksa-client/kuksa_client/grpc/__init__.py
```diff
@@ -161,7 +161,7 @@ def from_message(cls, message: types_pb2.Metadata):
         metadata = cls(data_type=DataType(message.data_type),
                        entry_type=EntryType(message.entry_type))
         for field in ('description', 'comment', 'deprecation', 'unit'):
-            if message.HasField(field):
+            if getattr(message, field):
```
SebastianSchildt, line 164: What is this change doing, does this change hint at some incompatibility, or is it just a "style" change

argerus: It is the result of the changed protocol.

The alternative would be to do a wire-protocol compatible change by introducing `oneof`s. I'm not sure it's better in the long run though, as using it would be clunkier. But it would mean existing users could freely move between versions.

argerus: I think the alternative PR that remains wire compatible is preferred at this point.

The usage of `optional` (or the wire compatible alternative) isn't really needed _at all_ if the API is structured just a little bit differently. But that would break backwards compatibility (once), and we can defer any such change until later.

PR #692, opened 2023-10-25
[databroker] Update calls to subscribe

argerus:
Internal subscribe function signature was modified and not all call sites got updated.


erikbosch (approved):
Looks good, all builds pass, so will merge it

PR #694, opened 2023-10-26
[databroker] Update dependencies to support RISC-V

argerus:


erikbosch:
Cargo.lock needs to be regenerated as my Pr also touched cargo.lock

SebastianSchildt:
It builds for me, I was so far not b able to run it on milkv. But issue may be that milkv has an musl based "buildroot" system image, and the cross tooling could only build dynamically linked glibc version. I did copy all the required .so, but then I got this error

```
RUST_BACKTRACE=1 ./databroker
thread 'main' panicked at kuksa_databroker/databroker/src/[main.rs:446](http://main.rs:446/):7:
Failed building the Runtime: Os { code: 38, kind: Unsupported, message: "Function not implemented" }
stack backtrace:
note: Some details are omitted, run with RUST_BACKTRACE=full for a verbose backtrace.
[root@milkv-duo]~#
```

(that is basically the last line of main.rs, so something already fails at early strtup I would say). On that platform (small root fs, less than 30MiB avaialable RAM :-P ) I could not run the debug build.


Anyway, I feel this might be more related to this special platform than the build. Will try on a Vision Five 2 once I find the time (that is far more powerful, and also has a more vanilla debian/ubuntu OS)

SebastianSchildt:
Did test on a VisionFive 2 (https://www.starfivetech.com/en/site/boards with some debian-ish OS), and it works. Started databroker on the target, and could connect dbcfeeder and client with subscription form another host. Seems to work. As this does update quite a bit of dependencies, let's merge this after next minor release is out. In case the dependecies updates breaks some minor things somewhere, we'll have some time to notice.

PR #700, opened 2023-10-31
Merge V0.4.1 back to master

erikbosch:
For this PR I suggest to merge with a merge commit


SebastianSchildt (approved):

SebastianSchildt:
@erikbosch Plain merge seems we have disabled

image

Rebase or try the squash?

erikbosch:
@SebastianSchildt  - I assume it is some form of branch protection. We can rebase or squash, th only drawback is that it not will be as visible that the release branch has been "merged" back compared to if we would have a real merge commit.

Issue #666, opened 2023-09-22
RISC V support


SebastianSchildt:
As this question pops up pretty regularly these days... Create this issue for tracking

Currently RISC V is not supported, the main culprit are the required RUST crypto libs for TLS and the JWT tokens that rely on the "ring" base library https://github.com/briansmith/ring

There have been issues and PRs raised for almost two years, but while the Git is sort of active the whole project seems to be semi-abandoned and has not had a release in a long time.

The most recent developments seem to be in this PR, 

https://github.com/briansmith/ring/pull/1627

so if RISC V support is crucial for you, try to move things over there.

argerus:
Maybe this newly [announced funding](https://briansmith.org/lets-build-a-fips-rust-crypto) will result in a new release supporting more targets (including RISC-V).

> In parallel with the FIPS-related work, I will release the 0.17 version of *ring* that supports a much wider set of targets than *ring* has supported up to now.

SebastianSchildt:
Nice, somebody found money for that guy. Lets hope 🙏  something comes out, my RISC board is getting bored without KUKSA and a second one is on the way already (https://milkv.io/duo 😄  ).

SebastianSchildt:
Just tracking some progress: 
Meanwhile https://github.com/briansmith/ring/pull/1627 is merged, however just doing

```
[patch.crates-io]
ring = { git = 'https://github.com/briansmith/ring.git', branch='main' }
```

is not enough, as this is currently tagged version 0.17 so 0.16.20 is still being pulled from rustls and jsonwebtoken.

As a "hack" I tried clonign ring, changing Cargo.toml claiming main to be version 0.16.22  and refernecing it like

```
ring = { path = '/path/to/patched-ring' }
```

However, "unfortunately" they use their version numbers correctly, i.e. the current ring library is not compatible at least with rustls (did not dig deeper).

So it looks like
 - still waiting for an "offiical" ring 0.17.0
 - Wiating for other prjects to pick it up

SebastianSchildt:
Apparently rustls is already work in progress: https://github.com/rustls/rustls/pull/1508

SebastianSchildt:
So we might wanna try -aftter 0.4.1 release - upgrading some dependecnies.

Meanwhile we have the required ring  0.17 released

https://crates.io/crates/ring

And rusttls
https://crates.io/crates/rustls/0.21.8

as well as jsonwebtoken
https://crates.io/crates/jsonwebtoken/9.1.0

[...]

PR #703, opened 2023-11-05
Build and test RISCV

SebastianSchildt:
This builds and pushes RISCV containers as well. One step further towards full #666 

I extended the integration test, so that all containers are tested (before only the one similar to the runner architecture)

As the riscv builds need to based on glibc for the moment, there was no easy way to adapt the general docker file, so I switched to building with cross and having the docker just "package" the binaries. On the plus side
 - this enables to run on standard Github runners as well
 - Does not need the hacky (ingenious?) copy-from-oci action anymore

Once this is in, I'd give the rust-cli the same treatment

### DIFF
--- a/.github/workflows/kuksa_databroker_build.yml
+++ b/.github/workflows/kuksa_databroker_build.yml
@@ -89,16 +89,21 @@ jobs:
     uses: eclipse-kuksa/kuksa-actions/.github/workflows/check_ghcr_push.yml@2
     secrets: inherit
 
-# Run on selfhosted, because our runner has native ARM build in a remote
-# builder (no need for qemu)
   build-container:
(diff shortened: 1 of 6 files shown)


.github/workflows/kuksa_databroker_build.yml
```diff
@@ -116,12 +118,12 @@ jobs:
           type=semver,pattern={{major}}
 
     # only needed for runners without buildx setup, will be slow
```
erikbosch, line 120: This line can be removed if we use this alt by default, right?

erikbosch (approved):

erikbosch:
Looks good to me. I did a very simple smoke test just starting the tll.sh container on amd64. I have not tested the risc part.

PR #704, opened 2023-11-08
RISCV builds for databroker-cli

SebastianSchildt:
uses the same pattern as datbaroker to build CLI. As a result it als emits RISCV binaries (glibc ones due to limitations in the ecosystem currently)

Hint: On a linux system with qemu-user-static (or similar) package installed the system can run/test any of the ocntainers, e.g. 

 docker run --rm -it --net=host --platform linux/riscv64 ttl.sh/kuksa.val/kuksa-databroker-cli-1924d6d3a814e19efd84ec1cd210dabc6fc6a387

Currently supported platforms are `linux/amd64` , `linux/arm64`, `linux/riscv64`

### DIFF
--- a/kuksa_databroker/Dockerfile-cli
+++ b/kuksa_databroker/Dockerfile-cli
@@ -12,51 +12,45 @@
 # ********************************************************************************/
 
 # This is expected to be executed in the kuksa.val top-level directory
+# You need to run build-all-targets-cli.sh first, as this docker file jsut
+# collects the artifacts
 
-# This builds the databroker-cli
-
-FROM ghcr.io/rust-cross/rust-musl-cross:x86_64-musl AS builder-amd64
+# AMD is a statically linked MUSL build
+FROM scratch AS target-amd64
 ENV BUILDTARGET="x86_64-unknown-linux-musl"
+COPY ./target/x86_64-unknown-linux-musl/release/databroker-cli /app/databroker-cli
 
 
-FROM ghcr.io/rust-cross/rust-musl-cross:aarch64-musl AS builder-arm64
+# ARM64 is a statically linked GRPC build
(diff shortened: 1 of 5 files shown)


kuksa_databroker/Dockerfile-cli
```diff
@@ -12,51 +12,45 @@
...
-# This will speed up fetching the crate.io index in the future, see
-# https://blog.rust-lang.org/2022/06/22/sparse-registry-testing.html
-ENV CARGO_UNSTABLE_SPARSE_REGISTRY=true
 
-RUN echo "Building for $TARGETARCH"
-ENV SRC="/home/rust/src"
-WORKDIR ${SRC}
+# RISCV is a glibc build. Rust toolchain not supported for MUSL
+# Distroless has not RISCV support yet, using debian for now
+#FROM gcr.io/distroless/base-debian12:debug as target-riscv64
```
erikbosch, line 32: Any reason to keep this one?

kuksa_databroker/build-all-targets-cli.sh
```diff
@@ -0,0 +1,75 @@
+#!/bin/bash
```
erikbosch, line 1: What is our view for *.sh files. Shall we have copyright+license header? Currently not required by our SPDX checker

SebastianSchildt: Hmm maybe I'll add at least a SPDX-header to make scanners happy. I am always a bit torn, normal these things are not very magic to warrant "protection", but likely putting license explicitly is better

erikbosch (approved):
Looks good to me, some minor comments filed.

SebastianSchildt:
Part of #666

PR #708, opened 2023-11-13
Add authorization integration test

lukasmittag:
This adds integration tests for authorizing against databroker and checking for the right permissions. 
#675 needs to be merged first because it is based on the new client lib. Therefore also some changes in the existing integration test took place.


kuksa_databroker/databroker/tests/world/mod.rs
```diff
@@ -194,12 +195,21 @@ impl DataBrokerWorld {
                 state.address = Some(addr);
             }
 
+            let mut _authorization = databroker::authorization::Authorization::Disabled;
+
+            if authorization_enabled{
+                match databroker::authorization::Authorization::new("-----BEGIN PUBLIC KEY----- MIICIjANBgkqhkiG9w0BAQEFAAOCAg8AMIICCgKCAgEA6ScE9EKXEWVyYhzfhfvg+LC8NseiuEjfrdFx3HKkb31bRw/SeS0Rye0KDP7uzffwreKf6wWYGxVUPYmyKC7jPji5MpDBGM9r3pIZSvPUFdpTE5TiRHFBxWbqPSYt954BTLq4rMu/W+oq5PdfnugbvoYpLf0dclBl1g9KyszkDnItz3TYbWhGMbsUSfyeSPzH0IADzLoifxbc5mgiR73NCA/4yNSpfLoqWgQ2vdTM1182sMSmxfqSgMzIMUX/tiaXGdkoKITF1sULlLyWfTo979XRZ0hmUwvfzr3OjMZNoClpYSVbKY+vtxHyux9KOOtv9lPMsgYIaPXvisrsneDZfCS0afOfjgR96uHIe2UPSGAXru3yGziqEfpRZoxsgXaOe905ordLD5bSX14xkN7NCz7rxDLlxPQyxp4Vhog7p/QeUyydBpZjq2bAE5GAJtiu+XGvG8RypzJFKFQwMNswg1BoZVD0mb0MtU8KQmHcZIfY0FVer/CR0mUjfl1rHbtoJB+RY03lQvYNAD04ibAGNI1RhlTziu35Xo6NDEgs9hVs9k3WrtF+ZUxhivWmP2VXhWruRakVkC1NzKGh54e5/KlluFbBNpWgvWZqzWo9Jr7/fzHtR0Q0IZwkxh+Vd/bUZya1uLKqP+sTcc+aTHbnAEiqOjPq0D6X45wCzIwjILUCAwEAAQ== -----END PUBLIC KEY-----".to_string()){
```
erikbosch, line 201: We have some hard-coded token and certificates/keys in this PR. Should we possibly include a comment on where the string comes from? Like where does this one come from or how has it been created?

lukasmittag: Yes, another option would be to give the exact files. But then when we split out certificates etc it would be a bit ugly. So I opted to hardcode them in. But yeah comments where they come from would not hurt.

erikbosch: I nothing else that makes it easier the day they expire and the tests by that reason fails.

erikbosch (approved):
Looks good to me

kuksa_databroker/databroker/tests/features/read_write_values.feature
```diff
@@ -1,46 +1,75 @@
...
+
+    Scenario: Write the current value of a Data Entry without right permissions fails
+    """
+      The token comes from kuksa.val/jwt/read-vehicle-speed.token. Expire date is Thursday, 1 January 2026
+    """
+      When a client sets the current value of Vehicle.Speed of type float to 13.4 with token eyJ0eXAiOiJKV1QiLCJhbGciOiJSUzI1NiJ9.eyJzdWIiOiJsb2NhbCBkZXYiLCJpc3MiOiJjcmVhdGVUb2tlbi5weSIsImF1ZCI6WyJrdWtzYS52YWwiXSwiaWF0IjoxNTE2MjM5MDIyLCJleHAiOjE3NjcyMjU1OTksInNjb3BlIjoicmVhZDpWZWhpY2xlLlNwZWVkIn0.QyaKO-vjjzeimAAyUMjlM_jAvhlmYVguzXp76jAnW9UUiWowrXeYTa_ERVzZqxz21txq1QQmee8WtCO978w95irSRugTmlydWNjS6xPGAOCan6TEXBryrmR5FSmm6fPNrgLPhFEfcqFIQm07B286JTU98s-s2yeb6bJRpm7gOzhH5klCLxBPFYNncwdqqaT6aQD31xOcq4evooPOoV5KZFKI9WKkzOclDvto6TCLYIgnxu9Oey_IqyRAkDHybeFB7LR-1XexweRjWKleXGijpNweu6ATsbPakeVIfJ6k3IN-oKNvP6nawtBg7GeSEFYNksUUCVrIg8b0_tZi3c3E114MvdiDe7iWfRd5SDjO1f8nKiqDYy9P-hwejHntsaXLZbWSs_GNbWmi6J6pwgdM4XI3x4vx-0Otsb4zZ0tOmXjIL_tRsKA5dIlSBYEpIZq6jSXgQLN2_Ame0i--gGt4jTg1sYJHqE56BKc74HqkcvrQAZrZcoeqiKkwKcy6ofRpIQ57ghooZLkJ8BLWEV_zJgSffHBX140EEnMg1z8GAhrsbGvJ29TmXGmYyLrAhyTj_L6aNCZ1XEkbK0Yy5pco9sFGRm9nKTeFcNICogPuzbHg4xsGX_SZgUmla8Acw21AAXwuFY60_aFDUlCKZh93Z2SAruz1gfNIL43I0L8-wfw
```
sophokles73, line 22: instead of specifying the (signed?) token in the feature, I would suggest to specify the authorities that the token should include and then create a corresponding token in the binding. That way we do not need to update all of the features if we decide to change the token format and/or the key used to sign the tokens.

kuksa_databroker/databroker/tests/features/read_write_values.feature
```diff
@@ -1,46 +1,75 @@
...
+
   Rule: Accessing unregistered Data Entries fails
 
     Background:
-      Given a running Databroker server
+      Given a running Databroker server with authorization false
 
     Scenario: Setting the current value of an unregistered Data Entry fails
-      When a client sets the current value of No.Such.Path of type float to 13.4
+      When a client sets the current value of No.Such.Path of type float to 13.4 with token ""
```
sophokles73, line 38: How about introducing a separate `When` expression for adding a token, e.g.
```
When a client sets the current value of No.Such.Path of type float to 13.4
And uses a token with authorization Set Vehicle.Speed
Then ...
```

We could then store the token in the World state and later use the token when invoking the Databroker API.

This way we wouldn't need to change all existing scenarios and could reuse the step in other scenarios as well

kuksa_databroker/databroker/tests/features/read_write_values.feature
lukasmittag: Seems like a reasonable adoption. Will check how this would look like.

erikbosch:
I looked at the Dash output and saw this.
Do we need to do anything?

```

[main] INFO Querying Eclipse Foundation for license data for 208 items.
[main] INFO Found 46 items.
[main] INFO Querying ClearlyDefined for license data for 162 items.
[main] INFO Found 162 items.
[main] INFO License information could not be automatically verified for the following content:
[main] INFO 
[main] INFO crate/cratesio/-/kuksa/0.1.0
[main] INFO 
[main] INFO This content is either not correctly mapped by the system, or requires review.
```

sophokles73:
> Do we need to do anything?

The Dash tool tries to find licensing information for dependencies on Clearly Defined and the Eclipse Foundation systems. The `kuksa` crate seems to not have been analyzed yet, neither by CD nor the EF.

What you could do is run the Dash tool with the [`--review` switch](https://github.com/eclipse/dash-licenses#automatic-ip-team-review-requests) to automatically create a corresponding review request with the EF. You would then need to provide the source location manually in the issue created in the EF GitLab instance for the review.

However, before you do so, you might want to think about
* whether `kuksa` is the right name for the crate. IMHO it is quite abstract and can mean anything or nothing.
* whether `kuksa` is actually a crate that is helpful to others as well and therefore should be published to crates.io

kuksa_databroker/databroker/tests/read_write_values.rs
```diff
@@ -75,9 +74,10 @@ fn get_data_entries_from_table(
     data_entries
 }
 
-#[given(regex = "^a running Databroker server.*$")]
-async fn start_databroker_server(w: &mut DataBrokerWorld, step: &Step) {
-    w.start_databroker(get_data_entries_from_table(step)).await;
+#[given(regex = "^a running Databroker server with authorization (true|false).*$")]
```
sophokles73, line 77: can we use `(enabled|disabled)` instead? That would make the scenario descriptions more *readable* FMPOV ...

kuksa_databroker/databroker/tests/world/mod.rs
```diff
@@ -284,34 +316,85 @@ impl DataBrokerWorld {
...
+                    panic!("No function error {:?} should occur", e)
+                }
+                ClientError::Status(status) => {
+                    panic!("No status error {:?} should occur", status)
+                }
+            }
         }
     }
+
+    pub fn create_token(&self, _scope: String) -> String {
```
sophokles73, line 374: FMPOV we should remove the underscore from the `_scope` parameter ...

kuksa_databroker/databroker/tests/read_write_values.rs
```diff
@@ -90,7 +90,22 @@ async fn a_known_data_entry_has_value(
     value: String,
 ) {
     set_value(w, value_type, path, data_type, value).await;
-    w.assert_set_response_has_succeeded()
+    w.assert_set_succeeded()
+}
+
+#[when(expr = "a client uses a token with auhtorization {word}")]
```
sophokles73, line 96: in the Token struct, what is called *authorization* here, is called *scope* instead. Using common language across the two places will make it easier to read/understand.

kuksa_databroker/databroker/tests/read_write_values.rs
```diff
@@ -244,6 +244,24 @@ fn assert_request_failure(w: &mut DataBrokerWorld, expected_status_code: i32) {
...
+            ClientError::Connection(e) => {
+                panic!("No connection error {:?} should occcur", e)
+            }
+            ClientError::Function(e) => panic!("No function error {:?} should occur", e),
+            ClientError::Status(status) => assert_eq!(status.code(), Code::Unauthenticated),
+        }
+    }
+}
+
+#[then(expr = "the set operation succeeds without an error")]
```
sophokles73, line 260: can it also succeed *with an error*?

lukasmittag: No

kuksa_databroker/databroker/tests/read_write_values.rs
```diff
@@ -244,6 +244,24 @@ fn assert_request_failure(w: &mut DataBrokerWorld, expected_status_code: i32) {
     w.assert_status_has_code(expected_status_code)
 }
 
+#[then(expr = "the current value for {word} can not be accessed because we are unauthorized")]
```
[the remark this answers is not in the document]
lukasmittag: No word is not used as of now. Would it be okay to use it like this:
```
the request fails because the client is unauthorized
```
I agree that your option would be good but since the error would be a grpc error code I think it is nicer to name it (do not know if anybody is familiar with grpc codes. For the standard error codes used in the other examples I think everyone knows what they mean.) Otherwise yeah we could do it like you suggested. What do you think?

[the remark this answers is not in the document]
lukasmittag: Ah, but I see now there is a scenario from you which checks for status codes as well and you linked documentation to it in code. Then I could use this. But as I mentioned I think for clarity maybe it would be best to name it in theses cases. Let me know what your thoughts are :)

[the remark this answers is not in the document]
sophokles73: FMPOV it is ok to refer to the gRPC codes since we are testing a gRPC API here ...

kuksa_databroker/databroker/tests/read_write_values.rs
sophokles73: I guess you get my point? ;-)

sophokles73 (approved):
LGTM

SebastianSchildt (approved):

PR #710, opened 2023-11-16
Remove kuksa-client

erikbosch:
As kuksa-client has been moved to https://github.com/eclipse-kuksa/kuksa-python-sdk it should better be removed from old location, to avoid that pull-requests and issues are created in this repo. 

Possible regressions:

1. Tests relying on that kuksa-client exist in master branch of this repo will fail. That is good, then they can start using the other repo
2. Tests relying on using kuksa-client docker contain from this repo will continue to work, but they will no longer test against "latest" released/unreleased version. If we consider it important to make sure that we everyone migrates we could upload a "broken" docker container, but that is hopefully not needed


lukasmittag (approved):
LGTM

SebastianSchildt:
This works for me, I also meanwhile tagged a "last 0.4.2" that already pushed a container printing a waring message (I did opt for that instead of pushing an "evil" not working container.

So we could move forward removing this in master, but anyway, it does need tiny rebase first :-D

erikbosch:
@SebastianSchildt - It is rebased now

SebastianSchildt (approved):
Looking good. Also good to keep a "stub Readme".

Will merge. Let's see what breaks 🍿

PR #712, opened 2023-11-20
kuksa-client move: Add notice about new location to get updated versions

SebastianSchildt:
This is intended to build for master as well as next minor release. It is fully functional, but will warn people that they might update their source location to get updated versions.

PyPi users will never see this (as we will not publish it there), but I added PyPi location as well as "reassurance", i.e. if somebody is using kuksa-client pypi package and also the container

Should look like this

image

### DIFF
--- a/kuksa-client/kuksa_client/__main__.py
+++ b/kuksa-client/kuksa_client/__main__.py
@@ -675,6 +675,12 @@ using {'KUKSA GRPC' if config['protocol'] == 'grpc' else 'VISS' } protocol."
 
 
 def main():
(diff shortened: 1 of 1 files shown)


erikbosch:
@SebastianSchildt  - I have found a likely root cause for the failed pre-commit test.

- Ubuntu-latest seems to have been updated to use something with Python 3.12
- There seems to have been a bug with multi-line f-strings in flake 8 for 3.12 , see https://github.com/PyCQA/flake8/pull/1849
- Updating to a new version seems to make the trick, see #713 


You can either use the changed commit from there and close this one, or integrated the change from there

SebastianSchildt:
Closing in favor of #713

PR #722, opened 2023-12-19
Issue 719 - Renamed "databroker-cli" crate to "kuksa-cli"

tasnuva1:
Note:
- kuksa.val/Cargo.lock file has not been updated. Let me know, if this is a problem.
- git workflow kuksa_databroker-cli_build.yml file should be checked twice.


.github/workflows/kuksa_databroker-cli_build.yml
```diff
@@ -124,31 +124,31 @@ jobs:
...
       with:
-        name: databroker-cli-arm64
+        name: kuksa-cli-arm64
         path: ${{github.workspace}}/dist/arm64
 
     - name: "Archiving RISCV64 artifacts"
       uses: actions/upload-artifact@v3
       with:
-        name: databroker--cli-riscv64
+        name: kuksa--cli-riscv64
```
erikbosch, line 151: I notice that we since before has `--cli` here while other architectures have `-cli`. Maybe worth aligning when we anyway change name?

PR #728, opened 2024-01-23
Fix new Clippy issues

erikbosch:
Newer clippy does not seem to like `get(0)` but prefer `first()`

(clippy 0.1.75 complained, but not clippy 0.1.73)

This shall fix the failing build on master.


erikbosch:
FYI @rafaeling

SebastianSchildt (approved):
lgtm 🍨

PR #731, opened 2024-01-29
Remove self-hosted dependency for KUKSA Server

erikbosch:
Recent builds on master using self-hosted failed (see https://github.com/eclipse/kuksa.val/actions/workflows/kuksa_val_docker.yml)

With this change it works - but it takes a lot longer time (building grpc, boost, and so on for arm64) - so alternative solutions are welcome. But better than failing builds


lukasmittag (approved):
LGTM

PR #734, opened 2024-02-08
Add continuous subscription to subscribe method at a specific frequency

rafaeling:
This PR solves https://github.com/eclipse/kuksa.val/issues/652

**How can be tested?**

- Use this test.json file as the VSS input for the broker. It contains the necessary overlays modified to test it with the _ChangeType_ values _Continuous_ and _OnChange_ to verify the expected behavior
[test.json](https://github.com/eclipse/kuksa.val/files/14220629/test.json)

- You can subscribe with databroker-clie to the signals Vehicle.Driver.HeartRate (Continuous update) and Vehicle.Body.Trunk.Front.IsLightOn (OnChange update).

- Kuksa-client seems to have some bugs that need testing for this new feature.

**What to expect?**

- The value for Vehicle.Driver.HeartRate should be set, as well as for Vehicle.Body.Trunk.Front.IsLightOn.
- Subscribe to both paths, and the terminal should display the _Continuous_ values every second and _OnChange_ values only when updated.

**Kuksa client to test it**
https://github.com/eclipse-kuksa/kuksa-python-sdk/pull/21

### DIFF
--- a/kuksa_databroker/databroker/src/broker.rs
+++ b/kuksa_databroker/databroker/src/broker.rs
@@ -1592,7 +1592,7 @@ impl<'a, 'b> AuthorizedAccess<'a, 'b> {
 
     pub async fn subscribe(
         &self,
-        valid_entries: HashMap<i32, (HashSet<Field>, types::ChangeType)>,
+        valid_entries: HashMap<i32, HashSet<Field>>,
     ) -> Result<impl Stream<Item = EntryUpdates>, SubscriptionError> {
         if valid_entries.is_empty() {
             return Err(SubscriptionError::InvalidInput);
@@ -1601,21 +1601,36 @@ impl<'a, 'b> AuthorizedAccess<'a, 'b> {
         let mut entries_on_changed: HashMap<i32, HashSet<Field>> = HashMap::new();
         let mut entries_continuous: HashMap<i32, HashSet<Field>> = HashMap::new();
 
-        for (id, (fields, change_type)) in valid_entries {
-            match change_type {
-                types::ChangeType::OnChange => {
-                    entries_on_changed
-                        .entry(id)
-                        .and_modify(|existing_fields| existing_fields.extend(fields.clone()))
-                        .or_insert(fields.clone());
+        let db_read = self.broker.database.read().await;
+        let db_read_access = db_read.authorized_read_access(self.permissions);
+
+        for (id, fields) in valid_entries {
+            match db_read_access.get_entry_by_id(id) {
+                Ok(entry) => {
+                    let change_type = entry.metadata.change_type.clone();
+                    match change_type {
+                        types::ChangeType::OnChange => {
+                            entries_on_changed
+                                .entry(id)
+                                .and_modify(|existing_fields| {
+                                    existing_fields.extend(fields.clone())
+                                })
+                                .or_insert(fields.clone());
+                        }
+                        types::ChangeType::Continuous => {
+                            entries_continuous
+                                .entry(id)
+                                .and_modify(|existing_fields| {
+                                    existing_fields.extend(fields.clone())
+                                })
+                                .or_insert(fields.clone());
+                        }
+                        types::ChangeType::Static => {}
+                    }
                 }
-                types::ChangeType::Continuous => {
-                    entries_continuous
-                        .entry(id)
-                        .and_modify(|existing_fields| existing_fields.extend(fields.clone()))
-                        .or_insert(fields.clone());
+                Err(_) => {
+                    debug!("notify: could not find entry with id {}", id)
                 }
-                types::ChangeType::Static => {}
             }
         }
 
@@ -3145,19 +3160,8 @@ mod tests {
             .await
             .expect("Register datapoint should succeed");
 
-        let my_hashmap: HashMap<i32, (HashSet<Field>, types::ChangeType)> = [(
-            id1,
-            (
-                HashSet::from([Field::Datapoint]),
-                types::ChangeType::OnChange,
-            ),
-        )]
-        .iter()
-        .cloned()
-        .collect();
-
         let mut stream = broker
-            .subscribe(my_hashmap)
+            .subscribe(HashMap::from([(id1, HashSet::from([Field::Datapoint]))]))
             .await
             .expect("subscription should succeed");
 
--- a/kuksa_databroker/databroker/src/broker.rs
+++ b/kuksa_databroker/databroker/src/broker.rs
@@ -25,7 +25,8 @@ use std::collections::{HashMap, HashSet};
 use std::convert::TryFrom;
 use std::sync::atomic::{AtomicI32, Ordering};
 use std::sync::Arc;
-use std::time::SystemTime;
+use std::sync::Mutex;
+use std::time::{Duration, SystemTime};
 
 use crate::query::{CompiledQuery, ExecutionInput};
 use crate::types::ExecutionInputImplData;
@@ -90,7 +91,7 @@ pub enum Field {
     MetadataUnit,
 }
 
-#[derive(Default)]
+#[derive(Default, Debug)]
 pub struct Database {
     next_id: AtomicI32,
     path_to_id: HashMap<String, i32>,
@@ -101,6 +102,7 @@ pub struct Database {
 pub struct Subscriptions {
     query_subscriptions: Vec<QuerySubscription>,
     change_subscriptions: Vec<ChangeSubscription>,
+    continuous_subscriptions: Arc<Mutex<Vec<ContinuousSubscription>>>,
 }
 
 #[derive(Debug, Clone)]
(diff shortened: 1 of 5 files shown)


kuksa_databroker/databroker/src/broker.rs
```diff
@@ -1340,32 +1487,78 @@ impl<'a, 'b> AuthorizedAccess<'a, 'b> {
 
     pub async fn subscribe(
         &self,
-        valid_entries: HashMap<i32, HashSet<Field>>,
+        valid_entries: HashMap<i32, (HashSet<Field>, types::ChangeType)>,
```
argerus, line 1490: Why is `ChangeType` added as an argument here? Each signal already has a `ChangeType` associated with it, so why should the caller of `subscribe` supply this information?

What would be a relevant argument to add here though, is at what frequency the caller wants continuous signals to be updated at. And this would only be required for signals of type `ChangeType::Continuous`.

kuksa_databroker/databroker/src/viss/v2/server.rs
```diff
@@ -253,7 +253,15 @@ impl Viss for Server {
...
-            .map(|id| HashMap::from([(id, HashSet::from([broker::Field::Datapoint]))]))
+            .map(|id| {
+                HashMap::from([(
+                    id,
+                    (
+                        HashSet::from([broker::Field::Datapoint]),
+                        ChangeType::Static,
+                    ),
+                )])
+            })
```
argerus, line 265: I don't see how it makes sense to add the `ChangeType` argument to the subscribe function as that is a property already associated with each signal (see other comment). And if it did make sense, why set it to `ChangeType::Static` here?

kuksa_databroker/databroker/src/broker.rs
rafaeling: I had to choose between changing the subscribe method parameters to pass the entire entry and change the method implementation so I went for the option that makes less changes in the code by only adding the change type info.
About the frequency I do agree, I dont if I should include it in this PR or make a new one for it since I would have to change the proto file to include the frequency param.

argerus: I'm pretty sure you don't need to change the signature at all as you have full access to the db in the `subscribe` function (and hence can get the change_type given an id).

kuksa_databroker/databroker/src/viss/v2/server.rs
rafaeling: Changes reverted

argerus:
> Are you happy with the content of this PR now @argerus ? I.e. is it ready for merge?

Probably not. As is currently stands, this PR sets a hard coded update frequency of 1 second for everything, with no way to set it to anything else.

This combined with the fact that we don't include a VSS db with the change types configured, and every signal now defaults to `Continuous` is probably not the behaviour we want. I do have a version of the VSS db with all change types configured, so perhaps that will make the out of the box behaviour better (if we merge that first).

I still think this PR should include the ability to change the update frequency through the API, otherwise I'm not sure much is gained by merging it.

rafaeling:
Will then add the frequency parameter to the subscribe request on this PR, was thinking to split it into different PRs but makes no sense.

kuksa_databroker/databroker/src/broker.rs
```diff
@@ -921,6 +961,117 @@ impl QuerySubscription {
     }
 }
 
+impl ContinuousSubscription {
+    async fn notify(
```
lukasmittag, line 965: I see that this is very similar to the ChangeSubscription notify. Couldn't it be reused somehow? What I mean is just use different subscribe function but maybe one notify? If this is not doable let me know :)

proto/kuksa/val/v1/val.proto
```diff
@@ -98,6 +98,7 @@ message SubscribeEntry {
 // Subscribe to changes in datapoints.
 message SubscribeRequest {
   repeated SubscribeEntry entries = 1;
+  optional uint64 frequency_hertz = 2;
```
erikbosch, line 101: Some thoughts

- Do we see Hertz and uint64 as a reasonable representation. That makes it impossible to support intervals longer than 1 second. An alternative solution could be to for example use `interval_ms`, i.e. wanted interval in milliseconds. That would allow us to represent both 100 Hz as well as 0.1 Hz
- Should we possible already in this file describe semantics? That would be good if we later intend to promote this API to a "standardized" API. I.e. document topics like
  -  What is default if this field/member is not given? Max frequency? On change? Other?
  - What shall happen if 0 as given as frequency? Does that imply "on change" or does it just mean "as fast as possible, changed or not"
  - What is the semantic meaning of specifying frequency - is it getting latest value at that frequency regardless of whether the value has been updated (new timestamp) or changed (changed value)
  - What shall happen if frequency/interval cannot be supported. Like if you request 2000 Hz - should you get an error?

kuksa_databroker/databroker/src/broker.rs
```diff
@@ -605,6 +615,27 @@ impl Subscriptions {
...
+            .lock()
+            .unwrap()
+            .push(subscription);
+
+        tokio::spawn(async move {
+            // Asynchronous code to be executed in the new task
+            while !local_subscription.sender.is_closed() {
+                let _ = local_subscription.notify(None).await;
+                // Simulate some asynchronous work
+                tokio::time::sleep(Duration::from_millis(1 / frequency * 1000)).await;
```
erikbosch, line 634: How does Rust manage type conversion? `from_millis` expects a u64.  You use `1` and `1000`, will it really be a float value? Or do you rather need something like:

`Duration::from_millis((1000.0/ frequency).round() as u64)`

(I actually do not know what resolution/performance we can expect from the system, like if we request 1000 Hz, will we really receive signals with 1000 Hz?

lukasmittag: That would be really interesting e.g. how fast is databroker able to notify subscribers. Does it change if we have 100 subscribers at the same time. What is the limit e.g. what Erik wrote. But I guess not that much related to this PR.

lukasmittag:
we need to make sure that the changes are reflected in the sdk's. Python one https://github.com/eclipse-kuksa/kuksa-python-sdk/pull/21 is taken care of already. Do we want to have it in GO fixed as well? Rust SDK is working I guess?

proto/kuksa/val/v1/val.proto
rafaeling: I have completely missed that frequency could range from 0 to 1 as well, so uint64 is totally wrong, better to use float type (will push the fix). I would still consider to use frequency hertz representation better than interval_ms since hertz is also used by VHAL interface for Continuous properties -> https://source.android.com/docs/automotive/vhal/vhal-interface

Regarding your questions:

> 1. What is default if this field/member is not given? Max frequency? On change? Other?

- The initial default behaviour implemented works as follow: 

&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;1. if the Subscribe request contains **any** continuous signal without a frequency parameter it returns an _InvalidArgument_ error.
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;2. Same in case the request doesn't contain **any** continuous signal and a frequency parameter is given it must return an _InvalidArgument_ error.

> 2. What shall happen if 0 as given as frequency? Does that imply "on change" or does it just mean "as fast as possible, changed or not"
- No behaviour specified, I would suggest to return an _InvalidArgument_ error as well since it would mean to divide some number by 0. What do you think?

> 3. What is the semantic meaning of specifying frequency - is it getting latest value at that frequency regardless of whether the value has been updated (new timestamp) or changed (changed value)

- The semantic meaning is to get latest value at that frequency regardless of whether the value has been updated or changed with its latest update or change timestamp.

> 4. What shall happen if frequency/interval cannot be supported. Like if you request 2000 Hz - should you get an error?

- We should discuss this again, we can suggest the customer to specific its own frequency limit rate at broker initialization (should to be implemented) depending on its hardware capabilities.

erikbosch: I think it could be good to have a configurable "max frequency", similar to the "max number of signals" topic we discussed as part of the Audit findings. Then one could possible let the frequency 0 mean "as fast as possible" rather than considering it as an illegal value.

Anyway, maybe we should document this with comments so we could generate documentation with frameworks like https://github.com/pseudomuto/protoc-gen-doc so that expected behavior is clearly defined

rafaeling:
After community call was agreed to avoid changes in VSS and to add the "ChangeType" directly to the subscription request implicitly with the optional frequency parameter.

erikbosch:
Reopening all PRs closed by accident by boschglobal maintenance

proto/kuksa/val/v1/val.proto
BjoernAtBosch: Findings:
* Regardless, if we use "frequency" or "interval": It should be already clear from the parameter name, what it means. Currently, I cannot see (without reading the documentation) if it is minimum, typical, or maximum frequency/interval.
* Don't we add detailed documentation about the behaviour in the proto files? I agree with Erik: We should also describe the semantics in the interface definition. Imho, the "outside visible behaviour" always belongs to the contract description of an interface definition.
* In the current definition this parameter is related to the overall subscription, but not to certain signals - which is fine for me if the parameter defines the max update frequency. But the behaviour should be clearly defined if the the subscription comprises signals with different update frequencies.
* What is the behaviour, if signals are updated faster than the max frequency requested by the client: 
  * Does it get "just" some sample value at the moment the update interval is reached,
  * do we send some mean value of the last samples (how would that be calculated?),
  * do we accumulated values and send them as a batch,
  * ...?
If we are unsure about the behaviour to be achieved by this PR, we should rethink merging it ...

BjoernAtBosch (commented):
Just looked at the proto changes. (My Rust knowledge is too poor to rate the rest, currently)

erikbosch:

PR #739, opened 2024-02-15
Disable sdv.databroker.v1 by default

argerus:
- Disable `sdv.databroker.v1` by default. It can be enabled by supplying a flag at startup.
- Change databroker-cli to use `kuksa.val.v1` by default.

### DIFF
--- a/kuksa_databroker/databroker-cli/src/cli.rs
+++ b/kuksa_databroker/databroker-cli/src/cli.rs
@@ -52,8 +52,8 @@ pub struct Cli {
     ca_cert: Option<String>,
 
     #[arg(value_enum)]
-    #[clap(long, short = 'p', value_enum, default_value = "sdv-databroker-v1")]
-    protocol: CliAPI,
+    #[clap(long, short = 'p', value_enum, default_value_t = Protocol::KuksaValV1)]
+    protocol: Protocol,
 
     // Sub command
     #[clap(subcommand)]
@@ -77,7 +77,7 @@ impl Cli {
         self.server.clone()
     }
 
-    pub fn get_protocol(&mut self) -> CliAPI {
+    pub fn get_protocol(&mut self) -> Protocol {
         self.protocol
     }
 }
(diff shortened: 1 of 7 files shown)


erikbosch:
Great progress! When everything is working we need to update the documentation. I noticed some inconsistencies in the existing documentation. In the main [README](https://github.com/eclipse/kuksa.val/blob/master/kuksa_databroker/README.md) we mention the `--protocol` option, but we do not mention it in the [user guide](https://github.com/eclipse/kuksa.val/blob/master/kuksa_databroker/doc/user_guide.md).

Do we possibly need some form of changelog to keep track of more important changes (like this one?). I guess we also need to discuss if this change is so big that it implies that we shall call the next Databroker version  0.5?

SebastianSchildt:
Integration test  currently fails as this https://github.com/eclipse/kuksa.val/blob/master/kuksa_databroker/integration_test/test_databroker.py depends on the old API.

Should be updated. This could be rewritten using the python sdk, but as this is a very basic "GRPC-level" test might be better to keep it like this, just chosing a number of kuksa.val.v1 calls. This is currently not a real test whether the API behaves 100% in a certain way, but rather a rain check that it came up at all, is connectable  and seems to to do "something useful"

SebastianSchildt (changes requested):
https://github.com/eclipse/kuksa.val/blob/master/kuksa_databroker/integration_test/test_databroker.py needs to be adapted to have a working integration test

erikbosch:
Reopening all PRs closed by accident by boschglobal maintenance

SebastianSchildt:
is there still progress here? Somebody fixing the test?

erikbosch:
> is there still progress here? Somebody fixing the test?

We have sort of agreed to NOT fix this in this repo to avoid causing backward compatibility for anyone using the old repo, but rather fix it after migration, i.e. in https://github.com/eclipse-kuksa/kuksa-databroker

PR #744, opened 2024-02-25
Switch to 2021 edition of Rust

### LINKED ISSUE #328: Cargo edition/version requirements for databroker
In the databroker cargo file edition 2018 is stated as requirement, but when I tried to build it complained. Reason seems to be the dependency to clap v 3.1.10 (using 2018) that by my cargo was replaced by the newer v3.2.16 which use edition 2021.

My understanding (with limited cargo/rust knowledge) is that we:

- Either need to change the `Cargo.toml` file so that cargo does not try to load newer versions with edition 2021 requirements
- Or we mention edition 2021 as a build requirement (and possibly change 2018 to 2021 in the `Cargo.toml`

https://github.com/eclipse/kuksa.val/blob/master/kuksa_databroker/databroker/Cargo.toml

SebastianSchildt:
This upgrades RUST edition to 2021 and also explicitly enables the new dependency resolver

This is supported since RUST 1.56, and usually we are building with much newer versions anyway

https://blog.rust-lang.org/2021/05/11/edition-2021.html
https://doc.rust-lang.org/edition-guide/rust-2021/default-cargo-resolver.html

The "next" release will be the 2024 edition, so guess switching to 2021 is pretty sensible now
https://blog.rust-lang.org/2023/12/15/2024-Edition-CFP.html

Will close #328

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -52,9 +52,9 @@ dependencies = [
 
 [[package]]
 name = "anstream"
-version = "0.6.11"
+version = "0.6.12"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "6e2e1ebcb11de5c03c67de28a7df593d32191b44939c482e97702baaaa6ab6a5"
+checksum = "96b09b5178381e0874812a9b157f7fe84982617e48f71f4e3235482775e5b540"
 dependencies = [
  "anstyle",
  "anstyle-parse",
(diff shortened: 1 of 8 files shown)


erikbosch:
If we update a `Cargo.toml`, should we then request that PR creator also have done a `cargo update` so that `Cargo.lock` gets updated? 

As I understand it does not affect builds, but Dependabot use it when scanning for vulnerabilities. As of today we update it as part of creating releases, but do we see a need to update also in other cases like when a `Cargo.toml` is updated?

SebastianSchildt:
I updated Cargo.lock now. I think it is good practice when touching Cargo files

erikbosch (approved):
Looks good to me, but I am no Rust expert.

Do you see any problems @argerus

argerus:
> Do you see any problems @argerus

Nope, looks good to me.

argerus (approved):

PR #747, opened 2024-03-01
Remove token/certificate scripts from kuksa.val

erikbosch:
And instead refer to kuksa-common

To be merged after https://github.com/eclipse-kuksa/kuksa-common/pull/1


SebastianSchildt (approved):
Looking good, PR in common is merged, so less duplication. lgtm 🚯

PR #749, opened 2024-03-06
Use resolver=2 in Cargo.toml

lukasmittag:
use the resolver = 2 in Cargo.toml since it is the default resolver for rust edition 2021. ```cargo update``` failed for lib crate.


erikbosch (approved):
LGTM

PR #750, opened 2024-03-06
LOW-5 Malicious JWT access token can crash a thread of the databroker

lukasmittag:
use safe addition ```checked_add``` in decoder.rs to catch overflows and do not ```panic```. Instead returning ```ClaimsError```.

### DIFF
--- a/kuksa_databroker/databroker/src/authorization/jwt/decoder.rs
+++ b/kuksa_databroker/databroker/src/authorization/jwt/decoder.rs
@@ -130,8 +130,13 @@ impl TryFrom<Claims> for Permissions {
             }
         }
 
-        permissions = permissions
-            .expires_at(std::time::UNIX_EPOCH + std::time::Duration::from_secs(claims.exp));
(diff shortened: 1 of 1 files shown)


rafaeling:
Code looks good, some unit test would be good

kuksa_databroker/databroker/src/authorization/jwt/decoder.rs
```diff
@@ -171,4 +177,34 @@ AEiqOjPq0D6X45wCzIwjILUCAwEAAQ==
             Err(err) => panic!("decode should succeed but failed with:{}", err),
         }
     }
+
+    #[test]
+    fn test_expiration_safe_addition() {
```
erikbosch, line 182: Would it possibly be useful with a small comment here specifying the purpose of the test, possibly something like "Rust can only handle date up to XXXX, verify that an error is reported if token expires in year 2424"

erikbosch (approved):
Looks good to me

PR #751, opened 2024-03-08
INFO-3 Subscription channels remain open after token expiration

lukasmittag:
This fixes that a subscription channel stays open after the token of the subscriber has expired. We need to decide if we want to fix this. Currently it is only fixed for ```ChangeSubscription``` which the new API uses. If we want to do it for the old query subscription as well it can  be done similiar. To test it take a jwt token of kuksa-val/jwt/ and use a timestamp that expires in the near future -> go to https://www.epochconverter.com/ for timestamps ;)

It is not that pretty in code I think so do not now if we want to fix this in this way/at all. One observation I made was that even if the channel is kept open somehow if you update the token new updates do not come in in the databroker-cli again. Haven't figured out why yet.

### DIFF
--- a/kuksa_databroker/databroker/src/broker.rs
+++ b/kuksa_databroker/databroker/src/broker.rs
@@ -23,6 +23,7 @@ use tokio_stream::Stream;
 
 use std::collections::{HashMap, HashSet};
 use std::convert::TryFrom;
+use std::mem::take;
 use std::sync::atomic::{AtomicI32, Ordering};
 use std::sync::Arc;
 use std::time::SystemTime;
@@ -154,7 +155,7 @@ pub struct QuerySubscription {
 
 pub struct ChangeSubscription {
     entries: HashMap<i32, HashSet<Field>>,
-    sender: mpsc::Sender<EntryUpdates>,
+    sender: Option<mpsc::Sender<EntryUpdates>>,
     permissions: Permissions,
 }
 
@@ -606,7 +607,7 @@ impl Subscriptions {
     }
 
     pub async fn notify(
-        &self,
+        &mut self,
         changed: Option<&HashMap<i32, HashSet<Field>>>,
         db: &Database,
     ) -> Result<Option<HashMap<String, ()>>, NotificationError> {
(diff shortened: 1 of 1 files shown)


lukasmittag:
and I opted for closing if a new notification would be sent out so no general repeated check. This means if you do not update soem things channel can stay open a while.

rafaeling:
Code looks fine to me, tested with python sdk and the channel closes correclty

erikbosch (approved):
Looks ok for me. But I did not really understood one of the comment, see below. Is that one still valid, or do you think the code is pretty enough (kenough) now?

"It is not that pretty in code I think so do not now if we want to fix this in this way/at all."

lukasmittag:
I just thought that it's maybe not the ideal way code wise. But since @argerus had nothing to complain yet :) From my side we can merge.

kuksa_databroker/databroker/src/broker.rs
```diff
@@ -626,10 +627,14 @@ impl Subscriptions {
...
 
-        for sub in &self.change_subscriptions {
+        for sub in &mut self.change_subscriptions {
             match sub.notify(changed, db).await {
                 Ok(_) => {}
-                Err(err) => error = Some(err),
+                Err(err) => {
+                    error = Some(err);
+                    let taken_sender = take(&mut sub.sender);
+                    drop(taken_sender);
```
argerus, line 636: I don't understand the logic of what is going on here at all... No need to use `mem::take` (to replace the sender with a "default" value?) only to immediately `drop` it.

Just set it to `None`, which I assume was your intention when changed it to `Option<Sender>`, right?

```suggestion
                    sub.sender = None:
```

argerus (changes requested):
It would probably be a lot easier to treat expired permissions the same way we treat channels that have been closed:

e.g. add a check to `Subscriptions::cleanup()` that will remove all subscriptions with expired permissions. This will be triggered both periodically, and whenever a `SubscriptionError` is returned when when calling notify.

kuksa_databroker/databroker/src/broker.rs
argerus: (or revert this part and drop subscriptions the same way it's done when a closed channel is detected, see general comment)

lukasmittag: your second suggestion makes the most sense from my POV. I will handle it in cleanup method if it gets called periodically and on error. Thanks for the remarks :)

argerus: Awesome :wink:

erikbosch:

PR #758, opened 2024-03-18
Fixed query table header in docs

sgloutnikov:
Updated markdown table header to properly render in preview


erikbosch:
Thanks for the PR. There is a pre-commit error, caused by some trailing blanks and other errors. It does not seem to be caused by your change, rather part of our somewhat new policy to require no trailing line breaks for touched files, so if you happen to touch a file with trailing line blanks CI will fail unless you fix them. If possible, it would be great if you could fix them,see https://github.com/eclipse/kuksa.val?tab=readme-ov-file#pre-commit-set-up

Previous table as rendered:

image

erikbosch (approved):

Issue #752, opened 2024-03-11
Unknown/unknown docker images


erikbosch:
I noticed that most of our docker images published on ghcr.io also shows an unknown/unknown version, like [Databroker](https://github.com/eclipse/kuksa.val/pkgs/container/kuksa.val%2Fdatabroker)

![image](https://github.com/eclipse/kuksa.val/assets/30996601/150371b6-cfc1-4de9-8ab6-9c7a04632334)

We addressed this for SOME/IP feeder long time ago, see https://github.com/eclipse/kuksa.val.feeders/pull/118, by specifying `provenance: false`. We may consider if we should add it also to all other docker builds

```
 uses: docker/build-push-action@v4
with:
    # Provenance to solve that an unknown/unkown image is shown on ghcr.io
    # Same problem as described in https://github.com/orgs/community/discussions/45969
    provenance: false
```

erikbosch:
We are about to archive this repo soon. If you consider this issue as important please file a new issue at one of the new Kuksa repos at https://github.com/eclipse-kuksa

PR #759, opened 2024-03-19
Fix unknown/unknown docker images

erikbosch:
See https://github.com/eclipse/kuksa.val/issues/752

Also fixing some NodeJS warnings


lukasmittag (approved):
LGTM

PR #760, opened 2024-03-19
Fix dependabot issue

erikbosch:
See https://github.com/eclipse/kuksa.val/security/dependabot/57

Also addressing some Node.js  and linter warnings.
This became somewhat bigger than expected. We want to install latest protobuf/grpc and they require go 1.19 at least, but in 1.19 some code is deprecated which the linter did not like.

Smoke test according to https://github.com/eclipse/kuksa.val/wiki/Release-Testing#go-client performed


SebastianSchildt (approved):
Looking good.

Now having invested so much love into go client, we should really move it t incubation 😄

PR #763, opened 2024-04-03
Update and deprecate node-red example

erikbosch:
Update due to a vulnerability
Also make it clear that it is deprecated.
We do not intend to migrate it.

Lock file updated by `npm i --package-lock-only`

Verification done: Checked that we could start NodeRed like below.

image


lukasmittag (approved):
LGTM

PR #769, opened 2024-04-29
Removing databroker

erikbosch:
As migration is finished no need to keep it here.
Version 0.4.3 tagged and docker containers created, verified that they give warnings


erikbosch:
Anything more you want that we shall do before removing it @SebastianSchildt .
I suggest that after you approve it that we ping (or add as reviewer) John, Mike, SvenErik, Lukas, so they are aware that it will be deleted, and can complain before we do it

SebastianSchildt (approved):
Looks ok to me, and main Readme already has a deprecation notice on top.

MIght be argued it doesn't _need_ to be deleted in main, but on the other hand, doing so reduces the chance, that somebody "accidentally' starts working with an old code.

Would be good 1 or 2 more opinions on that, as you mentioned, e.g. @argerus  @eriksven  @lukasmittag  @rafaeling  @mikehaller

erikbosch:
We have used that policy in the past, to remove the code in the old location. Reduces the risk for PRs and Issues in the old repo.

lukasmittag:
couldn't we keep a branch open like deprecated_databrokeror something like this? Then documenting clearly that it is there and you should not use it. But at least we then keep the state for I do not know end of the year?

erikbosch:
@lukasmittag - I added a branch https://github.com/eclipse/kuksa.val/tree/deprecated_databroker from previous commit, and references that branch from the README now. So if someone by some reason do not want to use the partially refactored databroker in kuksa-databroker and they want to check out code by branch rather than tag they can use the `deprecated_databroker` branch.

PR #773, opened 2024-11-05
adds archival message to Readme

eriksven:


eriksven:
Since there has not been much activity in the past 6 months on this repository, I suggest that we archive the repository even before the initially set EOL date. The whole development has already been moved to the repositories referenced in the Readme file. 

The reason why I am bringing this up again is that the Eclipse Foundation plans to move repositories from the Eclipse GitHub organization to project-specific organizations in GitHub. For more details see https://gitlab.eclipse.org/eclipsefdn/helpdesk/-/issues/4956 . 

To not block this process and avoid confusion in case this deprecated repository still gets moved to the eclipse-kuksa organization, I think it is better to do the archival now.

bs-jokri (approved):
LGTM and +1 for archiving the repo

erikbosch (approved):
Approved

